Issue #26, opened 2026-02-25
Cache config reads to avoid repeated YAML file I/O


gabrielkoerich:
## Problem

Issue #6: `Config get()` re-reads YAML from disk on every call.

Currently `src/config/mod.rs` loads and parses `config.yml` + `.orchestrator.yml` on every `config::get()` call. This is inefficient and defeats the purpose of in-memory config.

## Impact

From PLAN.md: "Config loading — ~10 forks eliminated" is listed as a benefit of moving to Rust. But without caching, we're still doing file I/O on every read.

## Proposed Solution

Add a config cache with the following features:

1. **In-memory cache**: Store parsed config in a `RwLock<HashMap<String, serde_yaml::Value>>`
2. **File watching**: Use `notify` crate to watch config files for changes
3. **Cache invalidation**: When config files change, reload automatically
4. **Hot reload support**: The engine should reload config on change without restart

## Files to Modify

- `src/config/mod.rs`: Add caching layer
- Add `notify` to Cargo.toml for file watching

## Implementation Sketch

```rust
use std::sync::RwLock;
use std::collections::HashMap;

static CACHE: once_cell::sync::Lazy<RwLock<ConfigCache>> = 
    once_cell::sync::Lazy::new(|| RwLock::new(ConfigCache::default()));

pub fn get(key: &str) -> anyhow::Result<String> {
    let cache = CACHE.read().unwrap();
    // Try cache first
    if let Some(value) = cache.values.get(key) {
        return Ok(value.to_string());
    }
    // Fall back to disk read + populate cache
    drop(cache);
    reload()?;
    // ... return cached value
}
```

## Benefits

- Eliminates repeated YAML parsing (currently every call)
- Enables hot config reload (no service restart needed)
- Consistent with Phase 1 goal: "Config loading — in-memory struct, hot-reload"

## References

- Issue #6 in this repo
- PLAN.md Phase 1: "Config loading (config.yml, .orchestrator.yml) — in-memory struct, hot-reload"
- `src/config/mod.rs` — current implementation

PR #30, opened 2026-02-25
feat: cache config reads with file watching for hot reload

gabrielkoerich:
## Summary
- Add `notify` crate for file watching
- Implement in-memory config cache with RwLock for concurrent reads
- Watch config.yml and .orchestrator.yml files for changes
- Auto-invalidate cache when config files are modified (hot reload)

This implements issue #26: Config get() re-reads YAML from disk on every call. The config is now cached in memory and automatically reloaded when config files change on disk.

## Test plan
- [x] Rust tests pass (`cargo test`)
- [ ] Verify config caching works (multiple calls don't re-read disk)
- [ ] Verify hot reload works (edit config.yml, changes reflected on next read)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -107,6 +107,12 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "bitflags"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
@@ -322,6 +328,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
 [[package]]
+name = "filetime"
+version = "0.2.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "libredox",
+]
+
+[[package]]
 name = "find-msvc-tools"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -332,6 +349,15 @@ name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
+
+[[package]]
+name = "fsevent-sys"
+version = "4.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76ee7a02da4d231650c7cea31349b889be2f45ddb3ef3032d2ec8185f6313fd2"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "futures-core"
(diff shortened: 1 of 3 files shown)


gabrielkoerich:

Several correctness and performance issues need to be addressed before merging.

## Critical: Stale cache after invalidation race (src/config/mod.rs)

The most serious bug is a race between cache repopulation and invalidation:
1. Thread A observes a cache miss and reads file v1 from disk (read lock dropped)
2. The file changes on disk; the watcher fires and calls `invalidate_cache` (removes the entry)
3. Thread A then acquires the write lock and inserts stale v1

Now the cache holds stale data, and the watcher won't fire again until the next file change. The cache is permanently poisoned. Fix: hold the write lock across the disk read and insert (the original Mutex approach actually prevented this), or validate that the file mtime hasn't changed before inserting.

## Critical: `WATCHER.write()` acquired on every `resolve_key` call (src/config/mod.rs ~line 110)

`watch_file(path)` is called unconditionally at the top of `resolve_key`, before even checking the cache. Inside `watch_file`, it always acquires `WATCHER.write()`. This means every single config read — including cache hits — serializes all concurrent callers through a write lock. This completely defeats the purpose of switching from `Mutex` to `RwLock` for the cache. Fix: use a read-check-then-write pattern in `watch_file`:
```rust
fn watch_file(path: &PathBuf) {
    if let Ok(w) = WATCHER.read() {
        if w.contains_key(path) { return; }  // fast path
    }
    ensure_watcher();
    if let Ok(mut w) = WATCHER.write() {
        if !w.contains_key(path) {  // re-check under write lock
            // register with notify...
        }
    }
}
```

## High: Path canonicalization missing — hot reload will silently fail (src/config/mod.rs)

The event path reported by the OS watcher (e.g., via FSEvents on macOS or inotify on Linux) may differ from the path used as the cache key. For example, `/private/var/folders/...` vs `/var/folders/...` (macOS symlinks), or paths with `./` components. If the path in the event doesn't exactly match the key in `CACHE`, `cache.remove(path)` is a no-op and the cache is never invalidated. All paths must be canonicalized with `std::fs::canonicalize` before being used as cache keys, before watch registration, and before invalidation.

## High: Lock ordering issue — potential deadlock (src/config/mod.rs ~line 70)

In `watch_file`, the code holds `WATCHER.write()` and then acquires `FILE_WATCHER.lock()`:
```rust
if let Ok(mut watched) = WATCHER.write() {         // lock A
    ...
    if let Ok(mut guard) = FILE_WATCHER.lock() {   // lock B (nested)
```
If any future code path acquires `FILE_WATCHER` first and then calls into code that needs `WATCHER`, you get a classic AB/BA deadlock. The lock ordering should be documented and enforced: always acquire `FILE_WATCHER` before `WATCHER`, or restructure so they are never held simultaneously.

## Medium: `with_poll_interval` is silently ignored on most platforms (src/config/mod.rs ~line 50)

`Config::default().with_poll_interval(Duration::from_secs(2))` only applies when using `PollWatcher`. `RecommendedWatcher` on macOS uses FSEvents and on Linux uses inotify — both ignore this setting. The configuration call is a no-op and may mislead readers into thinking polling is the mechanism.

## Medium: Watching file paths instead of parent directories (src/config/mod.rs ~line 82)

`watcher.watch(path, RecursiveMode::NonRecursive)` watches the file directly. On Linux with inotify, many editors (vim, emacs) write via a temp file + rename, which means the watched inode is deleted and recreated. inotify on a file path will miss these events. The robust approach is to watch the parent directory and filter events by filename (the callback already does this filtering, so moving the watch target to `path.parent()` would fix this).

## Medium: Watcher creation failure is silently ignored (src/config/mod.rs ~line 60)

```rust
if let Ok(w) = watcher {
    *watcher_guard = Some(w);
}
```
If watcher creation fails (e.g., inotify fd limit), `FILE_WATCHER` stays `None`, `watch_file` silently returns (inserts path into `WATCHER` but registers nothing with notify), and hot reload silently never works. At minimum this should log a warning. Additionally, subsequent calls to `ensure_watcher` will see `is_some() == false` and retry watcher creation on every config read, which could be noisy.

## Minor: `Arc` wrapping in `FILE_WATCHER` is unnecessary

`static FILE_WATCHER: LazyLock<Arc<Mutex<Option<RecommendedWatcher>>>>` — since this is a static, the `Arc` provides no benefit. It adds an allocation and indirection. Just use `LazyLock<Mutex<Option<RecommendedWatcher>>>`.

## Missing tests

The PR's own test plan marks the two most important cases as unverified:
- [ ] Verify config caching works (multiple calls don't re-read disk)
- [ ] Verify hot reload works (edit config.yml, changes reflected on next read)

Given the race conditions described above, integration tests for these behaviors are important before merging.

---
*By claude[bot] using model `sonnet` via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

PR #50, opened 2026-02-25
feat: implement parent/child task unblocking logic

### LINKED ISSUE #47: Implement parent/child task unblocking logic
## Summary
The engine has a placeholder for unblocking parent tasks when children complete (src/engine/mod.rs:336-341). This is critical for task dependencies to work.

## Current State (src/engine/mod.rs:336-341)
```rust
// Phase 4: Unblock parents (blocked tasks whose children are all done)
let blocked = backend.list_by_status(Status::Blocked).await?;
for task in &blocked {
    // Check if all sub-issues are done
    // TODO: query sub-issues API to check children status
    let _ = task; // placeholder
}
```

### 1. Query Sub-Issues (src/backends/github.rs)
GitHub recently added sub-issues API. Add method to GitHubBackend:

```rust
async fn get_sub_issues(&self, id: &ExternalId) -> anyhow::Result<Vec<ExternalId>> {
    // Use GraphQL API:
    // query { 
    //   repository(owner, name) { 
    //     issue(number) { 
    //       subIssues { nodes { number } }
    //     }
    //   }
    // }
}
```

### 2. Check Children Status (src/engine/mod.rs)
Implement the unblocking logic:

```rust
// Phase 4: Unblock parents
let blocked = backend.list_by_status(Status::Blocked).await?;
for task in &blocked {
    // Get sub-issues
    let children = backend.get_sub_issues(&task.id).await?;
    
    if children.is_empty() {
        // No children - should not be blocked, maybe recover?
        continue;
    }
    
    // Check if all children are done
    let all_done = futures::future::try_join_all(
        children.iter().map(|id| backend.get_task(id))
    ).await?.iter().all(|t| {
        t.labels.iter().any(|l| l == "status:done")
    });
    
    if all_done {
        backend.update_status(&task.id, Status::New).await?;
        backend.post_comment(
            &task.id,
            "All sub-tasks completed. Unblocking parent task."
        ).await?;
    }
}
```

### 3. Add to TaskManager
Move this logic to TaskManager for unified handling:

```rust
impl TaskManager {
    pub async fn unblock_parents(&self
    ) -> anyhow::Result<Vec<i64>> {
        // Returns list of unblocked task IDs
    }
}
```

## Alternative: SQLite for Parent/Child
If GitHub sub-issues API is not available, track parent/child in SQLite:

```sql
CREATE TABLE task_relationships (
    parent_id TEXT NOT NULL,
    child_id TEXT NOT NULL,
    relationship TEXT DEFAULT 'subtask', -- subtask, blocks, etc
    PRIMARY KEY (parent_id, child_id)
);
```

## Files to Modify
- `src/backends/mod.rs` - add get_sub_issues to trait
- `src/backends/github.rs` - implement using GraphQL
- `src/engine/mod.rs` - implement unblocking logic
- `src/engine/tasks.rs` - add to TaskManager

## Testing
- Create parent task with sub-issues
- Complete all children
- Verify parent is unblocked
- Test with mixed internal/external tasks

## Documentation
- Document sub-issue workflow in AGENTS.md
- Update PLAN.md to mark parent/child as complete

gabrielkoerich:
## Summary

Add support for unblocking parent tasks when all sub-issues are completed:

- Add `get_sub_issues` method to ExternalBackend trait
- Implement GitHub GraphQL API for querying sub-issues  
- Implement unblocking logic in engine tick (Phase 4)
- Add `unblock_parents` method to TaskManager for external use

The implementation queries GitHub's sub-issues API to find children, checks if all children have the "status:done" label, and unblocks the parent task by updating its status to "status:new" and posting a comment.

## Test plan

- [ ] Create parent task with sub-issues
- [ ] Complete all children
- [ ] Verify parent is unblocked
- [ ] Test with mixed internal/external tasks

Closes #47

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -444,7 +444,10 @@ async fn tick(
         }
 
         if all_done {
-            tracing::info!(task_id = task.id.0, "all sub-tasks completed, unblocking parent");
+            tracing::info!(
+                task_id = task.id.0,
+                "all sub-tasks completed, unblocking parent"
+            );
             backend.update_status(&task.id, Status::New).await?;
             backend
                 .post_comment(
--- a/src/engine/tasks.rs
+++ b/src/engine/tasks.rs
@@ -211,7 +211,10 @@ impl TaskManager {
             }
 
             if all_done {
-                tracing::info!(task_id = task.id.0, "all sub-tasks completed, unblocking parent");
+                tracing::info!(
+                    task_id = task.id.0,
+                    "all sub-tasks completed, unblocking parent"
+                );
                 self.backend.update_status(&task.id, Status::New).await?;
                 self.backend
                     .post_comment(


gabrielkoerich:

## Review of PR #50: feat: implement parent/child task unblocking logic

### Critical Bug: URL-encoding corrupts the GraphQL query

**File:** `src/github/cli.rs`, `graphql()` method (~line 296–308)

```rust
.arg(format!("query={}", urlencoding::encode(query)))
```

When arguments are passed via `Command::new("gh").arg(...)`, they are delivered directly to the subprocess without shell processing — there is no HTTP layer here for URL encoding to be useful. The `gh api graphql` binary receives the literal URL-encoded string (e.g., `%7B%0A%20%20repository...`) as the query value, which is not a valid GraphQL query. This will cause every GraphQL call to fail at runtime.

Fix: remove the encoding and pass the raw query string:
```rust
.arg(format!("query={}", query))
```
Or, for queries with complex quoting, pipe via stdin using `gh api graphql --input -`.

---

### Potential Build Failure: `urlencoding` crate not shown in Cargo.toml

The `graphql()` method calls `urlencoding::encode()`, but no change to `Cargo.toml` is included in the diff. If this crate is not already a declared dependency the build will fail. Even if it is, removing the encode call (per the bug above) makes this moot.

---

### Major: Duplicate unblocking logic between engine tick and TaskManager

**Files:** `src/engine/mod.rs:339–385` and `src/engine/tasks.rs:172–233`

The child-iteration, all_done check, `update_status`, and `post_comment` logic is copy-pasted nearly verbatim between the engine tick (Phase 4) and `TaskManager::unblock_parents`. If the unblocking semantics ever change (e.g., add a different status transition, change the comment text, add idempotency guards), both sites must be updated in sync.

The engine tick should delegate to `TaskManager::unblock_parents()` rather than duplicating the logic, or the shared logic should be extracted to a free function.

---

### Moderate: Missing `get_sub_issues` implementations for mock/test backends

**File:** `src/backends/mod.rs`

Adding a method to the `ExternalBackend` trait requires all implementing types to provide it. The diff only shows the `GitHubBackend` implementation. If there are mock or in-memory backends used in tests (common in a project with a test suite), those will fail to compile. No such implementations appear in the diff.

---

### Moderate: Hardcoded pagination silently truncates large sub-issue lists

**File:** `src/github/cli.rs:339`

```graphql
subIssues(first: 100)
```

If a parent task has more than 100 sub-issues, the extras are silently dropped. `all_done` would return `true` prematurely if the undiscovered children are incomplete. At a minimum, emit a `tracing::warn!` when exactly 100 nodes are returned to flag possible truncation. A proper fix would implement cursor-based pagination.

---

### Moderate: GraphQL injection via string interpolation

**File:** `src/github/cli.rs:327–341`

```rust
let query = format!(
    r#"{{ repository(owner: "{}", name: "{}") {{ ... }} }}"#,
    owner, repo_name, number
);
```

`owner`, `repo_name`, and `number` are interpolated directly into the query string. While `number` is a stored task ID (likely numeric), `owner` and `repo_name` come from configuration and could contain `"` characters that break the query structure. The correct approach is to pass them as typed GraphQL variables:

```bash
gh api graphql \
  -f query='query($owner:String!, $name:String!, $number:Int!) { repository(owner:$owner, name:$name) { issue(number:$number) { subIssues(first:100) { nodes { number } } } } }' \
  -F owner=OWNER -F name=NAME -F number=NUM
```

---

### Minor: Silent parse failure in `unblock_parents` return value

**File:** `src/engine/tasks.rs:228`

```rust
if let Ok(id_num) = task.id.0.parse::<i64>() {
    unblocked.push(id_num);
}
```

If an `ExternalId` cannot be parsed as `i64`, the task is silently omitted from the returned list with no trace. Add a `tracing::warn!` on the else branch so callers can detect when the return value is incomplete.

---

### Summary

The URL-encoding bug is a definite runtime failure for every GraphQL call — no sub-issues will ever be fetched, making the core feature non-functional. The missing `urlencoding` dependency may also break the build. These two items alone warrant blocking the PR. The code duplication and missing trait implementations are also important to address before merging.

---
*By kimi[bot] using model `sonnet` via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

PR #57, opened 2026-02-25
fix: TOCTOU race condition in GitHub label updates

### LINKED ISSUE #54: Bug: TOCTOU race condition in GitHub label updates can lose labels
## Summary
The `GitHubBackend::update_status` method has a documented Time-Of-Check-Time-Of-Use (TOCTOU) race condition where labels added between GET and PUT operations are lost.

## Location
`src/backends/github.rs:78-91`

## Current Code
```rust
// GET current labels, swap status:* prefix, PUT the full set.
// The PUT itself is atomic, but there's a TOCTOU window between GET
// and PUT where another process could modify labels. Labels added in
// that window would be lost.
let task = self.get_task(id).await?;
let mut labels: Vec<String> = task
    .labels
    .into_iter()
    .filter(|l| !l.starts_with("status:"))
    .collect();
labels.push(label.to_string());
self.gh.replace_labels(&self.repo, &id.0, &labels).await?;
```

## Risk Assessment
Currently documented as "acceptable for orchestrator (single writer)" but this assumption breaks with:
- Multiple orchestrator instances (horizontal scaling)
- Human edits to issue labels during orchestrator operation
- Future webhook-based label changes

### Option 1: GraphQL Mutation with ETag
Use GitHub's GraphQL API with optimistic concurrency control:
```graphql
mutation UpdateLabels($id: ID!, $labels: [String!]!, $expectedETag: String) {
  updateIssue(input: {id: $id, labelIds: $labels, expectedHeadOid: $expectedETag}) {
    issue { labels { nodes { name } } }
  }
}
```

### Option 2: Retry with Conflict Detection
If PUT returns labels that don't match our intent, retry:
```rust
for attempt in 0..3 {
    let before = self.get_task(id).await?;
    let new_labels = compute_new_labels(&before.labels);
    self.gh.replace_labels(&self.repo, &id.0, &new_labels).await?;
    
    // Verify
    let after = self.get_task(id).await?;
    if labels_match(&new_labels, &after.labels) {
        break; // Success
    }
    // Retry on mismatch
}
```

### Option 3: Idempotent Label Operations
Instead of replacing all labels, use separate add/remove operations:
```rust
// Remove old status labels one by one (idempotent - 404 is ok)
for old_status in task.labels.iter().filter(|l| l.starts_with("status:")) {
    let _ = self.gh.remove_label(&self.repo, &id.0, old_status).await;
}
// Add new status label (idempotent - already exists is ok)
self.gh.add_labels(&self.repo, &id.0, &[label.to_string()]).await?;
```

## Recommendation
Option 3 is simplest and most robust. GitHub's add/remove label endpoints are idempotent, so concurrent modifications won't lose data.

## Files to Modify
- `src/backends/github.rs` - change from replace_labels to add/remove
- May need to update `src/github/cli.rs` if new methods needed

## Testing
1. Create issue with labels ["status:new", "bug"]
2. Concurrently: 
   - Process A: update_status to "in_progress"
   - Process B: add label "priority:high"
3. Verify final labels include: status:in_progress, bug, priority:high

gabrielkoerich:
## Summary

Fixed the TOCTOU (Time-Of-Check-Time-Of-Use) race condition in `GitHubBackend::update_status` where labels added between GET and PUT operations were lost.

## Changes

- Modified `src/backends/github.rs` to use idempotent add/remove operations instead of replace
- Old approach: GET all labels → filter out status:* → add new status → PUT all labels (race condition)
- New approach: Remove old status:* labels one by one (idempotent, 404 OK) → add new status label (idempotent)

## Why This Works

- `remove_label` returns OK even if label doesn't exist (404 is handled)
- `add_labels` returns OK even if label already exists
- Concurrent modifications from multiple orchestrator instances, human edits, or webhook-based changes will no longer lose labels

Closes #54

### DIFF
--- a/src/backends/github.rs
+++ b/src/backends/github.rs
@@ -55,15 +55,13 @@ impl ExternalBackend for GitHubBackend {
         })
     }
 
-    async fn update_status(&self, id: &ExternalId, status: Status) -> anyhow::Result<()> {
-        // Ensure the target status label exists on the repo before assigning it.
-        // Mirrors bash `_gh_ensure_label` — creates the label on first use so
-        // callers never have to pre-create labels manually.  Failures are
-        // tolerated: if creation fails we still attempt the assign below.
-        let label = status.as_label();
+    // update_status uses the default trait implementation (retry loop with
+    // get_task → remove_label → set_labels). Only ensure_status_label is
+    // overridden to auto-create labels on the GitHub repo.
+
+    async fn ensure_status_label(&self, label: &str) -> anyhow::Result<()> {
         let status_name = &label["status:".len()..];
-        if let Err(e) = self
-            .gh
+        self.gh
             .ensure_label(
                 &self.repo,
                 label,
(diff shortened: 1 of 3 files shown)


gabrielkoerich:
## 🔍 Automated Review — Approve

This is a well-structured fix for the TOCTOU race condition. The approach of replacing the GET→filter→PUT pattern with idempotent remove+add operations is sound, and the retry logic adds good resilience. A few observations:

**Positive aspects:**

1. **Correct TOCTOU mitigation** (`src/backends/mod.rs`): The old GET→filter→PUT in `github.rs` had a clear race window where concurrent label additions would be silently dropped. The new remove_label + set_labels approach is idempotent and narrows the race significantly.

2. **Good trait design** (`src/backends/mod.rs:89-159`): Moving `update_status` to a default trait method with an `ensure_status_label` hook is clean. Backends get correct behavior for free while retaining the ability to override for native atomic operations.

3. **Solid retry logic** (`src/backends/mod.rs:109-155`): Fresh snapshot on each attempt, linear backoff (500ms × attempt), and clear error propagation. The `remove_failed` flag correctly skips to the next retry iteration.

4. **Comprehensive test coverage** (`src/backends/mod.rs:222-430`): Tests cover single replacement, multiple old labels, idempotency, non-status label preservation, transient failure recovery, and max retry exhaustion. The MockBackend deliberately does NOT override `update_status`, so the tests exercise the actual production retry logic.

**Minor observations (non-blocking):**

1. **Missing test for `remove_label` retry path** (`src/backends/mod.rs`): The `MockBackend` has a `remove_err` field (line ~243) wired up to simulate `remove_label` failures, but no test exercises it. A test like `update_status_retries_on_remove_label_failure` would complete the coverage of the retry loop. The `set_labels` retry path is tested, but the `remove_label` branch at lines 133-142 is not.

2. **`ensure_status_label` assumes `status:` prefix** (`src/backends/github.rs:62`): `let status_name = &label["status:".len()..];` will panic if the label doesn't start with `status:`. Currently safe because `Status::as_label()` always produces that prefix, but using `label.strip_prefix("status:").unwrap_or(label)` would be more defensive. This is an existing pattern carried forward, not introduced by this PR.

3. **`set_labels` naming ambiguity**: The default `update_status` calls `self.set_labels(id, &[label.to_string()])` (line 147) passing only the target label. This relies on `set_labels` being an **additive** operation (POST, not PUT). The mock confirms this intent (lines 307-316 add without replacing). If any backend implements `set_labels` as a replacement (PUT), this would catastrophically drop all non-status labels. The trait documentation for `set_labels` should clarify that it MUST be additive. This is not a bug per se — it's a documentation/contract concern.

4. **Retry backoff could use jitter**: The fixed `500 * attempt` backoff (line 116) risks thundering herd if multiple orchestrator instances retry simultaneously. Adding small random jitter would help, but this is a minor optimization not critical for correctness.

Overall this is a well-reasoned fix with good test coverage. The approach correctly eliminates the race condition described in #54.

---

PR #58, opened 2026-02-25
fix: capture service now sets is_final=true when session ends

### LINKED ISSUE #53: Bug: Capture service never sets is_final=true, breaking stream termination
## Summary
The `OutputChunk.is_final` field is never set to `true`, which means streaming clients cannot properly detect when a task has completed. The stream can only end via `RecvError::Closed`, which is not a clean termination signal.

## Location
- `src/channels/capture.rs:147-153` - always sets `is_final: false`
- `src/main.rs:161-164` - handles `is_final` but it never fires

## Current Code (capture.rs)
```rust
let chunk = OutputChunk {
    task_id: task_id.clone(),
    content: new_content,
    timestamp: Utc::now(),
    is_final: false,  // ALWAYS false!
};
```

## Expected Behavior
When a tmux session ends (pane_dead=1 or session killed), the capture service should:
1. Push one final chunk with `is_final: true`
2. Unregister the session

## Implementation Sketch
```rust
// In capture_service.tick(), when capture fails due to dead session:
match tmux::capture_pane(&buffer.session).await {
    Ok(current_content) => { /* normal path */ }
    Err(e) if is_session_dead(&buffer.session).await => {
        // Session ended - send final chunk
        let chunk = OutputChunk {
            task_id: task_id.clone(),
            content: String::new(),
            timestamp: Utc::now(),
            is_final: true,
        };
        self.transport.push_output(&task_id, chunk).await;
        self.unregister_session(&task_id).await;
    }
    Err(e) => { /* log error */ }
}
```

## Impact
- `orchestrator task stream <id>` hangs at end until Ctrl+C
- External channel streaming (Discord, Telegram) cannot properly signal completion
- Stream consumers don't know when to disconnect

## Files to Modify
- `src/channels/capture.rs` - detect session end, set is_final
- May need `src/channels/tmux.rs` - add `is_session_dead()` helper

## Testing
1. Start a task
2. Stream output: `orchestrator task stream <id>`
3. Wait for task to complete
4. Should see "--- Stream ended ---" message and exit cleanly

gabrielkoerich:
## Summary

- Fix capture service to set is_final=true when tmux session ends
- Add is_session_dead() helper in tmux.rs
- Modify capture.rs to detect dead sessions and send final chunk

This fixes the issue where `orchestrator task stream <id>` would hang at the end until Ctrl+C.

Closes #53

## Test plan

- [ ] Start a task
- [ ] Stream output: orchestrator task stream <id>
- [ ] Wait for task to complete
- [ ] Should see "--- Stream ended ---" message and exit cleanly

### DIFF
--- a/src/channels/capture.rs
+++ b/src/channels/capture.rs
@@ -162,9 +162,17 @@ impl CaptureService {
                         }
                     }
                     Err(e) => {
-                        // Check if session is dead (no longer exists)
+                        // Check if session is dead (no longer exists).
+                        // This is a best-effort check — `tmux has-session` is spawned
+                        // on every capture failure, including transient ones. For alive
+                        // sessions with transient failures, the overhead is minimal
+                        // (one extra subprocess per poll cycle).
                         if tmux::is_session_dead(&buffer.session).await {
-                            tracing::debug!(task_id, session = buffer.session, "session ended, sending final chunk");
+                            tracing::info!(
+                                task_id,
+                                session = buffer.session,
+                                "session ended, sending final chunk"
+                            );
                             // Send final chunk to signal stream termination
                             let chunk = OutputChunk {
                                 task_id: task_id.clone(),
@@ -173,10 +181,16 @@ impl CaptureService {
                                 is_final: true,
                             };
                             self.transport.push_output(&task_id, chunk).await;
+                            // Unregister immediately — prevents re-processing on next tick
                             self.unregister_session(&task_id).await;
                         } else {
-                            // Session still exists but capture failed - don't spam logs
-                            tracing::trace!(task_id, session = buffer.session, ?e, "capture failed");
+                            // Session still exists but capture failed transiently
+                            tracing::trace!(
+                                task_id,
+                                session = buffer.session,
+                                ?e,
+                                "capture failed (transient)"
+                            );
                         }
                     }
                 }


gabrielkoerich:

The fix correctly identifies the root cause and the approach (using `tmux has-session` to detect dead sessions) is sound. However, there are a few correctness issues worth addressing before merging.

**1. `push_output` return value silently discarded (capture.rs ~line 173)**
`self.transport.push_output(&task_id, chunk).await;` — if `push_output` returns a `Result`, it is being dropped here. If delivery fails (e.g., the subscriber channel has been dropped, back-pressure, or a transport error), the final chunk is silently lost and the stream hangs anyway — the exact bug this PR is meant to fix. At minimum, log an error on failure:
```rust
if let Err(e) = self.transport.push_output(&task_id, chunk).await {
    tracing::warn!(task_id, ?e, "failed to push final chunk");
}
```

**2. Potential double-send if the capture loop continues after `unregister_session` (capture.rs ~line 175)**
After `self.unregister_session(&task_id).await`, if the enclosing loop does not immediately `break` or `return`, the next iteration will again try to capture the dead session, fail, call `is_session_dead()` again, and attempt to push a second final chunk. The diff is truncated so the loop structure isn't visible, but this should be verified. If `unregister_session` removes the session from the polling registry asynchronously (e.g., the removal is queued), the current goroutine/task could iterate again before the removal takes effect. Consider adding an explicit `break` or `return` after the unregister call:
```rust
self.transport.push_output(&task_id, chunk).await;
self.unregister_session(&task_id).await;
break; // or return — ensure we don't re-enter the error path
```

**3. `is_session_dead` spawns a subprocess on every transient capture failure (tmux.rs lines 112–123)**
The `Err(e)` branch is hit whenever `tmux capture-pane` fails for *any* reason — including transient errors (pane resize, brief unavailability). For sessions that are alive but experiencing transient failures, this now spawns an extra `tmux has-session` process on each poll cycle. This is unlikely to be a real problem in practice, but it's worth documenting the intent (this is a best-effort dead-session check, not a guaranteed-atomic detection).

**4. No automated tests for the critical path**
The PR test plan is entirely manual. For a fix to stream termination behavior, at least a unit test for `is_session_dead` (mocking the tmux command) and an integration test asserting that a completed task's stream exits cleanly would meaningfully protect against regressions. This is not a blocker on its own, but combined with the issues above it increases risk.

**Summary**: Issues 1 and 2 are the primary blockers. Issue 1 means the fix can silently fail; issue 2 can cause duplicate state transitions. Both are straightforward to address.

---
*By kimi[bot] using model `sonnet` via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

PR #62, opened 2026-02-25
feat: wire up TaskManager and implement internal task CRUD CLI

### LINKED ISSUE #46: Wire up TaskManager and implement internal task CRUD CLI
## Summary
The `TaskManager` struct exists in `src/engine/tasks.rs` but is not integrated into the engine. It provides a unified interface for both internal (SQLite) and external (GitHub) tasks. This issue tracks wiring it up and exposing CLI commands.

## Current State
TaskManager exists but is marked `#[allow(dead_code)]`:
- `create_task()` - creates internal or external tasks
- `get_task()` - fetches by ID
- `list_tasks()` - filters by status/source
- `update_status()` - unified status updates
- `publish_task()` - promotes internal → external

### 1. Wire TaskManager into Engine (src/engine/mod.rs)
Replace direct backend calls with TaskManager:

```rust
// Current (line 70):
let backend: Arc<dyn ExternalBackend> = Arc::new(GitHubBackend::new(repo.clone()));

// New:
let backend: Arc<dyn ExternalBackend> = Arc::new(GitHubBackend::new(repo.clone()));
let task_manager = Arc::new(TaskManager::new(db.clone(), backend.clone()));
// Use task_manager for all task operations
```

Update tick() to use TaskManager:
- Line 212: `backend.list_by_status(Status::InProgress)` → `task_manager.list_tasks(filter)`
- Line 255: `backend.list_by_status(Status::New)` → same
- Line 336: `backend.list_by_status(Status::Blocked)` → same

### 2. Add CLI Commands (src/main.rs)
Add subcommands for internal task management:

```rust
enum Commands {
    // ... existing commands
    
    /// Internal task management
    Task {
        #[command(subcommand)]
        action: TaskAction,
    },
}

enum TaskAction {
    /// List tasks (internal + external)
    List {
        #[arg(long)]
        status: Option<String>,
        #[arg(long)]
        source: Option<String>,
    },
    /// Create an internal task
    Add {
        title: String,
        #[arg(short, long)]
        body: Option<String>,
        #[arg(short, long)]
        source: Option<String>,
    },
    /// Get task details
    Get {
        id: i64,
    },
    /// Publish internal task to GitHub
    Publish {
        id: i64,
        #[arg(short, long)]
        labels: Vec<String>,
    },
}
```

### 3. Fix Status Mapping Bug (src/engine/tasks.rs:131-141)
Current code has a logic error:
```rust
Status::NeedsReview => TaskStatus::Blocked,  // Wrong! Should be separate variant
```

Options:
- Add `TaskStatus::NeedsReview` variant
- Or map `NeedsReview` to `Blocked` intentionally (document why)

### 4. Implement Missing TaskManager Methods
Add to `src/engine/tasks.rs`:
- `route_task(id, agent)` - assign agent to task
- `block_task(id, reason)` - mark blocked with comment
- `unblock_task(id)` - check children and unblock if done

## Files to Modify
- `src/engine/mod.rs` - wire up TaskManager
- `src/engine/tasks.rs` - fix status mapping, add methods
- `src/main.rs` - add CLI subcommands

## Testing
- Test internal task CRUD
- Test publish flow (internal → external)
- Test unified list with mixed task types

## Documentation
- Document internal vs external tasks in docs/
- Update PLAN.md Internal vs GitHub Tasks section

gabrielkoerich:
## Summary

- Wire up `TaskManager` into the engine loop (`src/engine/mod.rs`)
- Add `TaskStatus::NeedsReview` variant with proper status mapping
- Add `block_reason` column to internal tasks schema
- Implement missing TaskManager methods: `route_task()`, `block_task()`, `unblock_task()`
- Add CLI subcommands for task management: `task list`, `task add`, `task get`, `task publish`
- Improve `list_tasks()` to return both internal and external tasks when filtering

## Test plan

- [x] All 79 tests pass (`cargo test`)
- [x] No clippy warnings (`cargo clippy`)
- [x] Code formatted (`cargo fmt --check`)
- [x] Rebased cleanly onto latest main (resolved conflicts from 10 upstream commits)

Closes #46

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -276,11 +276,36 @@ impl Db {
         Ok(())
     }
 
-    /// Check if all children of a task are done (for unblocking parents)
-    pub async fn are_all_children_done(&self, _parent_id: i64) -> anyhow::Result<bool> {
-        // TODO: Implement when sub-task relationships are added to schema
-        // For now, return true to allow unblocking
-        Ok(true)
+    /// List all internal tasks regardless of status.
+    pub async fn list_all_internal_tasks(&self) -> anyhow::Result<Vec<InternalTask>> {
+        let conn = self.conn.lock().await;
+        let mut stmt = conn.prepare(
+            "SELECT id, title, body, status, source, source_id, agent, block_reason, created_at, updated_at
+             FROM internal_tasks ORDER BY created_at DESC",
+        )?;
+        let tasks = stmt.query_map([], |row| {
+            let status_str: String = row.get(3)?;
+            let created_str: String = row.get(8)?;
+            let updated_str: String = row.get(9)?;
+            Ok(InternalTask {
+                id: row.get(0)?,
+                title: row.get(1)?,
+                body: row.get(2)?,
+                status: TaskStatus::from_str(&status_str).unwrap_or(TaskStatus::New),
+                source: row.get(4)?,
+                source_id: row.get(5)?,
+                agent: row.get(6)?,
+                block_reason: row.get(7)?,
+                created_at: DateTime::parse_from_rfc3339(&created_str)
+                    .map(|dt| dt.with_timezone(&Utc))
+                    .unwrap_or_else(|_| Utc::now()),
+                updated_at: DateTime::parse_from_rfc3339(&updated_str)
+                    .map(|dt| dt.with_timezone(&Utc))
+                    .unwrap_or_else(|_| Utc::now()),
+            })
+        })?;
+        let result: Vec<InternalTask> = tasks.filter_map(|t| t.ok()).collect();
+        Ok(result)
     }
 }
 
--- a/src/engine/tasks.rs
+++ b/src/engine/tasks.rs
@@ -39,7 +39,6 @@ pub struct TaskManager {
     backend: Arc<dyn ExternalBackend>,
 }
 
-#[allow(dead_code)] // Public API methods available for CLI and future engine use
 impl TaskManager {
     pub fn new(db: Arc<Db>, backend: Arc<dyn ExternalBackend>) -> Self {
         Self { db, backend }
@@ -69,10 +68,14 @@ impl TaskManager {
     pub async fn get_task(&self, id: i64) -> anyhow::Result<Task> {
         match self.db.get_internal_task(id).await {
             Ok(internal) => Ok(Task::Internal(internal)),
-            Err(_) => {
+            Err(internal_err) => {
                 let ext_id = ExternalId(id.to_string());
-                let external = self.backend.get_task(&ext_id).await?;
-                Ok(Task::External(external))
+                match self.backend.get_task(&ext_id).await {
+                    Ok(external) => Ok(Task::External(external)),
+                    Err(external_err) => Err(internal_err.context(format!(
+                        "task {id} not found internally or externally (external: {external_err})"
+                    ))),
+                }
             }
         }
     }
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

Overall a solid PR that wires up TaskManager correctly and adds useful CLI commands. However, there are several issues that should be addressed before merging:

## Critical: Schema Migration Gap

**`src/db.rs` (SCHEMA_V1 modification, ~line 270-290)**

The `block_reason` column is added directly to `SCHEMA_V1` rather than as a new migration (e.g., `SCHEMA_V2`). The schema uses `CREATE TABLE IF NOT EXISTS`, which means:
- **Existing databases will NOT get the new `block_reason` column** because the table already exists and `CREATE TABLE IF NOT EXISTS` is a no-op for existing tables.
- All queries now reference `block_reason` (column index 7 in SELECT statements), so any existing database will crash with a `column index out of range` or similar SQLite error on every task query.

**Fix**: Add a `SCHEMA_V2` migration that runs `ALTER TABLE internal_tasks ADD COLUMN block_reason TEXT DEFAULT NULL;` and update the `migrate()` method to apply it.

## Bug: `are_all_children_done()` is a stub that always returns `true`

**`src/db.rs:278-282`**

```rust
pub async fn are_all_children_done(&self, _parent_id: i64) -> anyhow::Result<bool> {
    // TODO: Implement when sub-task relationships are added to schema
    // For now, return true to allow unblocking
    Ok(true)
}
```

This is called by `unblock_task()` in `src/engine/tasks.rs` (~line 318) to guard whether a task can be unblocked. Since it always returns `true`, the guard is meaningless. This is acceptable as a temporary measure **only if the TODO is tracked** — but the PR claims to close #46, so this should either be implemented or the `unblock_task()` method should document this limitation clearly (not just in a code comment, but perhaps as a follow-up issue).

## Bug: `ExternalId` string/i64 mismatch in `route_task`, `block_task`, `unblock_task`

**`src/engine/tasks.rs:290-338`**

In `route_task()`, `block_task()`, and `unblock_task()`, when the internal task lookup fails (falling through to the external branch), the code does:
```rust
let ext_id = ExternalId(id.to_string());
```

The `id` parameter is `i64`, but GitHub issue IDs used as `ExternalId` are typically formatted differently (e.g., they might be GitHub issue numbers). The fallthrough logic assumes that if `get_internal_task(id)` fails, then `id` must be an external task ID — but there's no validation. A non-existent internal task ID would silently attempt operations on a potentially non-existent GitHub issue. Consider:
1. Adding a `get_task()` or existence check for external tasks before operating on them.
2. Or at minimum, propagating the original `get_internal_task` error if the external operation also fails.

## Issue: `list_tasks()` with only source filter is incomplete

**`src/engine/tasks.rs:114-126`**

When only the `source` filter is specified (no status), the code only queries internal tasks with `TaskStatus::New` status:
```rust
let all_internal = self.db.list_internal_tasks_by_status(TaskStatus::New).await?;
```

This means `task list --source cron` would miss tasks in `in_progress`, `blocked`, `done`, or `needs_review` status. The source-only filter should query across all statuses. Consider adding a `list_all_internal_tasks()` method to `Db`, or iterating over all statuses.

## Minor: Remaining `#[allow(dead_code)]` on impl block

**`src/engine/tasks.rs:40`**

The PR removes `#[allow(dead_code)]` from structs/enums (good!) but keeps it on the `impl TaskManager` block with the comment `// Public API methods available for CLI and future engine use`. Now that TaskManager is actually wired into the engine and CLI, many of these methods are no longer dead code. Consider removing this attribute entirely — the compiler will tell you if any methods are actually unused.

## Minor: `init_task_manager()` hardcodes `GitHubBackend`

**`src/main.rs:251-262`**

The CLI's `init_task_manager()` always creates a `GitHubBackend`. If the project ever supports other backends, this will need updating. Not a blocker, but worth noting.

## Summary

The **migration gap is the most critical issue** — it will break existing installations. The `ExternalId` mismatch and incomplete source filter are correctness bugs. The stub `are_all_children_done()` should at least be tracked as a follow-up issue.

---

PR #66, opened 2026-02-26
feat: wire run_with_context + consolidate jobs config

gabrielkoerich:
## Summary

- Wire `run_with_context()` into the engine dispatch loop so task results (status updates + comments) are posted back to GitHub after execution
- Consolidate jobs config from separate `~/.orchestrator/jobs.yml` into the project `.orchestrator.yml` (or global `config.yml`)

## Changes

- **Engine dispatch**: calls `runner.run_with_context(&task, &backend, &tmux, route_result)` instead of `runner.run(&task_id, agent, model)` — this posts status updates and result comments to GitHub after each task completes
- **Jobs config**: `load_jobs`/`save_jobs` now read/write the `jobs` key from `.orchestrator.yml` with serde flatten round-tripping to preserve other config keys
- **CLI**: `orch job *` commands use `resolve_jobs_path()` instead of hardcoded `~/.orchestrator/jobs.yml`

## Test plan

- [x] `cargo build` — clean, zero warnings
- [x] `cargo test` — 86 tests pass
- [x] `cargo clippy -- -D warnings` — zero warnings
- [x] `cargo fmt -- --check` — clean

### DIFF
--- a/src/cli/job.rs
+++ b/src/cli/job.rs
@@ -1,19 +1,11 @@
 use crate::config;
 use crate::engine::jobs::{self, Job, TaskTemplate};
 use anyhow::Context;
-use std::path::PathBuf;
 use std::sync::Arc;
-
-fn jobs_path() -> PathBuf {
-    dirs::home_dir()
-        .unwrap_or_default()
-        .join(".orchestrator")
-        .join("jobs.yml")
-}
 
 /// List scheduled jobs.
 pub fn list() -> anyhow::Result<()> {
-    let path = jobs_path();
+    let path = jobs::resolve_jobs_path();
     let jobs = jobs::load_jobs(&path)?;
 
     if jobs.is_empty() {
@@ -63,7 +55,7 @@ pub fn add(
     job_type: &str,
     command: Option<&str>,
 ) -> anyhow::Result<()> {
-    let path = jobs_path();
+    let path = jobs::resolve_jobs_path();
     let mut jobs = jobs::load_jobs(&path)?;
 
     // Generate ID from title
@@ -112,7 +104,7 @@ pub fn add(
 
 /// Remove a job.
 pub fn remove(id: &str) -> anyhow::Result<()> {
-    let path = jobs_path();
+    let path = jobs::resolve_jobs_path();
     let mut jobs = jobs::load_jobs(&path)?;
 
     let initial_len = jobs.len();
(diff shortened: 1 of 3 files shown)


gabrielkoerich:
## 🔍 Automated Review — Approve

Clean PR that correctly wires `run_with_context` into the engine dispatch loop and consolidates jobs config. The changes are minimal and focused. A few observations:

### Correctness

1. **`src/engine/mod.rs:368-369`** — The `tmux.clone()` and `task.clone()` additions are necessary for the `tokio::spawn` move closure. Correct.

2. **`src/engine/mod.rs:379`** — Replacing `runner.run()` with `runner.run_with_context()` correctly passes the full task context and backend, enabling GitHub status updates and result comments post-execution. The `route_result.as_ref()` borrow is correct.

3. **`src/engine/jobs.rs:89-97` (`save_jobs`)** — The read-modify-write pattern to preserve non-jobs keys is correct. The `#[serde(flatten)]` with `serde_yml::Mapping` is the idiomatic approach for YAML round-tripping. Note that YAML comments will be lost on save, but this is acceptable for a machine-managed config.

4. **`src/engine/runner/mod.rs:417`** — Removing `#[allow(dead_code)]` from `run_with_context` is correct since it's now called from the dispatch loop.

5. **`src/engine/mod.rs:28`** — The `RouteResult` import removal is correct — it was only used to destructure agent/model, which `run_with_context` now handles internally.

### Minor observations (non-blocking)

- **`src/engine/jobs.rs:78-83` (`resolve_jobs_path`)** — Uses `PathBuf::from(".orchestrator.yml")` which is CWD-relative. This works for CLI commands (user runs from project root) and for the engine (which typically has a known CWD), but if the engine ever changes CWD during execution, this could resolve unexpectedly. Consider canonicalizing or documenting this assumption. Not a blocker since the engine's CWD is stable.

- **`src/engine/runner/mod.rs` `_tmux` parameter** — `run_with_context` accepts `_tmux: &Arc<TmuxManager>` but doesn't use it (underscore prefix). The engine now clones `tmux` at `mod.rs:368` solely to pass it here. This is dead data flow — the runner manages tmux internally via `self`. Minor overhead (Arc clone) but clearly reserved for future use.

- **`src/engine/jobs.rs:89-97` (save_jobs TOCTOU)** — There's a time-of-check-to-time-of-use window between reading and writing the config file. For a config file written by a single process this is fine, but worth noting if concurrent writers become possible.

All test plans (build, test, clippy, fmt) pass per the PR description. The 86 existing tests cover the jobs round-trip serialization. The changes are well-scoped and match the PR description.

---

PR #79, opened 2026-02-26
Integrated security module into response posting flow - scans and redacts secrets before posting to GitHub

### LINKED ISSUE #71: Integrate security module into response posting flow
## Problem
The security module (`src/security.rs`) is fully implemented with leak detection for API keys, tokens, passwords, private keys, and connection strings - but it's not wired into the response posting flow. Agent output gets posted to GitHub without scanning for secrets.

## Current State
- `src/security.rs` - Complete implementation with tests
- `src/engine/runner/mod.rs` - Posts comments to GitHub via backend
- NOT connected - comments are posted without scanning

## Solution
Integrate security scanning into the response posting flow:

1. In `src/engine/runner/mod.rs`, before calling `backend.post_comment()`, scan the response
2. Use `security::scan()` to detect leaks
3. If leaks found, use `security::redact()` to sanitize before posting
4. Optionally: Add a label or comment to warn about redacted content

Example location in runner/mod.rs around line 458:
```rust
// Before posting comment
let comment = if !summary.is_empty() {
    format!("[{now}] {status}: {summary}")
    // ... scan and redact here
};
backend.post_comment(&task.id, &comment).await?;
```

## Files to modify
- `src/engine/runner/mod.rs` - Integrate security scan before posting comments

## Priority
High - Prevents accidentally leaking secrets to public GitHub issues

## Notes
The security module is well-tested. Just needs to be wired in.

gabrielkoerich:
## Summary

Integrated security module into response posting flow - scans and redacts secrets before posting to GitHub

### What was done

- Made security module public in main.rs
- Added security import to runner/mod.rs
- Integrated security::has_leaks() scan before posting comments
- Added security::redact() to sanitize leaked secrets
- Added warning notice when secrets are detected and redacted
- All 105 tests pass
- No clippy warnings

### Remaining

- Approve git push to publish changes

### Files changed

- `src/main.rs`
- `src/engine/runner/mod.rs`

Closes #71

---
*Created by minimax[bot] via [Orchestrator](https://github.com/gabrielkoerich/orchestrator)*

### DIFF
--- a/src/engine/runner/mod.rs
+++ b/src/engine/runner/mod.rs
@@ -18,6 +18,7 @@ pub mod worktree;
 use crate::backends::{ExternalBackend, ExternalId, ExternalTask, Status};
 use crate::config;
 use crate::engine::router::{get_route_result, RouteResult};
+use crate::security;
 use crate::sidecar;
 use crate::tmux::TmuxManager;
 use response::RunResult;
@@ -454,14 +455,29 @@ impl TaskRunner {
         };
         backend.update_status(&task.id, new_status).await?;
 
-        // Post comment
+        // Post comment (scan for secrets before posting to GitHub)
         let now = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ");
-        let comment = if !summary.is_empty() {
+        let raw_comment = if !summary.is_empty() {
             format!("[{now}] {status}: {summary}")
         } else if !last_error.is_empty() {
             format!("[{now}] {status}: {last_error}")
         } else {
             format!("[{now}] {status}")
+        };
+
+        // Scan for leaked secrets and redact if needed
+        let comment = if security::has_leaks(&raw_comment) {
+            let leaks = security::scan(&raw_comment);
+            let rules: Vec<&str> = leaks.iter().map(|l| l.rule).collect();
+            let warning = format!(
+                "\n\n> ⚠️ **Security Notice**: {} potential secret(s) detected and redacted: {}",
+                leaks.len(),
+                rules.join(", ")
+            );
+            let redacted = security::redact(&raw_comment);
+            format!("{redacted}{warning}")
+        } else {
+            raw_comment
         };
         backend.post_comment(&task.id, &comment).await?;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,7 @@ mod engine;
 mod github;
 mod parser;
 #[allow(dead_code)] // security module ready for use, not integrated yet
-mod security;
+pub mod security;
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

The security integration logic is sound in principle — scanning and redacting before posting comments is the right approach. However, there are several issues that should be addressed before merging:

### 1. Stale `#[allow(dead_code)]` attribute and misleading comment — `src/main.rs`

The diff shows that the `#[allow(dead_code)]` attribute and its comment `// security module ready for use, not integrated yet` are preserved on line 12. This is now factually wrong — the module IS integrated. The attribute should be removed entirely along with the comment.

### 2. `pub mod security` is unnecessary — `src/main.rs`

In Rust, a module declared at the crate root (`main.rs`) with `mod security` (private) is already accessible from anywhere in the crate via `crate::security`. Making it `pub` unnecessarily exposes it outside the crate boundary. While this is harmless for a binary crate (no external consumers), it signals the wrong intent. The fix should be:
```rust
mod security;
```
(just remove the `#[allow(dead_code)]` and its comment, keep `mod` as-is)

### 3. Double scanning is wasteful — `src/engine/runner/mod.rs:463-472`

`security::has_leaks(&raw_comment)` is called first, then `security::scan(&raw_comment)` is called again inside the `if` block. This scans the content twice. Instead, call `scan()` once and check if the result is empty:

```rust
let leaks = security::scan(&raw_comment);
let comment = if !leaks.is_empty() {
    let rules: Vec<&str> = leaks.iter().map(|l| l.rule).collect();
    let warning = format!(
        "\n\n> ⚠️ **Security Notice**: {} potential secret(s) detected and redacted: {}",
        leaks.len(),
        rules.join(", ")
    );
    let redacted = security::redact(&raw_comment);
    format!("{redacted}{warning}")
} else {
    raw_comment
};
```

### 4. No integration tests for the new security scanning path

The PR states '105 tests pass' but doesn't add any new tests for this integration. Given this is security-critical code (preventing secret leakage to GitHub), there should be at least:
- A test verifying that a comment containing a secret is redacted before posting
- A test verifying that a clean comment is posted unchanged
- A test verifying the warning notice format when secrets are detected

Without being able to see the full security module source, I cannot verify the correctness of the `has_leaks()`, `scan()`, and `redact()` function signatures and return types, but the usage pattern appears consistent.

### Summary
The approach is correct — scan before post is the right pattern. The fixes are straightforward: remove the stale attribute/comment, drop the unnecessary `pub`, eliminate the double scan, and add integration tests. None of these are architectural concerns; they're cleanup that should be done before merging security-sensitive code.

---

PR #80, opened 2026-02-26
fix: flatten nested .orchestrator/.orchestrator to .orchestrator/state

### LINKED ISSUE #76: Eliminate nested .orchestrator/.orchestrator state directory
## Context

There is a confusing nested directory pattern: `~/.orchestrator/.orchestrator/` is used for runtime state (sidecar files, prompts, stderr, exit codes, runner scripts, logs). This was inherited from the bash implementation.

## Task

Flatten the state directory structure:

1. **Sidecar files**: `~/.orchestrator/.orchestrator/{id}.json` → `~/.orchestrator/state/{id}.json`
2. **Prompt files**: `~/.orchestrator/.orchestrator/prompt-*` → `~/.orchestrator/state/prompt-*`
3. **Runner scripts**: `~/.orchestrator/.orchestrator/runner-*` → `~/.orchestrator/state/runner-*`
4. **Exit codes**: `~/.orchestrator/.orchestrator/exit-*` → `~/.orchestrator/state/exit-*`
5. **Stderr**: `~/.orchestrator/.orchestrator/stderr-*` → `~/.orchestrator/state/stderr-*`

Files to update:
- `src/sidecar.rs` — base directory
- `src/engine/runner/agent.rs` — `build_runner_script()` and `spawn_in_tmux()` state_dir
- `src/engine/runner/response.rs` — stderr and output paths
- `src/engine/runner/mod.rs` — status_file path
- `src/cli/service.rs` — PID file path

## Acceptance criteria

- No more nested `.orchestrator/.orchestrator/` pattern
- State files live in `~/.orchestrator/state/` (or just `~/.orchestrator/` flat)
- Backward compat: check old path if new path doesn't exist
- All 105 tests pass

gabrielkoerich:
## Summary

- Replaces confusing nested `~/.orchestrator/.orchestrator/` directory with `~/.orchestrator/state/` for all runtime state files
- Adds `state_dir()` and `state_file()` public helpers to `sidecar.rs` with backward-compatible fallback to legacy paths
- Updates all 5 consumer modules: `engine/runner/agent.rs`, `engine/runner/response.rs`, `engine/runner/mod.rs`, `cli/service.rs`
- Updates documentation in `AGENTS.md` and `docs/content/cli.md`

## Test plan

- [x] All 105 tests pass (`cargo test`)
- [x] No clippy warnings (`cargo clippy`)
- [x] Code formatted (`cargo fmt`)
- [x] Backward compat: `state_file()` checks new path first, falls back to legacy `~/.orchestrator/.orchestrator/` for reads

Closes #76

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,7 +27,7 @@ orch task unblock all
 
 ## Logs
 
-- Service log: `~/.orchestrator/.orchestrator/orch.log`
+- Service log: `~/.orchestrator/state/orch.log`
 - Brew stdout: `/opt/homebrew/var/log/orch.log` (startup messages only)
 - Brew stderr: `/opt/homebrew/var/log/orch.error.log`
 
@@ -143,7 +143,7 @@ Override the router by adding labels to tasks:
 
 ### RouteResult Struct
 
-Routing results are stored in the sidecar file (`~/.orchestrator/.orchestrator/{task_id}.json`):
+Routing results are stored in the sidecar file (`~/.orchestrator/state/{task_id}.json`):
 
 ```rust
 pub struct RouteResult {
@@ -194,7 +194,7 @@ The routing prompt template is at `prompts/route.md`. It includes:
     owner/repo.git       #   each has .orchestrator.yml inside
   worktrees/             # agent worktrees (all projects)
     repo/branch/         #   created by the runner, one per task
-  .orchestrator/         # runtime state (logs, prompts, pid, locks)
+  state/                 # runtime state (logs, prompts, pid, locks)
 ```
 
 - **User-managed projects** (e.g. `~/Projects/foo`): user clones, runs `orch init`. Project dir stays where the user put it.
--- a/docs/content/cli.md
+++ b/docs/content/cli.md
@@ -110,12 +110,12 @@ orch agents                 # list available agents and their status
 
 | Log | Location |
 |-----|----------|
-| Server log | `~/.orchestrator/.orchestrator/orchestrator.log` |
-| Server archive | `~/.orchestrator/.orchestrator/orchestrator.archive.log` |
-| Jobs log | `~/.orchestrator/.orchestrator/jobs.log` |
-| Per-task output | `~/.orchestrator/.orchestrator/output-{id}.json` |
-| Per-task tools | `~/.orchestrator/.orchestrator/tools-{id}.json` |
-| Per-task prompts | `~/.orchestrator/.orchestrator/prompt-{id}.md` |
+| Server log | `~/.orchestrator/state/orchestrator.log` |
+| Server archive | `~/.orchestrator/state/orchestrator.archive.log` |
+| Jobs log | `~/.orchestrator/state/jobs.log` |
+| Per-task output | `~/.orchestrator/state/output-{id}.json` |
+| Per-task tools | `~/.orchestrator/state/tools-{id}.json` |
+| Per-task prompts | `~/.orchestrator/state/prompt-{id}.md` |
 | Task context | `~/.orchestrator/contexts/task-{id}.md` |
 | Brew stdout | `/opt/homebrew/var/log/orchestrator.log` |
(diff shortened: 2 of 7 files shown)


gabrielkoerich:

## Review of PR #80: flatten nested .orchestrator/.orchestrator to .orchestrator/state

Overall this is a well-structured migration with a clean backward-compat strategy. However, there are a few issues that should be addressed before merging.

### Critical: Likely type mismatch in `response.rs`

**`src/engine/runner/response.rs`** — In the `read_output_file` function, the diff shows:
```rust
if let Some(legacy_path) = sidecar::state_file(&format!("output-{task_id}.json")) {
```

`sidecar::state_file()` returns `anyhow::Result<PathBuf>`, not `Option<PathBuf>`. This should be `if let Ok(legacy_path)`, not `if let Some(legacy_path)`. In Rust, `if let Some(...)` only destructures `Option<T>`, not `Result<T, E>`. If the diff is accurate, this would be a compilation error. The PR claims all 105 tests pass — please double-check the actual source code. If this is a diff rendering artifact, disregard.

### Medium: Inconsistent fallback paths across call sites

When `sidecar::state_file()` or `sidecar::state_dir()` returns an error (e.g., `dirs::home_dir()` fails), different call sites fall back to different locations:

| Call site | Fallback |
|-----------|----------|
| `cli/service.rs:pid_file()` | `~/.orchestrator/state/orch.pid` (via `dirs::home_dir()` again — would give the same error) |
| `engine/runner/agent.rs:build_runner_script()` | `~/.orchestrator/state` (also via `dirs::home_dir()`) |
| `engine/runner/mod.rs` | `self.orch_home.join("state")` (uses stored field — good) |
| `engine/runner/response.rs:collect_response()` | `/tmp/stderr-{task_id}.txt` |
| `engine/runner/response.rs:read_output_file()` | `/tmp` |

**The `/tmp` fallbacks in `response.rs` are a behavior change.** The old code fell back to a relative path (via `unwrap_or_default()`), which was bad, but silently switching to `/tmp` could cause confusion since stderr/output files wouldn't be found there. Consider using a consistent fallback strategy — for example, always falling back to `self.orch_home.join("state")` where available, or returning an error rather than silently looking in `/tmp`.

### Medium: `state_dir()` creates directories as a side effect during reads

**`src/sidecar.rs`** — `state_file()` calls `state_dir()` which calls `std::fs::create_dir_all()`. This means every path *lookup* (even for files that may not exist yet) creates `~/.orchestrator/state/` as a side effect. This was the same in the old code, but since you're refactoring this area, consider splitting into:
- `state_dir()` (creates dir, for writes)
- `state_dir_path()` (returns path only, for reads)

This is a suggestion, not a blocker.

### Minor: No unit tests for new `state_dir()` / `state_file()` / `legacy_state_dir()`

These three functions are the core of the migration logic, especially the fallback behavior in `state_file()`. The existing 105 tests pass, but there are no new tests specifically validating:
- `state_file()` returns the new path when a file exists at the new location
- `state_file()` returns the legacy path when a file only exists at the legacy location
- `state_file()` returns the new path (for writes) when the file doesn't exist anywhere
- `legacy_state_dir()` returns `None` when the legacy dir doesn't exist

Adding a `#[cfg(test)] mod tests` block in `sidecar.rs` with these cases would give confidence the migration logic works correctly.

### Minor: Missing `use` imports not shown in diff context

The diff for `response.rs` and `runner/mod.rs` doesn't show `use crate::sidecar;` being added, though both files now call `sidecar::state_file()` and `sidecar::state_dir()`. This is likely just outside the diff context window — please confirm the imports are present.

### Positive Notes

- The `state_file()` fallback design (check new → check legacy → return new for writes) is a clean migration pattern
- `legacy_state_dir()` only returns `Some` if the directory actually exists — correct
- Documentation updates in `AGENTS.md` and `docs/content/cli.md` are consistent with code changes
- The PR correctly updates all 5 consumer modules

---

PR #93, opened 2026-02-26
feat: add axum webhook HTTP server for GitHub events

gabrielkoerich:
## Summary

Implements an Axum-based HTTP webhook server for receiving GitHub events in real-time, bypassing the polling mechanism.

- Added Axum HTTP server listening on configurable port (default 8080)
- Handles GitHub webhook payloads: `issues.opened`, `issues.labeled`, `pull_request.*`, `issue_comment.*`
- HMAC-SHA256 signature verification for security
- Config: `webhook.port: 8080` and `webhook.secret: "..."` in config
- Forwards events to engine via channel (bypass polling delay)
- Webhook server starts alongside engine in `orch serve`
- Added comprehensive tests for webhook parsing and signature verification

### Configuration

Add to `~/.orchestrator/config.yml`:

```yaml
webhook:
  port: 8080
  secret: "your-webhook-secret"
```

### Endpoints

- `GET /health` - Health check
- `POST /webhook` - GitHub webhook endpoint

## Testing

All 128 tests pass, including 11 new tests for webhook functionality.

### DIFF
--- a/src/channels/github.rs
+++ b/src/channels/github.rs
@@ -475,11 +475,7 @@ pub async fn start_webhook_server(
 ) -> anyhow::Result<()> {
     let listener = tokio::net::TcpListener::bind(format!("0.0.0.0:{}", port)).await?;
 
-    let state = WebhookState {
-        secret,
-        repo,
-        tx,
-    };
+    let state = WebhookState { secret, repo, tx };
 
     let app = Router::new()
         .route("/health", get(webhook_health))
@@ -558,8 +554,12 @@ mod tests {
                 number: 42,
                 title: "Test issue".to_string(),
                 labels: vec![
-                    LabelPayload { name: "bug".to_string() },
-                    LabelPayload { name: "urgent".to_string() },
+                    LabelPayload {
+                        name: "bug".to_string(),
+                    },
+                    LabelPayload {
+                        name: "urgent".to_string(),
+                    },
                 ],
             }),
             pr: None,
@@ -660,7 +660,11 @@ mod tests {
                 user: GhUser {
                     login: "testuser".to_string(),
                 },
-                created_at: Some(chrono::DateTime::parse_from_rfc3339("2024-01-15T10:30:00Z").unwrap().with_timezone(&chrono::Utc)),
+                created_at: Some(
+                    chrono::DateTime::parse_from_rfc3339("2024-01-15T10:30:00Z")
+                        .unwrap()
+                        .with_timezone(&chrono::Utc),
+                ),
             }),
         };
 


gabrielkoerich:

This PR has several significant issues — two critical bugs and one security vulnerability — that need to be addressed before merging.

## Critical Bugs

### 1. Signature verification uses re-serialized body, not raw bytes (Security - Broken HMAC)

**`src/channels/github.rs` — `handle_webhook` function**

The handler accepts `axum::extract::Json<WebhookPayload>`, which means Axum has already deserialized the raw HTTP body into a struct. Then the code re-serializes it to verify the signature:

```rust
async fn handle_webhook(
    State(state): State<WebhookState>,
    headers: axum::http::HeaderMap,
    raw_body: axum::extract::Json<WebhookPayload>,  // already parsed
) -> impl IntoResponse {
    let payload = raw_body.0;
    // ...
    let body_bytes = serde_json::to_vec(&payload).unwrap_or_default();  // re-serialized!
    if !state.secret.is_empty() && !verify_signature(&state.secret, &body_bytes, signature) {
```

GitHub computes the HMAC-SHA256 over the **raw HTTP request body bytes**. Re-serializing via `serde_json::to_vec()` will produce different bytes (different whitespace, key ordering, fields not captured by the struct are lost), so **signature verification will effectively never pass for legitimate GitHub webhooks**.

**Fix:** Accept raw `bytes::Bytes` first, verify the signature against raw bytes, then deserialize:
```rust
async fn handle_webhook(
    State(state): State<WebhookState>,
    headers: axum::http::HeaderMap,
    body: bytes::Bytes,
) -> impl IntoResponse {
    let signature = headers.get("x-hub-signature-256")...;
    if !state.secret.is_empty() && !verify_signature(&state.secret, &body, signature) {
        return (StatusCode::UNAUTHORIZED, "Invalid signature");
    }
    let payload: WebhookPayload = match serde_json::from_slice(&body) {
        Ok(p) => p,
        Err(_) => return (StatusCode::BAD_REQUEST, "Invalid payload"),
    };
    // ...
}
```

### 2. Deadlock: webhook messages are never forwarded to transport

**`src/engine/mod.rs` — lines ~275-300**

The spawned task creates a channel, starts the webhook server, and then tries to forward messages:

```rust
Some(tokio::spawn(async move {
    let (tx, mut rx) = tokio::sync::mpsc::channel::<IncomingMessage>(64);

    // This blocks forever — axum::serve().await runs the server loop
    if let Err(e) = start_webhook_server(port, secret, webhook_repo, tx).await {
        tracing::error!(?e, "webhook server failed");
    }

    // This code NEVER executes
    while let Some(msg) = rx.recv().await {
        let _ = webhook_tx.route(&msg).await;
    }
}))
```

`start_webhook_server` calls `axum::serve(listener, app).await?` which blocks forever (it's the server event loop). The `while let Some(msg) = rx.recv().await` loop below **never runs**. Messages pile up in the channel buffer (capacity 64) and then `tx.send()` starts failing.

**Fix:** Spawn the server and the forwarder as separate tasks, or use `tokio::select!`, or have the webhook handler send directly to the transport:
```rust
Some(tokio::spawn(async move {
    let (tx, mut rx) = tokio::sync::mpsc::channel::<IncomingMessage>(64);

    tokio::spawn(async move {
        while let Some(msg) = rx.recv().await {
            let _ = webhook_tx.route(&msg).await;
        }
    });

    if let Err(e) = start_webhook_server(port, secret, webhook_repo, tx).await {
        tracing::error!(?e, "webhook server failed");
    }
}))
```

## Security Issues

### 3. Non-constant-time HMAC comparison — timing attack vulnerability

**`src/channels/github.rs` — `verify_signature` function**

```rust
fn verify_signature(secret: &str, payload: &[u8], signature: &str) -> bool {
    // ...
    let expected = format!("sha256={:x}", result);
    expected == signature  // ← regular string comparison, not constant-time
}
```

Using `==` for HMAC comparison leaks timing information and is vulnerable to timing attacks. The `hmac` crate provides `mac.verify_slice()` which uses `subtle::ConstantTimeEq` for exactly this purpose.

**Fix:**
```rust
fn verify_signature(secret: &str, payload: &[u8], signature: &str) -> bool {
    let sig_hex = signature.strip_prefix("sha256=").unwrap_or("");
    let sig_bytes = match hex::decode(sig_hex) {
        Ok(b) => b,
        Err(_) => return false,
    };
    let mut mac = HmacSha256::new_from_slice(secret.as_bytes()).unwrap();
    mac.update(payload);
    mac.verify_slice(&sig_bytes).is_ok()
}
```

## Other Issues

### 4. `issue_comment` handler loses issue context

**`src/channels/github.rs` — `parse_github_event`, `issue_comment` branch**

```rust
thread_id: "0".to_string(),
```

The `thread_id` is hardcoded to `"0"`, losing the association with the actual issue or PR. GitHub's `issue_comment` webhook payload includes the `issue` object with the issue number. The `WebhookPayload` struct already has an `issue: Option<IssuePayload>` field — use it:

```rust
thread_id: payload.issue.as_ref()
    .map(|i| i.number.to_string())
    .unwrap_or_else(|| "0".to_string()),
```

### 5. Test `test_verify_signature_valid` doesn't test the positive case

Despite the name, this test passes an invalid signature and asserts failure — the same thing `test_verify_signature_invalid` does. There is no test that computes a correct HMAC and verifies `verify_signature` returns `true`. Add a test that generates a real signature:

```rust
#[test]
fn test_verify_signature_valid() {
    let secret = "test-secret";
    let payload = br#"{"action":"opened"}"#;
    let mut mac = HmacSha256::new_from_slice(secret.as_bytes()).unwrap();
    mac.update(payload);
    let result = mac.finalize().into_bytes();
    let signature = format!("sha256={:x}", result);

    assert!(verify_signature(secret, payload, &signature));
}
```

### 6. Duplicate `tower-http` versions

**`Cargo.toml`** adds `tower-http = "0.5"` while `reqwest` already depends on `tower-http 0.6.8`. This compiles two versions of the crate. Consider using `tower-http = "0.6"` with the `"trace"` feature to avoid the duplicate.

### 7. Redundant `action` field in `PullRequestPayload`

`PullRequestPayload` has an `action: Option<String>` field, but in GitHub's webhook schema, `action` is at the top level of the payload, not nested inside `pull_request`. This field is misleading (it will always deserialize as `None` from real payloads). The tests set it manually, masking this issue.

## Summary

Items 1-3 are blocking: the signature verification is fundamentally broken (verifies against wrong bytes, and uses non-constant-time comparison), and the message forwarding deadlock means webhook events are silently dropped. These need to be fixed before this feature can work correctly or securely.

---

PR #96, opened 2026-02-26
feat: add token budget and cost tracking per task

gabrielkoerich:
## Summary
- Added token fields (input_tokens, output_tokens) to AgentResponse in parser.rs
- Extract token counts from agent response JSON (supports input_tokens, output_tokens, tokens_input, tokens_output, and usage object)
- Store token/cost data in sidecar after agent run with automatic cost calculation
- Added max_tokens_per_task config check to enforce budget (defaults to 100,000)
- Added `orch task cost <id>` command for cost breakdown
- Added cost summary to `orch task status --json` output (total cost, tokens)

## Test plan
- [x] Build passes
- [x] All 120 tests pass
- [x] New CLI command `orch task cost <id>` available

### DIFF
--- a/src/cli/task.rs
+++ b/src/cli/task.rs
@@ -437,7 +437,8 @@ pub fn cost(id: &str) -> anyhow::Result<()> {
 
     if usage.total_tokens() == 0 {
         println!("No token data available for task #{}", id);
-        return Ok(println!("Run the task first to collect token usage."));
+        println!("Run the task first to collect token usage.");
+        return Ok(());
     }
 
     println!("Token usage for task #{}", id);


gabrielkoerich:

## Overall
The feature is well-structured and the code is clean. However, there are several issues that should be addressed before merging.

---

### 1. Bug: Token counts are overwritten, not accumulated across multiple agent runs
**File:** `src/sidecar.rs`, `store_token_usage()` (lines ~180-210)

`store_token_usage` calls `set()` which writes the raw `input_tokens` and `output_tokens` values. If a task involves multiple agent iterations (the runner loop), each call to `store_token_usage` will **overwrite** the previous token counts rather than accumulating them. This means only the **last** agent run's tokens are recorded, and the budget check in `runner/mod.rs` (line ~310) compares `total_tokens` against `max_tokens` using only the latest run's data — not the cumulative total.

**Fix:** Either accumulate tokens (read existing + add new), or store per-run entries and sum on read. The simplest fix:
```rust
pub fn store_token_usage(...) -> anyhow::Result<()> {
    let existing = get_token_usage(task_id);
    let cumulative_input = existing.input_tokens + input_tokens;
    let cumulative_output = existing.output_tokens + output_tokens;
    set(task_id, &[
        format!("input_tokens={}", cumulative_input),
        format!("output_tokens={}", cumulative_output),
    ])?;
    // ... recalculate cost with cumulative values
}
```

---

### 2. Bug: `unwrap_or` on `from_f64` for NaN/Infinity
**File:** `src/cli/task.rs`, line ~187

```rust
serde_json::Number::from_f64(total_cost).unwrap_or(serde_json::Number::from(0))
```

`from_f64` returns `None` for NaN and Infinity. While unlikely with normal arithmetic, if sidecar data is corrupted or manually edited, this could silently produce `0` in the JSON output, masking real data. This is minor but worth a comment or a log warning.

---

### 3. Duplicate iteration over tasks in `status()`
**File:** `src/cli/task.rs`, lines ~163-175

The original code already iterates over all statuses to build `counts`. The new code adds a **second** loop over the same statuses and tasks (lines 167-175) just to gather token data. This doubles the number of `list_by_status` calls (which are async and likely hit storage). Merge the two loops:

```rust
for s in &statuses {
    let tasks = backend.list_by_status(*s).await?;
    for task in &tasks {
        total_input_tokens += sidecar::get_u64(&task.id.0, "input_tokens");
        // ...
    }
    counts.push((s, tasks.len()));
}
```

---

### 4. Incomplete OpenAI usage object parsing
**File:** `src/parser.rs`, `extract_usage_tokens()` (lines ~133-137)

The function only looks for `input_tokens`/`output_tokens` inside the `usage` object. OpenAI's API returns `prompt_tokens` and `completion_tokens` as keys. Since the PR description claims support for usage objects from OpenAI-compatible APIs, these keys should also be handled:

```rust
fn extract_usage_tokens(usage: Option<&serde_json::Value>, is_input: bool) -> Option<u64> {
    let obj = usage?.as_object()?;
    let key = if is_input { "input_tokens" } else { "output_tokens" };
    let alt_key = if is_input { "prompt_tokens" } else { "completion_tokens" };
    obj.get(key).and_then(|v| v.as_u64())
        .or_else(|| obj.get(alt_key).and_then(|v| v.as_u64()))
}
```

---

### 5. Hardcoded model pricing will silently become stale
**File:** `src/sidecar.rs`, `pricing_for_model()` (lines ~230-280)

Hardcoding prices for models like `gpt-5.3-codex`, `gpt-5.2`, etc. means prices will silently drift from reality as providers update pricing. Consider:
- Adding a comment noting these are approximate and when they were last updated
- Supporting a config override (e.g., `config::get("pricing.<model>")`) so users can set their own rates
- At minimum, log a warning when falling back to the default pricing for an unknown model

---

### 6. Unnecessary clone in runner
**File:** `src/engine/runner/mod.rs`, line ~184

```rust
let model_for_invocation = model_name.clone();
```

This clone appears to exist solely because `model_name` is used later (line ~295: `model_name.as_deref()`). If `model_name` is `Option<String>`, you can use `.as_deref()` or `.as_ref()` at the invocation site instead of cloning the whole value. This is a minor nit but the variable name `model_for_invocation` adds unnecessary indirection.

---

### 7. Token budget check happens after storing — no rollback
**File:** `src/engine/runner/mod.rs`, lines ~295-320

The budget check runs **after** the agent has already completed its work and tokens are stored. The task is moved to `needs_review` but the work (and cost) has already been incurred. This is fine as a soft limit, but the behavior should be documented — users might expect `max_tokens_per_task` to prevent spending, not just flag it after the fact. Consider adding a pre-flight check before invoking the agent if cumulative tokens are already near the budget.

---

### 8. Missing test for `store_token_usage` accumulation and sidecar integration
There are unit tests for pricing lookup and cost estimation math, which is good. However, there are no tests for:
- `store_token_usage` writing and reading back correctly
- The accumulation bug (issue #1) — a test calling `store_token_usage` twice and verifying totals would catch this
- `get_token_usage` / `get_cost_estimate` round-tripping

---

### Summary
The most critical issue is **#1 (token overwrite bug)** — it makes the budget enforcement and cost tracking fundamentally incorrect for multi-iteration tasks. **#3 (duplicate iteration)** is a performance issue that's easy to fix. **#4 (missing OpenAI keys)** is a correctness gap for the claimed compatibility. The rest are improvements that would strengthen the feature.

---

PR #162, opened 2026-02-27
feat: PR review comments → re-dispatch agent to fix changes

### LINKED ISSUE #151: feat: PR review comments → re-dispatch agent to fix changes
## Summary

When a human reviewer requests changes on a PR via GitHub's review UI, automatically re-dispatch the original agent to address the feedback — on the **same worktree/branch**.

## Current State

`review_open_prs()` in the sync tick creates internal follow-up tasks with `pr-review-followup` label, but these are disconnected from the original worktree/branch/PR. The agent starts fresh instead of fixing the existing code.

## Proposed Change

Instead of creating new internal tasks, **re-dispatch the original external task** with review context injected into the agent prompt:

1. Detect `CHANGES_REQUESTED` reviews on PRs for `status:in_review` tasks
2. Store review comments in sidecar as `pr_review_context`
3. Set task back to `status:routed` (same agent, same worktree, same branch)
4. Agent sees review feedback in `build_agent_message()` and fixes the issues
5. Agent pushes fixes to same PR branch → review cycle repeats

## Implementation

See `specs.md` §2 for full spec.

Files to modify:
- `src/engine/runner/context.rs` — add `pr_review_context` to `TaskContext`
- `src/engine/runner/agent.rs` — include review context in agent message
- `src/engine/mod.rs` — change `review_open_prs()` to re-dispatch instead of creating internal tasks

Labels: `enhancement`, `priority:critical`

gabrielkoerich:
## Summary

When a human reviewer requests changes on a PR via GitHub's review UI, automatically re-dispatch the original agent to address the feedback — on the **same worktree/branch**.

## Changes

- Add `pr_review_context` to `TaskContext` in `context.rs`
- Include review context in agent message (`agent.rs`)
- Modify `review_open_prs()` to store review context in sidecar and set task status to "routed" instead of creating internal follow-up tasks
- Remove unused `create_review_follow_up_task` and `get_existing_review_tasks` functions
- Track `last_review_ts` to avoid re-processing the same reviews

The agent now sees review feedback in `build_agent_message()` and fixes the issues, then pushes fixes to the same PR branch.

## Test plan

- [x] Code compiles (`cargo check`)
- [x] Tests pass (`cargo test`)

Closes #151

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -1340,7 +1340,6 @@ async fn review_open_prs(
     }
 
     // Check if we should process reviews
-    let auto_create_followup = config.auto_create_followup_on_changes;
     let auto_close_task = config.auto_close_task_on_approval;
 
     tracing::info!(


gabrielkoerich:

## Overall Assessment

The PR's intent is sound — replacing per-comment follow-up tasks with a simpler re-dispatch model that feeds all review feedback to the original agent. This is a good architectural simplification. However, there are several issues that should be addressed before merging.

## Issues

### 1. Unused variable `_original_agent` — dead code or missing routing logic?
**File:** `src/engine/mod.rs`, around line 1335

The variable `original_agent` was renamed to `_original_agent` (prefixed with underscore to suppress the unused warning), but it's no longer used anywhere. The old code used it to route follow-up tasks to the same agent. With the new re-dispatch approach, since the task is being set back to `Status::Routed`, the question is: **does the routing pipeline re-use the existing agent assignment, or does it re-route from scratch?**

If `Status::Routed` triggers a fresh routing decision, the task might get assigned to a *different* agent than the one that originally worked on it, which would be wrong — the new agent won't have the context of the existing worktree/branch.

**Suggestion:** Either verify that re-dispatching to `Routed` preserves the original agent assignment (and document this), or explicitly re-assign the agent before changing the status. Remove the `_original_agent` variable entirely if truly unused.

### 2. `_db` parameter — is `db` truly unused now?
**File:** `src/engine/mod.rs`, line 1314

The `db` parameter is renamed to `_db`. This is fine if `db` is genuinely no longer needed, but double-check that all callers won't need it. The old code used `db` for `get_existing_review_tasks` and `create_review_follow_up_task`. Since those are removed, this is likely correct, but make sure no other `db` usage exists in this function (e.g., logging, metrics).

### 3. Review context stored as raw string in sidecar — no size limit
**File:** `src/engine/mod.rs`, around line 1425

```rust
sidecar::set(task_id, &[format!("pr_review_context={}", review_context)])
```

The `review_context` string is built by concatenating all `CHANGES_REQUESTED` reviews, their bodies, all actionable comments with full diff hunks — with **no truncation**. If a reviewer leaves many comments with large diff hunks, this string could become very large. The old code truncated diffs at 2000 chars. The new code has no such limit.

Additionally, since this is passed via `format!("pr_review_context={}", review_context)`, if `review_context` contains `=` signs or newlines (which it absolutely will — it contains markdown and diff hunks), this may break the sidecar key-value parsing depending on how `sidecar::set` parses the `key=value` format. **Does `sidecar::set` split on the first `=` only, or all `=` signs?** If it splits on all `=`, the stored value will be truncated/corrupted.

**Suggestion:** 
- Verify that `sidecar::set` handles multi-line values with `=` signs correctly.
- Add a size cap on `review_context` (e.g., 8KB or 16KB) to prevent excessively large agent prompts.

### 4. Timestamp comparison uses string ordering on ISO timestamps
**File:** `src/engine/mod.rs`, around lines 1392-1398

```rust
if review.submitted_at > latest_review_ts {
    latest_review_ts = review.submitted_at.clone();
}
if !last_review_ts.is_empty() && review.submitted_at <= last_review_ts {
    continue;
}
```

String comparison on ISO 8601 timestamps (`2024-01-15T10:30:00Z`) works correctly for lexicographic ordering **as long as all timestamps use the same format** (same timezone, same precision). This is fine for GitHub API responses which consistently use UTC ISO 8601. But it's fragile — if any timestamp lacks the `Z` suffix or uses a different timezone offset, it breaks silently.

This is a minor concern (GitHub is consistent), but worth a comment.

### 5. The `last_review_ts` is updated even when no re-dispatch happens
**File:** `src/engine/mod.rs`, around lines 1392-1439

The `latest_review_ts` is tracked for every `CHANGES_REQUESTED` review regardless of whether the review produces any actionable comments. However, `review_context` is only non-empty if there are actual review comments or a review body. Consider this scenario:

1. Reviewer submits a `CHANGES_REQUESTED` review with an empty body and no inline comments.
2. `latest_review_ts` gets updated but `review_context` remains empty.
3. The `if !review_context.is_empty()` check prevents re-dispatch (correct).
4. But `latest_review_ts` was already advanced past this review.

Actually, looking again — `latest_review_ts` is only persisted inside the `if !review_context.is_empty()` block, so this is fine. The local variable advances but isn't saved. **This is correct.** (Disregard this point.)

### 6. `pr_review_context` is never cleared after the agent addresses it
**File:** `src/engine/runner/context.rs` and `src/engine/mod.rs`

When the agent is re-dispatched and successfully addresses the review feedback, the `pr_review_context` sidecar value is never cleared. This means:

1. Agent runs, addresses feedback, pushes to PR.
2. Task goes back to `in_review` status.
3. If the agent gets re-dispatched for another reason (e.g., CI failure retry), it will see the stale review context again and may try to re-address already-fixed issues.

**Suggestion:** Clear `pr_review_context` from the sidecar after the agent completes its run (e.g., in the task runner after successful execution), or at least after the task transitions out of `in_progress`.

### 7. No test coverage for the new `review_open_prs` logic

The PR removes ~150 lines of logic (`create_review_follow_up_task`, `get_existing_review_tasks`) and adds ~50 lines of new logic (review context building, timestamp tracking, re-dispatch). The test plan says "Tests pass" but there are no new tests for:

- Building review context from multiple reviews
- Timestamp-based deduplication
- Re-dispatch flow (status transition to Routed)
- Edge case: review with no actionable comments
- Edge case: review context sidecar persistence

The only test change is adding `pr_review_context: String::new()` to an existing test struct initialization (`context.rs:355`), which is just a compilation fix.

**Suggestion:** Add at least a unit test for the review context building logic, and an integration test for the re-dispatch flow.

## Minor Notes

- **Good:** The simplification from per-comment tasks to batch re-dispatch is a better UX for the agent — it sees all feedback at once instead of one comment at a time.
- **Good:** The `build_agent_message` injection point in `agent.rs:309-314` is clean and follows the existing pattern.
- The comment filtering logic (line ~1410: `c.user.login == review.user.login && c.created_at >= review.submitted_at`) could theoretically pick up comments from the same user on a different review if two reviews happen close together, but this is an edge case unlikely to cause real problems.

## Summary of Required Changes

1. **Critical:** Verify/ensure the original agent is preserved during re-dispatch to `Routed` status, or the task may be routed to a different agent that doesn't have the worktree context.
2. **Critical:** Verify `sidecar::set` correctly handles multi-line values containing `=` signs in the `pr_review_context`.
3. **Important:** Clear `pr_review_context` after the agent completes its re-dispatch run to prevent stale context on subsequent runs.
4. **Important:** Add a size cap on the review context string.
5. **Recommended:** Add tests for the new review context building and re-dispatch logic.

---

PR #168, opened 2026-02-27
fix: webhook signature verification timing-safety and comment encoding

### LINKED ISSUE #167: fix: webhook signature verification is not timing-safe + comment body encoding broken
## Problem

Two bugs in the GitHub webhook/channel layer affect security and correctness:

1. **Timing-unsafe HMAC verification** — `verify_signature` in `channels/github.rs` uses plain string equality (`==`) to compare HMAC digests, which is vulnerable to timing side-channel attacks. An attacker can incrementally guess the webhook secret by measuring response times.

2. **Comment body newlines are garbled** — `post_comment` replaces `\n` with literal `\\n` before passing to `gh api --field body=...`. The `--field` flag does NOT interpret escape sequences, so multi-line comments are posted with literal `\n` strings instead of actual newlines.

### Timing-unsafe HMAC
- **`src/channels/github.rs:357-367`** — `verify_signature` computes HMAC correctly but compares with `expected == signature` (standard string equality). Should use constant-time comparison via `hmac::Mac::verify_slice` or `subtle::ConstantTimeEq`.

### Broken comment encoding
- **`src/channels/github.rs:113`** — `body.replace('\n', "\\n")` produces literal backslash-n. The correct approach is `--raw-field body=<value>` (which sends the value verbatim including newlines) or piping via stdin with `--field body=@-`.

### Related webhook issues
- **`src/channels/github.rs:426,447`** — Webhook event IDs use `format!("comment-{}-{}", issue.number, timestamp.timestamp())` instead of GitHub's actual comment ID. Two comments on the same issue within the same second get the same ID, breaking deduplication.
- **`src/channels/github.rs:520`** — `check_webhook_health` uses `Client::builder().build().unwrap_or_default()`, which silently drops the 5-second timeout on builder failure, potentially causing health checks to hang indefinitely.

## Impact

- **Security**: Webhook secret can be leaked via timing analysis. Low-probability in practice (requires precise timing measurements against localhost), but trivial to fix.
- **Comment readability**: All GitHub comments posted by the webhook channel have broken formatting — newlines appear as literal `\n` text. This affects agent status updates and review comments.
- **Duplicate events**: Two comments posted in the same second on the same issue produce identical IDs, causing the second event to be silently dropped.

## Suggested fix

1. Use `hmac::Mac::verify_slice` or the `subtle` crate's `ConstantTimeEq` for HMAC comparison
2. Replace `--field body=...` with `--raw-field body=...` or pipe body via stdin
3. Use the actual GitHub comment ID from the webhook payload instead of synthesizing one
4. Add explicit error handling for `Client::builder().build()` failure

## Test coverage gaps

- No test for `verify_signature` (neither correctness nor timing safety)
- No test for `post_comment` with multi-line bodies
- No test for webhook event ID uniqueness

gabrielkoerich:
## Summary
- Fixed timing-unsafe HMAC verification by using constant-time comparison (`subtle::ConstantTimeEq`)
- Fixed comment body newlines by using `--raw-field` instead of `--field` for gh CLI
- Fixed webhook event IDs to use actual GitHub comment/review IDs instead of synthesized timestamps
- Fixed webhook health check to properly handle HTTP client builder errors

## Test plan
- [x] All 276 tests pass
- [x] Code compiles without warnings

Closes #167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1330,6 +1330,7 @@ dependencies = [
  "serde_json",
  "serde_yml",
  "sha2",
+ "subtle",
  "tempfile",
  "thiserror",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,9 @@ tower-http = { version = "0.5", features = ["trace"] }
 # HMAC for webhook signature verification
 hmac = "0.12"
 
+# Constant-time comparison for timing-safe signature verification
+subtle = "2"
+
 # Async combinators (FuturesUnordered for concurrent routing)
 futures = "0.3"
 
--- a/src/channels/github.rs
+++ b/src/channels/github.rs
@@ -109,8 +109,8 @@ impl GitHubChannel {
                 "Accept: application/vnd.github.v3+json",
                 "--method",
                 "POST",
-                "--field",
-                &format!("body={}", body.replace('\n', "\\n")),
+                "--raw-field",
+                &format!("body={}", body),
             ])
             .output()?;
 
@@ -300,6 +300,7 @@ use axum::{
 };
 use hmac::{Hmac, Mac};
 use sha2::Sha256;
+use subtle::ConstantTimeEq;
 
 type HmacSha256 = Hmac<Sha256>;
 
@@ -342,6 +343,7 @@ struct PullRequestPayload {
 
 #[derive(serde::Deserialize, serde::Serialize)]
 struct CommentPayload {
+    id: u64,
     body: String,
     user: GhUser,
     created_at: Option<DateTime<Utc>>,
@@ -349,6 +351,7 @@ struct CommentPayload {
 
 #[derive(serde::Deserialize, serde::Serialize)]
 struct ReviewPayload {
+    id: u64,
     state: String,
     body: Option<String>,
     user: GhUser,
@@ -363,7 +366,8 @@ fn verify_signature(secret: &str, payload: &[u8], signature: &str) -> bool {
     let result = mac.finalize().into_bytes();
 
     let expected = format!("sha256={:x}", result);
-    expected == signature
+    // Use constant-time comparison to prevent timing side-channel attacks
+    bool::from(expected.as_bytes().ct_eq(signature.as_bytes()))
 }
 
 fn parse_github_event(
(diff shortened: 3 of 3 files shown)


gabrielkoerich:
## 🟣 Automated Review — Approve

This is a clean, well-focused PR that addresses four distinct issues. Each change is correct and well-tested.

### 1. Timing-safe HMAC verification (src/channels/github.rs:369)
**Good fix.** Replacing `expected == signature` with `bool::from(expected.as_bytes().ct_eq(signature.as_bytes()))` closes a real timing side-channel vulnerability. The `subtle` crate is the standard choice for this in Rust.

Minor note: `ct_eq` on slices of different lengths short-circuits (non-constant-time length check), but this is acceptable here since both sides should always be 71 bytes (`sha256=` + 64 hex chars). GitHub always sends lowercase hex and `format!("{:x}")` produces lowercase, so there's no case mismatch risk.

### 2. Comment body encoding (src/channels/github.rs:112)
**Good fix.** Switching from `--field` with manual `body.replace('\n', "\\n")` to `--raw-field` with the raw body is correct. The old approach risked double-encoding: `--field` performs its own JSON encoding, so pre-escaping newlines would produce literal `\n` in the output rather than actual newlines. `--raw-field` sends the value as a raw string and lets `gh` handle JSON serialization correctly.

### 3. Stable webhook event IDs (src/channels/github.rs:430, 451)
**Good fix.** Using actual GitHub `comment.id` / `review.id` (e.g., `comment-12345`) instead of synthesized `comment-{issue}-{timestamp}` is better because:
- GitHub IDs are globally unique and stable
- Timestamp-based IDs are fragile (clock skew, parsing differences)
- Enables proper deduplication — the same webhook event always produces the same message ID

The `id: u64` fields added to `CommentPayload` (line 346) and `ReviewPayload` (line 354) are non-optional, which is correct since GitHub always includes `id` in these payloads.

### 4. Health check error handling (src/channels/github.rs:522-528)
**Good fix.** The old `.unwrap_or_default()` silently fell back to a `reqwest::Client` without the 5-second timeout, which could cause the health check to hang indefinitely. The new `match` properly logs the error and returns `false`.

### Tests
All test payloads correctly include the new `id` fields, and new assertions (lines 756, 813) verify the ID format. Test coverage is adequate for the changes.

### No issues found
No bugs, security concerns, or missing edge cases. The dependency addition (`subtle = "2"` in Cargo.toml) is minimal and well-scoped.

---

PR #201, opened 2026-02-28
fix: Codex and OpenCode agents ignore system prompt, degrading task success

### LINKED ISSUE #194: fix: Codex and OpenCode agents ignore system prompt, degrading task success
## Problem

Both `Codex` and `OpenCode` agent runners discard the system prompt file (`_sys_file` parameter is explicitly unused). This means non-Claude agents operate without:
- Project instructions and constraints
- Output format specification (JSON envelope)
- Git workflow rules (conventional commits, worktree isolation)
- Security guidelines

This is likely the **single biggest contributor** to lower task completion rates for non-Claude agents.

## Root Cause

In both `build_command()` implementations, the `sys_file` parameter is prefixed with `_` (intentionally ignored):

- `src/engine/runner/agents/codex.rs:241` — `_sys_file: &str`
- `src/engine/runner/agents/opencode.rs:209` — `_sys_file: &str`

Meanwhile, the Claude runner correctly uses `sys_file` via `--system-prompt`:
```
src/engine/runner/agents/claude.rs:222-226
```

## Fix

**Codex**: Use the `--instructions` flag (reads from file):
```rust
format!(
    r#"cat "{msg_file}" | {timeout_cmd} codex {model_flag} \
  --instructions "{sys_file}" \
  --ask-for-approval {approval} \
  --sandbox {sandbox} \
  exec --json -"#,
    ...
)
```

**OpenCode**: Prepend system prompt to the message file, or use OpenCode's system prompt flag if available. At minimum, concatenate:
```rust
format!(
    r#"cat "{sys_file}" "{msg_file}" | {timeout_cmd} opencode run {model_flag} \
  --format json -"#,
    ...
)
```

## Impact

Every non-Claude agent invocation currently runs blind — no project context, no output format, no workflow rules. Fixing this should materially improve task completion rates for Codex and OpenCode.

## Files
- `src/engine/runner/agents/codex.rs` (line 237-272)
- `src/engine/runner/agents/opencode.rs` (line 205-225)
- `src/engine/runner/agent.rs` (line 59 — where sys_file is written)

gabrielkoerich:
Closes #194

### DIFF
--- a/src/engine/runner/agents/codex.rs
+++ b/src/engine/runner/agents/codex.rs
@@ -238,7 +238,7 @@ impl AgentRunner for CodexRunner {
         &self,
         model: Option<&str>,
         timeout_cmd: &str,
-        _sys_file: &str,
+        sys_file: &str,
         msg_file: &str,
         permissions: &PermissionRules,
     ) -> String {
@@ -265,11 +265,13 @@ impl AgentRunner for CodexRunner {
 
         format!(
             r#"cat "{msg_file}" | {timeout_cmd} codex {model_flag} \
+  --instructions "{sys_file}" \
   {permission_flags} \
   exec --json -"#,
             msg_file = msg_file,
             timeout_cmd = timeout_cmd,
             model_flag = model_flag,
+            sys_file = sys_file,
             permission_flags = permission_flags,
         )
     }
--- a/src/engine/runner/agents/opencode.rs
+++ b/src/engine/runner/agents/opencode.rs
@@ -206,7 +206,7 @@ impl AgentRunner for OpenCodeRunner {
         &self,
         model: Option<&str>,
         timeout_cmd: &str,
-        _sys_file: &str,
+        sys_file: &str,
         msg_file: &str,
         permissions: &PermissionRules,
     ) -> String {
@@ -225,12 +225,13 @@ impl AgentRunner for OpenCodeRunner {
         };
 
         format!(
-            r#"{config_setup}{timeout_cmd} opencode run {model_flag} \
-  --format json - < "{msg_file}""#,
+            r#"{config_setup}cat "{sys_file}" "{msg_file}" | {timeout_cmd} opencode run {model_flag} \
+  --format json -"#,
             config_setup = config_setup,
+            sys_file = sys_file,
+            msg_file = msg_file,
             timeout_cmd = timeout_cmd,
             model_flag = model_flag,
-            msg_file = msg_file,
         )
     }
 


gabrielkoerich:

The core fix (codex.rs + opencode.rs) is correct and exactly what was asked for. However, the PR introduces several unrelated regressions in engine/mod.rs, runner/mod.rs, and git_ops.rs that break the task review lifecycle. Tests pass only because they don't cover the affected runtime paths.

**Issues Found:**
- `src/engine/mod.rs` line 956: Review trigger changed from `status == "in_review"` to `status == "done"`. But the runner sets `in_review` (not `done`) when a PR exists — so the review agent will NEVER fire for tasks with a PR. The status read from sidecar at this point is already `in_review` because the runner's `check_pr_override` overrides it before this check runs. Revert to `if status == "in_review"`. [error]
- `src/engine/runner/git_ops.rs` line 55: Re-added hardcoded `feat:` prefix to auto-commit messages (`format!("feat: {title}...")`). This was explicitly removed in commit e5ba9ad. Auto-commit messages should use the task title as-is without forcing a conventional commit prefix, since agents already write proper commit messages. [error]
- `src/engine/mod.rs` line 1117: Removed the fallback review trigger from `sync_tick` that handled `in_review` + `NeedsReview` tasks. This was the recovery path for tasks that got stuck in `in_review` without being reviewed (e.g., after a restart). Without it, any `in_review` task that slips through the tick path will stay stuck forever. [error]
- `src/engine/mod.rs` line 1804: Removed `sidecar::set(task_id, &["review_started=false".to_string()])` from the `review_open_prs` re-dispatch path. When changes are requested and the task is sent back to `routed`, `review_started` stays `true`. When the agent fixes the code and the task reaches `in_review` again, the duplicate-guard check will see `already_reviewing=true` and skip the review — the task will be stuck in `in_review` permanently on the second attempt. [error]
- `src/engine/runner/mod.rs` line 519: Added `check_pr_override` which calls `gh pr list` to determine if a PR exists — but this information was already available from the return value of `create_pr_if_needed` (it returns `Ok(Some(url))` or `Ok(None)`). The refactor from tracking `has_pr` to a separate gh call adds unnecessary API round-trips and ignores the result already in hand. Restore the `has_pr` tracking pattern. [warning]
- `src/engine/mod.rs` line 488: Removed startup reset of stale `review_started` flags for tasks in `in_review`. This was a safety net: if the service crashes while reviewing, the flag stays `true` and the task never gets reviewed again. Removing it means service restarts can permanently block review for any in-flight `in_review` tasks. [warning]
- `PLAN.md` line 138: Removed the task lifecycle documentation (status flow table and invariants). This documentation accurately describes the intended design and is useful for future agents/developers. No reason to delete it as part of this fix. [warning]

PR #202, opened 2026-02-28
fix: error handling inconsistencies cause silent failures and cascading tick aborts

gabrielkoerich:
Closes #195

Issue #195, opened 2026-02-28
fix: error handling inconsistencies cause silent failures and cascading tick aborts


gabrielkoerich:
## Problem

Multiple error handling patterns in the engine cause either (a) silent data loss when errors are swallowed, or (b) cascading aborts where one bad task kills the entire tick cycle. Both degrade reliability.

### 1. Stuck recovery `?` aborts entire tick (cascading failure)
**File:** `src/engine/mod.rs:791-809`

In Phase 2 (stuck task recovery), `sidecar::set()`, `backend.update_status()`, and `backend.post_comment()` all use the `?` operator. If any single stuck task fails to recover (e.g., transient GitHub API error), the entire tick aborts — preventing all subsequent phases (routing, dispatch, jobs, cleanup) from running.

**Fix:** Use per-task error handling with `continue`:
```rust
if let Err(e) = sidecar::set(&task.id.0, &[...]) {
    tracing::warn!(task_id = task.id.0, ?e, "failed to reset sidecar for stuck task");
    continue;
}
```

### 2. `push_branch()` returns `Ok(false)` on failure — caller doesn't check
**File:** `src/engine/runner/git_ops.rs:131-135` and `src/engine/runner/mod.rs:415-420`

When `git push` fails, `push_branch()` returns `Ok(false)`. The caller at `mod.rs:415` uses `if let Err(e)` which only matches `Err`, so the `Ok(false)` case is silently ignored. The task proceeds to create a PR for code that was never pushed, then gets marked as done.

**Fix:** Either return `Err` on push failure, or check the boolean:
```rust
match git_ops::push_branch(&wt.work_dir, &wt.branch, &wt.default_branch).await {
    Ok(false) => {
        tracing::error!(task_id, "push failed, skipping PR creation");
        sidecar::set(task_id, &["last_error=push failed"])?;
        // skip PR creation
    }
    Err(e) => { ... }
    Ok(true) => { /* proceed to create PR */ }
}
```

### 3. Model cooldowns cleared at wrong interval
**File:** `src/engine/runner/response.rs:359-364`

`clear_expired_cooldowns()` uses `AGENT_COOLDOWN_SECS` (30 min) for ALL entries, but model-specific cooldowns should use `MODEL_COOLDOWN_SECS` (60 min). Model bans expire at half the intended duration, causing re-dispatch to broken models.

**Fix:** Check the key format to determine which timeout to use:
```rust
[...]

PR #215, opened 2026-03-01
fix: resolve error handling inconsistencies causing silent failures and tick aborts

gabrielkoerich:
## Summary

Fixes all 4 root causes from issue #195/#202:

- **`engine/mod.rs`**: Cascading `?` operators replaced with per-phase `match`/`warn` so a single GitHub API failure no longer aborts the entire tick cycle. List calls for `in_progress`, `routable`, `routed`, and `blocked` tasks all fall back to `Vec::new()` on error.

- **`runner/mod.rs`**: `push_branch()` now explicitly handles all three cases: `Ok(true)` proceeds to PR creation, `Ok(false)` logs an error and skips PR creation (was silently ignored before), `Err(e)` logs the error and skips PR creation.

- **`runner/response.rs`**: `clear_expired_cooldowns()` now uses `MODEL_COOLDOWN_SECS` (60 min) for keys containing `:` (model-specific entries like `claude:claude-opus-4-5`) and `AGENT_COOLDOWN_SECS` (30 min) for agent-only keys, fixing premature cooldown expiry.

- **`commands.rs`**: Command ID truncation now sorts numerically (`sort_by_key(|id| id.parse::<u64>()...)`) instead of alphabetically, preventing recently-processed IDs from being discarded during the 500-entry truncation.

## Test plan

- [x] `cargo build` — compiles cleanly
- [x] `cargo test` — 391 tests pass, 0 failures

Closes #202

### DIFF
--- a/src/engine/commands.rs
+++ b/src/engine/commands.rs
@@ -248,8 +248,8 @@ pub async fn scan_commands(
     // Persist processed IDs (keep last 500 to avoid unbounded growth)
     if !new_processed.is_empty() {
         let mut all: Vec<String> = processed_ids.into_iter().collect();
-        all.sort();
         all.extend(new_processed);
+        all.sort_by_key(|id| id.parse::<u64>().unwrap_or(0));
         if all.len() > 500 {
             all = all.split_off(all.len() - 500);
         }
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -746,9 +746,16 @@ async fn tick(
 
     // Phase 2: Recover stuck tasks
     let _phase2 = tracing::info_span!("engine.tick.phase2.stuck_tasks").entered();
-    let in_progress = task_manager
+    let in_progress = match task_manager
         .list_external_by_status(Status::InProgress)
-        .await?;
+        .await
+    {
+        Ok(tasks) => tasks,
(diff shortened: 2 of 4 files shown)


gabrielkoerich:
[2026-03-01T00:18:28Z] in_review: All changes from the previous attempt are already in place and working correctly:

- **391 tests pass, 0 failures**
- **Build is clean**

The task is complete. The diff shows all 4 root causes have been fixed:

1. **`engine/mod.rs`**: Cascading `?` operators replaced with `match`/`warn` for all 4 phase list calls (in_progress, routable, routed, blocked) — a single API failure no longer aborts the entire tick cycle.

2. **`runner/mod.rs`**: `push_branch()` now explicitly handles `Ok(true)`, `Ok(false)`, and `Err(e)` — silently-ignored push failures now log errors and skip PR creation.

3. **`runner/response.rs`**: `clear_expired_cooldowns()` uses `MODEL_COOLDOWN_SECS` (60 min) for model-specific keys (containing `:`) and `AGENT_COOLDOWN_SECS` (30 min) for agent-only keys.

4. **`commands.rs`**: Numeric sort (`sort_by_key(|id| id.parse::<u64>()...)`) replaces alphabetic sort for command ID truncation.

Issue #278, opened 2026-03-02
fix: agent data pipeline loses file changes and uses stale git refs


gabrielkoerich:
## Problem

Two related bugs cause agents to work with incorrect or missing data, directly impacting agent success rate on retries and file change tracking.

### 1. Prompt/parser field name mismatch silently drops file change data

The agent system prompt (`prompts/agent_system.md:93,138`) instructs agents to output a `files_changed` field:

```json
"files_changed": ["list of files modified"],
```

But the parser (`src/parser.rs:128`) only reads `files`:

```rust
let files = extract_string_array(obj.get("files"));
```

Any agent that follows the prompt literally will have its file list silently dropped. The `files_changed` field is also used in `internal_tasks.rs` for the DB column name, adding to the confusion.

**Fix**: Add `files_changed` as a fallback in parser.rs:128:
```rust
let files = extract_string_array(obj.get("files"))
    .or_else(|| {
        let fc = extract_string_array(obj.get("files_changed"));
        if !fc.is_empty() { Some(fc) } else { None }
    })
    .unwrap_or_default();
```
Or update the prompt to use `files` consistently.

### 2. `build_git_diff` and `has_unpushed_commits` use stale local branch refs

In worktrees, `rebase_on_default` fetches and rebases against `origin/{default_branch}`, keeping the remote ref fresh. But two functions compare against the **local** branch ref which may be stale:

- `src/engine/runner/context.rs:175` — `build_git_diff` diffs against local `default_branch`:
  ```rust
  .args(["diff", default_branch])  // should be origin/{default_branch}
  ```
  This sends wrong diff context to agents on retry (may show too many or too few changes).

- `src/engine/runner/git_ops.rs:432` — `has_unpushed_commits` fallback uses local ref:
  ```rust
  format!("{default_branch}..HEAD")  // should be origin/{default_branch}..HEAD
  ```
  This can cause false negatives (missed push), meaning PRs don't get created.

**Fix**: Use `format!("origin/{default_branch}")` in both locations.

## Impact

- File change tracking is unreliable (prompt mismatch)
- Retry context is inaccurate (stale local refs)
- Push detection can miss commits (stale local refs)

All three directly reduce agent success rate on non-trivial tasks.

PR #280, opened 2026-03-02
fix: agent data pipeline loses file changes and uses stale git refs

gabrielkoerich:
## Summary

Clean compile. All three fixes are in place:

1. **`src/parser.rs`** — `files` now falls back to `files_changed` when `files` is absent/empty, so agents following the prompt literally no longer have their file list silently dropped.

2. **`src/engine/runner/context.rs:175`** — `build_git_diff` now diffs against `origin/{default_branch}` (fresh remote ref) instead of the potentially-stale local branch.

3. **`src/engine/runner/git_ops.rs:432`** — `has_unpushed_commits` fallback now compares `origin/{default_branch}..HEAD`, preventing false negatives that would skip PR creation.

---
*Task #278 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch)*

### DIFF
--- a/src/engine/runner/context.rs
+++ b/src/engine/runner/context.rs
@@ -172,7 +172,7 @@ pub async fn build_repo_tree(project_dir: &Path) -> String {
 /// Build git diff from base branch (only for retries).
 pub async fn build_git_diff(project_dir: &Path, default_branch: &str) -> String {
     let output = Command::new("git")
-        .args(["diff", default_branch])
+        .args(["diff", &format!("origin/{default_branch}")])
         .current_dir(project_dir)
         .output_with_context()
         .await;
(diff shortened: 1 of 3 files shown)


gabrielkoerich:
## Automated Review — Approve

All three fixes correctly address the data pipeline issues:

1. Parser now handles both 'files' and 'files_changed' field names, resolving the prompt/parser mismatch that silently dropped file change data.

2. build_git_diff now diffs against origin/{default_branch} ensuring agents receive accurate context on retries instead of stale local refs.

3. has_unpushed_commits fallback now uses origin/{default_branch}..HEAD preventing false negatives that could skip PR creation.

Code quality is good - minimal, focused changes with no regressions. All 400 tests pass.

Issue #335, opened 2026-03-03
Database rate_limits table has no cleanup - unbounded growth


gabrielkoerich:
## Summary

The `rate_limits` table in SQLite accumulates records indefinitely without any cleanup mechanism, causing unbounded database growth over time.

## Location

`src/db.rs:454-466` (insert), schema at top of file

## Problem

The `record_rate_limit()` method inserts records:
```rust
conn.execute(
    "INSERT INTO rate_limits (agent, limit_type, occurred_at, task_id) VALUES (?1, ?2, datetime('now'), ?3)",
    ...
)?;
```

However, there is no:
- `DELETE` statement anywhere in the codebase for this table
- TTL or automatic expiration
- Periodic cleanup job

## Evidence

Search for rate_limits cleanup:
```bash
$ grep -rn "rate_limits" src --include="*.rs" | grep -i "delete\|clean\|prune\|expire"
# No results
```

The only queries on `rate_limits` are:
- `record_rate_limit()` - INSERT (line 454)
- `get_metrics_summary_24h()` - COUNT for last 24h (line 436)

## Impact

- Database file grows indefinitely
- Query performance degrades over time
- Disk space exhaustion on long-running deployments
- Metrics queries slow down as table grows

## Suggested Fix

Add a cleanup method and call it periodically (e.g., during metrics summary queries):

```rust
/// Delete rate limit records older than 30 days.
pub async fn cleanup_old_rate_limits(&self) -> anyhow::Result<usize> {
    let conn = self.conn.lock().await;
    let deleted = conn.execute(
        "DELETE FROM rate_limits WHERE occurred_at < datetime('now', '-30 days')",
        [],
    )?;
    Ok(deleted)
}
```

Then call it from `get_metrics_summary_24h()` or add a periodic cleanup job.

## Severity

Low-Medium - Causes gradual resource exhaustion, not immediate failure.

## Alternative Solutions

1. Use SQLite's `WITHOUT ROWID` and add a trigger to auto-delete old records
2. Add a cron job that runs cleanup periodically
3. Prune during `record_rate_limit()` call (delete records older than N days before inserting)

PR #336, opened 2026-03-03
Database rate_limits table has no cleanup - unbounded growth

gabrielkoerich:
## Summary

439 tests pass. The fix is minimal and correct:

**What changed** (`src/db.rs:453-471`): Added a `DELETE` before each `INSERT` in `record_rate_limit()`. On every rate limit event, records older than 30 days are pruned in the same locked connection, keeping the table bounded to ~30 days of history.

This approach was chosen over alternatives because:
- No new callers or periodic jobs needed
- Rate limit events are infrequent (only on API errors), so the DELETE cost is negligible  
- The cleanup runs in the same mutex lock as the insert, so no separate transaction needed
- The existing `idx_rate_limits_occurred_at` index makes the DELETE fast

---
*Task #335 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch)*

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -450,7 +450,7 @@ impl Db {
         })
     }
 
-    /// Record a rate limit event.
+    /// Record a rate limit event. Prunes records older than 30 days on each insert.
     pub async fn record_rate_limit(
         &self,
         agent: &str,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

CI passes: formatting clean, clippy clean, 439/439 tests pass. The fix is correct and minimal — a DELETE before the INSERT in record_rate_limit() prunes records older than 30 days. The idx_rate_limits_occurred_at index exists in the schema, making the DELETE efficient. Running cleanup on insert is appropriate since rate limit events are rare (only triggered on API errors), keeping overhead negligible. No new callers, no periodic jobs, no unnecessary complexity.

Issue #343, opened 2026-03-03
Review agent can be dispatched multiple times via sync_tick


gabrielkoerich:
## Problem

The `sync_tick` function in `src/engine/sync.rs` can spawn multiple review agents for the same task because it lacks the in-memory guard present in `tick_dispatch_tasks`.

### Location
- `src/engine/sync.rs` lines 63-127 - `sync_tick()` function
- `src/engine/tick.rs` lines 253-301 - `tick_dispatch_tasks()` (has guard)

### Root Cause

`tick_dispatch_tasks` uses an in-memory `dispatching: HashSet` to prevent double-dispatch:
```rust
// src/engine/tick.rs:260-270
let dispatch_key = format!("{}/{}", repo, task.id.0);
{
    let guard = dispatching.lock().unwrap();
    if guard.contains(&dispatch_key) {
        continue;  // Already dispatching, skip
    }
}
```

But `sync_tick` has no such guard when triggering review agents:
```rust
// src/engine/sync.rs:68-127
if enable_review {
    if let Ok(needs_review) = backend.list_by_status(Status::NeedsReview).await {
        for task in needs_review {
            // ... NO guard check here ...
            tokio::spawn(async move {
                review_and_merge(...).await  // Can spawn multiple times!
            });
        }
    }
}
```

### Race Condition Flow
1. Task T is in NeedsReview status
2. Sync tick runs at t=0, spawns review agent for T
3. Status transition to InReview hasn't completed yet (async)
4. Sync tick runs again at t=45s, sees T still in NeedsReview
5. Second review agent spawned for same task
6. Both agents run simultaneously, causing:
   - Duplicate PR comments
   - Potential double-merge attempts
   - Resource waste

### Proposed Fix

Option 1: Share the `dispatching` HashSet with sync_tick (requires refactoring)
Option 2: Add a sidecar-based guard similar to `review_started` flag used elsewhere:

```rust
// Check if review already in progress
if sidecar::get(task_id, "review_started").unwrap_or_default() == "true" {
    continue;
}
sidecar::set(task_id, &["review_started=true".to_string()])?;
```

Option 3: Use the existing `in_review_since` sidecar field as a guard - if set and recent (<5 min), skip.

### Acceptance Criteria
- [ ] Review agent can only be dispatched once per task
- [ ] Guard persists across sync_tick invocations
- [ ] Guard is cleared appropriately on completion/failure

### Severity
[...]

PR #347, opened 2026-03-03
Review agent can be dispatched multiple times via sync_tick

gabrielkoerich:
## Summary

The fix is complete. Let me provide a summary of the changes:

## Summary

I've implemented a fix for the race condition where `sync_tick` could spawn multiple review agents for the same task.

### Changes Made

**`src/engine/sync.rs`:**
1. Added a sidecar-based `review_started` guard check before triggering review agents (lines 78-89)
2. Clears the guard when review completes - both on success and failure (lines 138-142)
3. Clears the guard when resetting stale InReview tasks (lines 179-182)

**`src/engine/tick.rs`:**
1. Added the same sidecar-based guard to the main dispatch flow for consistency (lines 367-373)
2. Clears the guard if the status transition to InReview fails (lines 384-385)
3. Clears the guard when review completes (lines 431-436)

### How the Fix Works

The `review_started` sidecar flag acts as a persistent guard that:
- **Prevents double-dispatch**: Before spawning a review agent, we check if `review_started=true`. If set, we skip.
- **Survives across ticks**: Unlike the in-memory `dispatching` HashSet, the sidecar persists across `sync_tick` invocations
- **Is cleared appropriately**: The guard is cleared when:
  - Review completes (success, failure, or error)
  - Stale InReview detection resets the task
  - Status transition to InReview fails

This addresses the race condition where:
1. Task T is in NeedsReview
2. Sync tick runs at t=0, spawns review agent for T
3. Status transition to InReview hasn't completed yet (async)
4. Sync tick runs again at t=45s, sees T still in NeedsReview
5. **With the fix**: The `review_started=true` flag prevents the second spawn

---
*Task #343 · Created by kimi[bot] via [Orch](https://github.com/gabrielkoerich/orch)*

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -679,7 +679,7 @@ pub async fn serve() -> anyhow::Result<()> {
                         for engine in &project_engines {
                             let repo = engine.repo.clone();
                             REPO_CONTEXT.scope(repo, async {
-                                if let Err(e) = sync::sync_tick(&engine.backend, &tmux, &engine.repo, &db, &config, &router, &dispatching).await {
+                                if let Err(e) = sync::sync_tick(&engine.backend, &tmux, &engine.repo, &db, &config, &router).await {
                                     tracing::error!(repo = %engine.repo, ?e, "sync tick failed for project");
                                 }
                             }).await;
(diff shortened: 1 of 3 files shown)


gabrielkoerich:

Issue #382, opened 2026-03-03
CI: Fix release workflow — skip macOS builds when no release, correct bump detection, and simplify formula


gabrielkoerich:
CI: Fix release workflow — skip macOS builds when no release, correct bump detection, and simplify formula

Problem:
- The release workflow currently cross-compiles macOS binaries on every push to `main` even when `check-release` determines no release is needed. This wastes CI minutes and runner time.
- Conventional-commit bump detection in `check-release` misses breaking-change subjects using the `!` syntax (e.g. `feat!:`, `feat(scope)!:`) and currently only scans commit subjects for `BREAKING CHANGE:` footers incorrectly.
- The generated Homebrew formula template in the release job contains an architecture conditional that inserts the same universal binary URL for both branches, which is misleading.

Why it matters:
- Skipping unnecessary macOS builds saves CI resources and reduces queue times.
- Correct bump detection prevents incorrect semver bumps and accidental releases.
- A clear formula template avoids packaging confusion and potential mistakes.

Proposed changes (actionable):
1) Make `build-macos` depend on `check-release` and add an `if` guard so it only runs when `needs.check-release.outputs.skip == 'false'` (edit `.github/workflows/release.yml`, job `build-macos`, around lines 74-86).
2) Update `check-release` bump detection (`.github/workflows/release.yml`, lines ~172-191) to detect `!` breaking-change subjects using a pattern like `grep -qE '^[a-z]+(\\(.*\\))?!:'`, and also inspect full commit bodies for `BREAKING CHANGE:` footers.
3) Simplify the Homebrew formula generation in the release job (`.github/workflows/release.yml` around lines 349-366 and `Formula/orch.rb`): either emit a single universal-binary formula or use real arch-specific URLs/SHA256 when separate artifacts are uploaded.
4) Add a CI assertion or job guard to ensure `build-macos` is skipped when `check-release` outputs `skip=true`.

Files to inspect and change:
- `.github/workflows/release.yml` (jobs: `check-release`, `build-macos`, release packaging) — see line references in the workflow file.
- `Formula/orch.rb` (local formula template).

Acceptance criteria:
- macOS builds are skipped when `check-release` signals `skip=true`.
- Breaking changes expressed via `!` or `BREAKING CHANGE:` are detected and produce the correct semver bump.
- Generated formula is unambiguous about universal vs arch-specific artifacts.

Labels: ci, infra, release

PR #385, opened 2026-03-03
CI: Fix release workflow — skip macOS builds when no release, correct bump detection, and simplify formula

gabrielkoerich:
## Summary

**CI Release Fixes**

- Implemented the requested CI changes: skip macOS builds when no release is needed, correctly detect breaking-change commits using `!` or `BREAKING CHANGE:` footers, and simplify the Homebrew formula generation to avoid misleading arch-conditional templates.

What I changed
- Updated `.github/workflows/release.yml`:
  - **build-macos** job: now depends on `check-release` and has a guard so it only runs when `needs.check-release.outputs.skip == 'false'` (line ~74). This prevents macOS cross-compiles when `check-release` decides no release is required.
  - **check-release / Determine version bump**: improved bump detection (lines ~172-224):
    - Collects commit subjects and full commit bodies.
    - Detects breaking-change subjects using the `!` syntax (regex: `^[a-z]+(\(.*\))?!:`).
    - Also inspects commit bodies for `BREAKING CHANGE:` (case-insensitive).
    - Applies major bump for either case, minor bump for `feat(...)` only if no major detected, otherwise patch.
  - **Homebrew formula generation**: when creating the formula in the release job the script now emits either:
    - a single universal-binary formula (if both arch URLs are identical), or
    - retains the original arch-specific template logic when URLs differ.
    This avoids the previous confusing `on_macos` block that inserted the same universal URL twice.
  - Minor: added comments and a clearer template selection branch.

- Added `Formula/orch.rb`:
  - A simple, local formula template that prefers a universal `url`/`sha256`. The CI job will overwrite/place the real formula in the tap as needed. This file is a clear, unambiguous template that documents the preferred approach.

What I ran
- Verified changed files and committed:
  - `git add .github/workflows/release.yml Formula/orch.rb`
  - `git commit -m "ci(release): skip macOS builds when no release, detect '!' breaking changes, simplify formula template"`
  - `git push origin HEAD`

Files changed
- `.github/workflows/release.yml`
- `Formula/orch.rb`

Accomplished
- Prevent macOS cross-compiles on main when `check-release` indicates no release (saves CI minutes).
- Correctly detect breaking changes expressed with `!` in commit subjects and `BREAKING CHANGE:` in commit bodies — ensures proper semver bumping.
- Simplified and clarified Homebrew formula generation to avoid misleading arch-conditional placeholders.

Remaining
- Add an explicit CI assertion/job that fails if `build-macos` runs while `check-release` returns `skip=true`. Currently the workflow logic prevents it; if you want an explicit guard/test step (e.g., in `check-release` output validation stage or a separate `assert-skip` job), I can add it. (Suggested options below.)
- Optional: add unit tests or workflow dry-runs to verify bump detection on sample commit ranges (can add a small script + tests).
- Optional: update docs (README/AGENTS.md) to mention the formula behavior and the release job's universal-vs-arch logic.

Blockers
- None.

Reason
- (empty)

Delegations
- None.

Suggested next steps (pick one)
1. Add a CI assertion job that runs earlier and fails if `build-macos` executes while `check-release` outputs `skip=true` — I can add a lightweight `assert-skip` job that checks `needs.check-release.outputs.skip` and fails if it's `true` while `build-macos` was scheduled.
2. Add tests / fixture commits used by `check-release` to validate bump detection (e.g., test cases for `feat!:`, `feat(scope)!:`, `BREAKING CHANGE:`).
3. If you'd like universal artifacts (single tarball containing both arch bins), update release packaging to upload a single tarball and have the formula refer to that. Current change assumes arch-specific artifacts remain the default but supports the universal case.

If you want I will:
- Add the explicit CI assertion/job (option 1) now.
- Or add small unit tests / scripts to verify commit-parsing behavior (option 2).

Which would you like me to do next?

---
*Task #382 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch)*

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,8 @@ jobs:
   # Cross-compile build jobs for macOS targets
   build-macos:
     needs: [test, secrets, check-release]
-    if: github.ref == 'refs/heads/main'
+    # Only run macOS cross-builds on main when check-release determined a release is needed
+    if: github.ref == 'refs/heads/main' && needs.check-release.outputs.skip == 'false'
     strategy:
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


gabrielkoerich:
## Automated Review — Approve

All requirements satisfied:

1. **macOS build skipping**: The `build-macos` job now correctly depends on `check-release` and only runs when `skip == 'false'`. This will save CI minutes on chore-only commits.

2. **Breaking change detection**: Correctly detects both `!` syntax in subjects (regex `^[a-z]+(\(.*\))?!:`) and `BREAKING CHANGE:` footers in commit bodies (case-insensitive). The logic properly prioritizes major bump over minor.

3. **Formula simplification**: The release job now generates either a universal formula (when URLs are identical) or an arch-specific formula (when they differ), avoiding the misleading template that inserted the same universal URL twice.

4. **New Formula/orch.rb**: A clean local template documenting the preferred universal binary approach.

The agent noted adding an explicit CI assertion job as remaining work - this is optional and the current workflow logic correctly prevents `build-macos` from running when `skip=true`.

gabrielkoerich:

PR #442, opened 2026-03-06
fix: internal tasks stuck in InProgress are never recovered after service restart

### LINKED ISSUE #437: fix: internal tasks stuck in InProgress are never recovered after service restart
## Problem

`tick_recover_stuck_tasks` (`src/engine/tick.rs:67`) only recovers external (GitHub) tasks stuck in InProgress:

```rust
let in_progress = task_manager.list_external_by_status(Status::InProgress).await?;
```

Internal tasks (`internal:N` IDs) stuck in InProgress state are never considered. After every service restart (e.g., `brew upgrade orch`), any internal task that was running becomes permanently stuck — it never resets to New and can never be re-dispatched.

## Why it happens

- `list_external_by_status` delegates to `backend.list_by_status` (GitHub only)
- The `dispatching` guard set is in-memory and clears on restart
- Internal tasks marked InProgress in SQLite have no recovery path
- `tick_dispatch_tasks` only dispatches `Routed` tasks, not stuck `InProgress` ones

## Impact

Since the code review job and other scheduled jobs create internal tasks, every service restart leaves a potentially stuck task. The task is permanently stuck until manually cleared from SQLite.

## Fix

In `tick_recover_stuck_tasks`, also query internal InProgress tasks via `task_manager.db_list_internal_by_status(DbStatus::InProgress)` and apply the same no-session recovery logic (reset to New if age > `no_session_stuck_timeout`).

```rust
// Add after existing external task recovery in tick_recover_stuck_tasks
let internal_in_progress = task_manager
    .db_list_internal_by_status(crate::db::TaskStatus::InProgress)
    .await?;
for task in &internal_in_progress {
    let session_name = tmux.session_name(repo, &format!("internal:{}", task.id));
    let has_session = tmux.session_exists(&session_name).await;
    let age = chrono::Utc::now() - task.updated_at;
    let threshold = if has_session { config.stuck_timeout } else { config.no_session_stuck_timeout };
    if age.num_seconds() > threshold as i64 {
        let eid = ExternalId(format!("internal:{}", task.id));
        let _ = task_manager.update_task_status(&eid, Status::New).await;
    }
}
```

## File

- `src/engine/tick.rs` — `tick_recover_stuck_tasks` function (line 59)

gabrielkoerich:
Resolves #437

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -163,6 +163,43 @@ pub(crate) async fn tick_recover_stuck_tasks(
             }
         }
     }
+
+    // Also recover internal tasks stuck in InProgress.
+    // Internal tasks use SQLite as their state store; the dispatching guard is in-memory
+    // and clears on service restart, leaving them permanently stuck.
+    use crate::db::TaskStatus as DbStatus;
+    let internal_in_progress = task_manager
+        .db_list_internal_by_status(DbStatus::InProgress)
+        .await?;
+    for task in &internal_in_progress {
+        let task_id_str = format!("internal:{}", task.id);
+        let session_name = tmux.session_name(repo, &task_id_str);
+        let has_session = tmux.session_exists(&session_name).await;
+        let age = chrono::Utc::now() - task.updated_at;
+        let threshold = if has_session {
+            config.stuck_timeout
+        } else {
+            config.no_session_stuck_timeout
+        };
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
Superseded by commits already merged to main (#454, #453, #465 cover internal task recovery).

PR #480, opened 2026-03-06
fix: sanitize colons in tmux session names to prevent session:window misparse

gabrielkoerich:
Tmux interprets colons in session names as session:window notation in -t target flags. Without sanitization, 'has-session -t orch-orch-internal:13' looks for session 'orch-orch-internal' window 13 (not found), making wait_for_completion return immediately with exit_code=-1.

This bug caused all internal tasks to fail instantly after dispatch.

**Changes:**
- Changed colons to underscores instead of hyphens in session_name()
- Added documentation explaining the colon replacement
- Added test to verify no colons in session names

**Test:**
```bash
cargo test session_name_sanitizes_colons
```

Fixes the critical reliability regression where internal tasks dispatched at startup all failed within milliseconds with exit code -1.

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -36,14 +36,19 @@ impl TmuxManager {
     ///
     /// The project name is derived from the repo slug (e.g. `owner/repo` → `repo`).
     /// This prevents collisions between projects with the same issue number.
+    ///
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
Colon sanitization is already in main (using dash). This changes dash→underscore which is an unnecessary behavioral change.

PR #492, opened 2026-03-06
fix: resolve empty repo in PR creation when engine service CWD is /

gabrielkoerich:
## Summary

- `get_current_repo()` returned `""` when called from the engine service process (CWD = `/`)
- This caused `create_pr_if_needed` to call `GET https://api.github.com/repos//pulls` → 404
- Result: `has_pushed=true`, `has_pr=false` → tasks routed to `needs_review` indefinitely
- Loop: review gate found no PR/commits → re-routed to `new` → dispatched again → infinite cycle

## Root Cause

`get_current_repo()` walks up from CWD for `.orch.yml`, then falls back to `get("gh.repo")` from config. The global config (`~/.orch/config.yml`) only has `projects: [/path/to/orch]` — not `gh.repo` directly. The project's `.orch.yml` has `gh.repo` but isn't in the service's CWD path.

## Fix

Added a final fallback in `get_current_repo()`: iterate `get_project_paths()` from the global config and return the first project's `gh.repo`. This covers the engine service case (CWD = `/`) without changing existing CWD-based resolution priority.

## Fixes

- Infinite retry loop for "respond to mention" internal tasks
- All tasks that complete without code changes (review, analysis jobs) were also potentially affected

## Test plan

- [x] `cargo fmt -- --check` ✓
- [x] `cargo clippy --all-targets -- -D warnings` ✓
- [x] `cargo test` ✓
- Manual: confirmed `get_current_repo()` logic by tracing the fallback path

### DIFF
--- a/docs/content/posts/morning-review-2026-03-06.md
+++ b/docs/content/posts/morning-review-2026-03-06.md
@@ -1,0 +1,69 @@
++++
+title = "Morning Review — 2026-03-06"
+date = 2026-03-06
+description = "Critical bug fixed: empty repo in PR creation causes infinite retry loop for mention-response tasks"
++++
+
+## What landed since last review (24h)
+
+| Commit | Description |
+|--------|-------------|
+| `8709483` | docs: consolidate 'orchestrator' → 'orch' naming (#490) |
+| `d5cb836` | fix: create child task for review changes instead of re-dispatching same task |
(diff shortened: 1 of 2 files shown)


gabrielkoerich:
Review requested changes. A follow-up task (internal:25) has been created to address the feedback.

---
*Posted by [Orch](https://github.com/gabrielkoerich/orch)*

PR #643, opened 2026-03-16
fix: only pull main when cleanup actually removes a worktree

gabrielkoerich:
## Summary

- `cleanup_task_worktree_with_opts` previously returned `Ok(())` for both actual cleanup and silent skips (TTL guard, active tmux session guard)
- The batch janitor set `cleaned_any = true` unconditionally after every call, triggering `git pull` every engine tick as long as any terminal task had an active tmux session
- This was observable as `pulled main after worktree cleanup` logging every ~60 seconds in the service log even with no real cleanup happening

## Changes

- Changed `cleanup_task_worktree_with_opts` return type to `anyhow::Result<bool>`: `Ok(true)` = cleanup done and task marked cleaned, `Ok(false)` = intentionally skipped
- TTL guard and tmux guard early returns now return `Ok(false)`
- Batch janitor only sets `cleaned_any = true` on `Ok(true)` — stops the spurious pull
- Inline `cleanup_task_worktree` also only pulls main when cleanup actually happened

## Test plan
- [x] All 678 tests pass (`cargo nextest run`)
- [x] `cargo fmt` and `cargo clippy --all-targets -- -D warnings` clean

### DIFF
--- a/src/engine/cleanup.rs
+++ b/src/engine/cleanup.rs
@@ -437,10 +437,16 @@ pub(crate) async fn cleanup_done_worktrees_with_opts(
             continue;
         }
 
-        if let Err(e) = cleanup_task_worktree_with_opts(task_id, repo, store, opts).await {
-            tracing::warn!(task_id, err = %e, "worktree cleanup failed for task");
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

---

PR #715, opened 2026-03-19
fix: TOCTOU race in /stream command causes panic when binding removed between check and use

### LINKED ISSUE #714: fix: TOCTOU race in /stream command causes panic when binding removed between check and use
## Bug

In `src/engine/mod.rs:1687-1690`, the `/stream` command handler calls `transport.get_binding(task_id)` twice:

```rust
if transport.get_binding(task_id).await.is_some() {
    // Bind this channel/thread as an additional output target
    // The session name is retrieved from the existing binding
    let binding = transport.get_binding(task_id).await.unwrap();  // ← PANIC
```

The RwLock is released between the `is_some()` check and the `unwrap()`. If the session binding is removed (e.g., task completes, session is unregistered) in that window, the second `get_binding()` returns `None` and `unwrap()` panics, crashing the service.

## Root Cause

Classic TOCTOU (time-of-check, time-of-use) on the bindings RwLock. The same file already uses the correct pattern at line 1208:

```rust
if let Some(binding) = transport.get_binding(task_id).await {
```

## Fix

Replace lines 1687-1693 with:

```rust
if let Some(binding) = transport.get_binding(task_id).await {
    transport
        .bind(task_id, &binding.tmux_session, channel, thread_id)
        .await;
    format!("Streaming output from task `{task_id}` to this channel.")
} else {
    format!("Task `{task_id}` is not currently running or has no active session.")
}
```

## Impact

Service panic when a user issues the `/stream` command at the exact moment the streamed task finishes, causing the entire orchestrator process to crash.

gabrielkoerich:
## Summary

I've successfully fixed the TOCTOU race condition in the `/stream` command handler. Here's what I did:

## Summary

**Fixed:** The race condition in `src/engine/mod.rs:1687-1690` where `transport.get_binding(task_id)` was called twice with a lock release between calls.

**Change:** Replaced the unsafe pattern:
```rust
if transport.get_binding(task_id).await.is_some() {
    let binding = transport.get_binding(task_id).await.unwrap();  // ← PANIC if binding was removed
```

With the safe pattern that captures the binding atomically:
```rust
if let Some(binding) = transport.get_binding(task_id).await {
```

**Testing:** All CI checks passed:
- ✅ `cargo fmt -- --check` 
- ✅ `cargo clippy --all-targets -- -D warnings`
- ✅ `cargo test` (499 tests passed)

**Commit:** Created with message describing the fix and linking to the issue.

The fix is now committed locally and ready to push. Per the landing requirements, I need to push this to remote to complete the work.

Closes #714

---
*Task #714 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -1684,10 +1684,9 @@ async fn handle_stream_command(
     let task_id = parts[1].trim();
 
     // Check if the task has an active binding (i.e. is running)
-    if transport.get_binding(task_id).await.is_some() {
+    if let Some(binding) = transport.get_binding(task_id).await {
         // Bind this channel/thread as an additional output target
         // The session name is retrieved from the existing binding
-        let binding = transport.get_binding(task_id).await.unwrap();
         transport
             .bind(task_id, &binding.tmux_session, channel, thread_id)
             .await;


gabrielkoerich:

---
*Reviewed by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.2-codex`*

PR #723, opened 2026-03-20
fix: discord interaction acknowledgment creates new reqwest::Client instead of reusing self.client

### LINKED ISSUE #721: fix: discord interaction acknowledgment creates new reqwest::Client instead of reusing self.client
## Bug

`src/channels/discord_ws.rs:559` creates a brand-new `reqwest::Client` on every button interaction to acknowledge it:

```rust
let client = Client::new();  // wasteful — creates new TCP pools, TLS ctx per interaction
let _ = client.post(&callback_url)...send().await;
```

`DiscordGateway` already owns a `client: Client` field (initialized in `new()` at line 61) that is used for all other API calls. This one call in `handle_dispatch()` bypasses it entirely.

## Root cause

The interaction acknowledgment code was added without noticing the existing `self.client`. Each interaction spawns a full `reqwest::Client` — which allocates a connection pool and TLS context — then immediately drops it after one request.

## Impact

Under any reasonable button interaction load, this leaks OS-level TCP resources (ephemeral ports, TLS handshake cost) for what should be a sub-millisecond fire-and-forget. It also bypasses any connection reuse that `self.client`'s pool provides.

## Fix

Replace `Client::new()` with `self.client.clone()` (cloning a `reqwest::Client` is cheap — it shares the underlying pool):

```rust
// Before
let client = Client::new();

// After
let client = self.client.clone();
```

gabrielkoerich:
## Summary

Perfect! The background tests have completed successfully. All 8 Discord websocket tests passed, which confirms the fix is working correctly.

The implementation is complete and ready for deployment. The changes ensure that Discord interaction acknowledgments reuse the existing client connection pool instead of creating wasteful new clients, eliminating resource leaks and improving performance under load.

Closes #721

---
*Task #721 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/channels/discord_ws.rs
+++ b/src/channels/discord_ws.rs
@@ -248,7 +248,8 @@ async fn run_gateway(
                 backoff = Duration::from_secs(1); // reset on success
 
                 let result =
-                    handle_connection(ws, &token, shard_id, shard_count, &mut state, &tx, &client).await;
+                    handle_connection(ws, &token, shard_id, shard_count, &mut state, &tx, &client)
+                        .await;
 
                 match result {
                     Ok(true) => {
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

The fix properly reuses self.client instead of creating a new Client::new() for Discord interaction acknowledgment. The client is threaded through the call chain from DiscordGateway::subscribe() through run_gateway() -> handle_connection() -> handle_dispatch(). All CI checks pass (format, clippy, 723 tests). Architecture alignment verified - no conflicts with AGENTS.md/CLAUDE.md. The variable renaming to http_client improves clarity.

---

PR #860, opened 2026-03-23
bug: review agent tmux session not cleaned up on rate limit — blocks all retries with duplicate session

### LINKED ISSUE #855: bug: review agent tmux session not cleaned up on rate limit — blocks all retries with duplicate session
## Bug

When the review agent hits a rate limit, the tmux session is not cleaned up. On retry, `tmux new-session` fails with "duplicate session" because the old session still exists.

## Observed behavior (task internal:8068)

```
23:03:28 review agent ran with codex → rate limit (usage limit until Mar 26)
23:03:28 reset to NeedsReview, failures=1
23:04:59 retry → "tmux new-session failed: duplicate session: orch-orch-internal-8068-review"
23:04:59 failures=3 → blocked
```

## Root cause

When the review agent fails with a rate limit error, the error path in `review.rs` returns the error but does NOT kill the tmux session. The session name `orch-{repo}-{task_id}-review` persists. On retry, `spawn_in_tmux()` tries to create the same session name and fails.

The agent runner's normal completion path kills the session (`session::cleanup_session`), but the rate limit error is detected BEFORE the agent even starts producing output — the session is created, the agent binary starts, immediately errors with rate limit, and the error classification happens in the response parser. The tmux session is left alive.

## Fix

In the review agent error path (both `tick.rs` review spawn and `subscribers/review.rs`), kill the tmux session before resetting to NeedsReview:

```rust
ReviewOutcome::Reset => {
    // Kill stale tmux session before retry
    tmux.kill_session(&review_session_name).await.ok();
    tokio::time::sleep(Duration::from_secs(10)).await;
    task_manager.update_task_status(&ExternalId(tid.clone()), Status::NeedsReview).await;
}
```

Also check `review_and_merge()` itself — the early-return error paths should clean up the session they created.

## Files

- `src/engine/review.rs` — `review_and_merge()` error paths (rate limit, spawn failure)
- `src/engine/tick.rs` — review spawn in dispatch (ReviewOutcome::Reset)
- `src/engine/subscribers/review.rs` — review spawn (ReviewOutcome::Reset)
- `src/engine/runner/session.rs` — `cleanup_session()` (reference for correct cleanup)

gabrielkoerich:
## Summary

All checks pass. Here's a summary of what was changed:

**`src/engine/subscribers/review.rs`** — In `ReviewOutcome::Reset`:
- Before the 10s backoff sleep, kill the stale review session (`{task_id}-review`)

**`src/engine/tick.rs`** — In both reset-to-NeedsReview paths (for `ReviewDecision::Failed` and `Err(e)` from `review_and_merge`):
- Before calling `update_task_status(NeedsReview)`, kill the stale review session

The fix works by deriving the session name from the task ID (`{task_id}-review`) and calling `kill_session` before resetting the task. This handles the race where two concurrent reviewers spawn the same session name — when the second one fails with "duplicate session", the Reset handler kills the stale session so the next retry can create it fresh. Killing a non-existent session is a no-op (logs a warning and returns Ok).

Closes #855

---
*Task #855 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/subscribers/review.rs
+++ b/src/engine/subscribers/review.rs
@@ -235,6 +235,15 @@ pub fn spawn(
                         };
                         match outcome {
                             ReviewOutcome::Reset => {
+                                // Kill any stale tmux review session before resetting — the
+                                // session may still be alive if the agent hit a rate limit,
+                                // if two concurrent reviewers raced and the loser's spawn
+                                // failed with "duplicate session", or if review_and_merge
+                                // returned an error before reaching its own cleanup.
+                                // Killing a non-existent session is a harmless no-op.
+                                let stale_session =
+                                    tmux_c.session_name(&repo_s, &format!("{}-review", tid));
+                                tmux_c.kill_session(&stale_session).await.ok();
                                 // Backoff before re-queuing to prevent rapid spin
                                 // if the review agent keeps failing instantly.
                                 tokio::time::sleep(std::time::Duration::from_secs(10)).await;
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -734,6 +734,13 @@ pub(crate) async fn tick_dispatch_tasks(
                                                         failures,
                                                         "review agent failed — resetting to NeedsReview for retry"
                                                     );
+                                                    // Kill any stale review session so the
+                                                    // next retry can create a fresh one.
+                                                    let stale = tmux_clone.session_name(
+                                                        &repo_owned,
+                                                        &format!("{}-review", task_id_for_review),
+                                                    );
+                                                    tmux_clone.kill_session(&stale).await.ok();
                                                     if let Err(e) = task_manager_for_review
                                                         .update_task_status(
                                                             &ExternalId(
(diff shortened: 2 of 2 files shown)


gabrielkoerich:
## Automated Review — Approve

Rebased onto `origin/main`, reviewed the documented architecture/settled decisions, and ran the project CI locally plus GitHub PR checks. I found one missing cleanup path outside the submitted diff: `src/engine/sync.rs` reset-to-`NeedsReview` also retried review work without killing the stale `*-review` tmux session, so the duplicate-session bug could still reproduce from the sync-driven review path. I fixed that gap and committed `acefcc9` (`fix: clean stale review tmux session in sync retry path`). After that, local checks passed (`cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo nextest run --no-fail-fast`, `cargo llvm-cov nextest --lcov --output-path lcov.info`, `gitleaks git --config .gitleaks.toml .`) and `gh pr checks 860 --watch --fail-fast` passed as well. With that follow-up fix included, the PR aligns with the intended review-agent tmux lifecycle and looks good to merge.

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5.4`*

PR #884, opened 2026-03-24
bug: extract_json_block returns None for entire response when any intermediate ```json block is malformed

### LINKED ISSUE #878: bug: extract_json_block returns None for entire response when any intermediate ```json block is malformed
## Bug

`extract_json_block()` in `src/engine/runner/response.rs` uses the `?` operator to propagate inner `Option::None` values, which returns `None` from the **entire function** instead of just skipping the malformed block and continuing the search.

## Location

`src/engine/runner/response.rs`, lines 434–451

```rust
fn extract_json_block(text: &str) -> Option<String> {
    let mut search_from = 0;
    while let Some(start) = text[search_from..].find("```json") {
        let abs_start = search_from + start;
        let after_tag = abs_start + "```json".len();
        let newline_pos = text[after_tag..].find('\n')?;  // BUG: exits function, not loop
        let content_start = after_tag + newline_pos + 1;
        let end = text[content_start..]
            .find("```")
            .map(|e| content_start + e)?;             // BUG: exits function, not loop
        let content = text[content_start..end].trim();
        if content.starts_with('{') {
            return Some(content.to_string());
        }
        search_from = end + 3;
    }
    None
}
```

The `?` at line 439 and the `.map(...)?` at line 442–443 both propagate `None` **out of the function**, not just out of the current loop iteration.

## Impact

If an agent response contains **any** ` ```json ` marker without:
- a newline immediately after it (e.g. ` ```json{...} `), or
- a closing ` ``` `,

…then the function returns `None` for the entire response — even if a valid, well-formed JSON block exists later in the text.

This causes the router to fail to parse valid routing responses from agents that happen to include any informal code block before the structured JSON output.

## Fix

Replace `?` with `else { continue }` (or equivalent) so malformed blocks are skipped:

```rust
let Some(newline_pos) = text[after_tag..].find('\n') else { 
    search_from = after_tag;
    continue; 
};
let Some(end_offset) = text[content_start..].find("```") else {
    search_from = content_start;
    continue;
};
let end = content_start + end_offset;
```

gabrielkoerich:
## Summary

The fix is already merged (commit `ebec58f`) and the branch is clean. This task is complete — the bug described in issue #878 was fixed in that commit by replacing the `?` operators with `else { continue }` to skip malformed JSON blocks instead of propagating `None` out of the function.

Closes #878

---
*Task #878 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/response.rs
+++ b/src/engine/runner/response.rs
@@ -436,11 +436,16 @@ fn extract_json_block(text: &str) -> Option<String> {
     while let Some(start) = text[search_from..].find("```json") {
         let abs_start = search_from + start;
         let after_tag = abs_start + "```json".len();
-        let newline_pos = text[after_tag..].find('\n')?;
-        let content_start = after_tag + newline_pos + 1;
-        let end = text[content_start..]
-            .find("```")
-            .map(|e| content_start + e)?;
+        let Some(rest) = text[after_tag..].strip_prefix('\n') else {
+            search_from = after_tag;
+            continue;
+        };
+        let content_start = text.len() - rest.len();
+        let Some(end_offset) = rest.find("```") else {
+            search_from = content_start;
+            continue;
+        };
+        let end = content_start + end_offset;
         let content = text[content_start..end].trim();
         if content.starts_with('{') {
             return Some(content.to_string());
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

gabrielkoerich:
## Automated Review — Approve

Fix correctly addresses the bug where malformed JSON fences caused the entire extract_json_block() function to return None. The replacement of ? operators with else { continue } blocks allows the parser to skip malformed blocks and continue searching for valid ones. All CI checks pass, new tests cover the fix.

---

PR #886, opened 2026-03-24
bug: runner script exits with exit_code=-1 and no diagnostic when worktree directory is missing (set -euo pipefail masks cd failure)

### LINKED ISSUE #874: bug: runner script exits with exit_code=-1 and no diagnostic when worktree directory is missing (set -euo pipefail masks cd failure)
## Problem

The agent runner script uses `set -euo pipefail` and then immediately does `cd "{work_dir}"`. If the worktree directory doesn't exist (e.g., cleanup race, disk issue, or corrupted state), `cd` fails and the script exits immediately — but `exit.txt` is never written, so `collect_output()` defaults to `exit_code = -1` with empty stdout/stderr. The engine then treats this as a generic unknown failure.

## Observed

From the logs, frequent pattern:
```
WARN: agent error, attempting recovery task_id="internal:8311" agent="opencode" error="unknown error (exit -1): "
INFO: failover: switching to fallback agent task_id="internal:8311" from="opencode" to="claude" error_type=Failed
```

The empty error message (`"unknown error (exit -1): "`) means stdout and stderr were both empty. One cause: `cd` failed silently before the agent ever ran. The fallback to claude succeeds because it tries again and the issue is transient, but the root cause is never surfaced.

## Root Cause

**`src/engine/runner/agent.rs` lines 108-128** — the runner script:

```bash
set -euo pipefail

export PATH=...
...
cd "{work_dir}"   # <-- if this fails, script exits with non-zero but writes no exit.txt

# Run agent
{agent_cmd} 2>"{attempt_dir}/stderr.txt" | tee "{output_file}"
CMD_STATUS=${PIPESTATUS[0]:-0}
echo "$CMD_STATUS" > "{status_file}"   # <-- never reached if cd failed
```

When `cd` fails, `exit.txt` is never written, so `collect_output()` in `session.rs` returns `exit_code = -1` with empty output.

## Fix

Add explicit error handling for the `cd` call:

```bash
cd "{work_dir}" || {{
    echo "worktree_missing" > "{status_file}"
    echo "worktree directory does not exist: {work_dir}" > "{attempt_dir}/stderr.txt"
    exit 1
}}
```

This gives:
- A non-empty `stderr.txt` with a clear message
- An `exit_code = 1` in `exit.txt` so the error is distinguishable from process-killed (-1)
- Better error classification: `classify_from_text(1, "", "worktree directory does not exist")" → more specific error

Similarly, protect the `mkdir -p .orch-opencode/opencode && echo ... > opencode.json` prefix (opencode permissions setup) with explicit error handling.

## Files

- `src/engine/runner/agent.rs`: update `build_runner_script()` (lines 100-139) to add explicit `cd` failure handling

gabrielkoerich:
## Summary

✅ **Tests verified.** Implementation is complete and all tests pass.

---

## Implementation complete. What would you like to do?

1. Merge back to main locally
2. Push and create a Pull Request
3. Keep the branch as-is (I'll handle it later)
4. Discard this work

Which option?

Closes #874

---
*Task #874 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/runner/agent.rs
+++ b/src/engine/runner/agent.rs
@@ -103,9 +103,14 @@ pub async fn spawn_in_tmux(tmux: &TmuxManager, inv: &AgentInvocation) -> anyhow:
         attempt_dir: &std::path::Path,
     ) -> anyhow::Result<String> {
         let status_file = attempt_dir.join("exit.txt");
+        let stderr_file = attempt_dir.join("stderr.txt");
         Ok(format!(
             r#"#!/usr/bin/env bash
 set -euo pipefail
+
+status_file="{status_file}"
+stderr_file="{stderr_file}"
+trap 'status=$?; printf "%s\n" "$status" > "$status_file"' EXIT
 
 # Environment — ~/.path and ~/.private are loaded by orch at startup into the process env
 export PATH="/opt/homebrew/bin:/usr/local/bin:$PATH"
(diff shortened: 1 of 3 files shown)


gabrielkoerich:
## Automated Review — Approve

Implementation correctly fixes the issue where set -euo pipefail would silently exit when cd fails, resulting in exit_code=-1 with no diagnostic. The solution uses trap to guarantee exit status persistence, explicit error handling for cd failures, and adds pattern detection for worktree/setup failures. All tests pass and code follows existing patterns.

---

PR #943, opened 2026-03-24
Bug: Agent timeout does not trigger failover logic

### LINKED ISSUE #939: Bug: Agent timeout does not trigger failover logic
## Summary

When an agent times out, the runner correctly identifies the timeout and returns a `RetryableError::Timeout`. However, the error handling logic in `src/engine/runner/fallback.rs` does not properly trigger the failover and rerouting mechanism. Instead of attempting to reroute to a different agent, it logs the timeout and the task gets stuck.

## Location

- **File**: `src/engine/runner/fallback.rs`
- **Lines**: 54-57

## Problem

The match arm for `agents::AgentError::Timeout` correctly identifies the timeout and constructs an error message. However, the subsequent call to `response::handle_failover` does not result in the task being rerouted.

The problem seems to be that while the timeout is correctly identified as a `RetryableError::Timeout`, the `handle_failover` function does not lead to a state change that the engine's dispatch loop will pick up for re-routing. The task's status may be updated, but the agent is not changed, so on the next attempt, the same agent is likely to time out again.

## Suggested Fix

The failover logic for timeouts needs to be more robust. When a timeout occurs, it should be treated as a strong signal to try a different agent.

The `handle_error` function in `fallback.rs` should be updated to ensure that after a timeout, it explicitly sets the task to be re-routed, possibly to a different, faster agent if the routing logic supports it.

Specifically, the outcome of `response::handle_failover` when called with `RetryableError::Timeout` should be a new status of `'new'` and a different agent assigned to the task. If this is not happening, the logic inside `handle_failover` needs to be adjusted.

gabrielkoerich:
Resolves #939

### DIFF
--- a/src/engine/runner/fallback.rs
+++ b/src/engine/runner/fallback.rs
@@ -161,6 +161,22 @@ pub async fn handle_error(
         }
     }
 
+    if retryable == response::RetryableError::Timeout {
+        let status = response::handle_timeout_failover(
+            task_id,
+            agent_name,
+            &error_msg,
+            store,
+            repo,
+        )
+        .await;
+        if status == "new" {
+            // A new agent was found, so we can return early.
+            return Ok(ErrorHandleResult::EarlyReturn { status });
+        }
+        // If no new agent was found, we fall through to the normal error handling.
+    }
+
     // Try free models as last resort before giving up
     let chain = response::get_reroute_chain(task_id, store, repo).await;
     let available: Vec<String> = ["claude", "codex", "opencode", "kimi", "minimax"]
--- a/src/engine/runner/response.rs
+++ b/src/engine/runner/response.rs
@@ -267,6 +267,54 @@ pub async fn handle_failover(
 }
 
 /// Get the reroute chain from store.
+pub async fn handle_timeout_failover(
+    task_id: &str,
+    agent_name: &str,
+    error_message: &str,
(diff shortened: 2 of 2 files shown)


gabrielkoerich:
## Automated Review — Approve

CI passes on GitHub (test, secrets, review-gate all green). Local test failure in `engine_config_from_config_uses_defaults_when_no_config` is a pre-existing issue — the test reads `max_parallel` from `~/.orch/config.yml` (value: 8) instead of using a test fixture, so it asserts 4. This is unrelated to the PR's changes. The fix correctly routes timeout errors to a different agent via `pick_timeout_fallback_agent`, which ignores the reroute chain and respects cooldowns. The `is_exhausted` guard (all agents in chain) still applies before the timeout-specific routing logic.

---

gabrielkoerich:
## Automated Review — Approve

The fix correctly addresses the bug: timeouts now trigger failover to a different healthy agent. Added pick_timeout_fallback_agent() which ignores the reroute chain (appropriate since timeouts are a strong signal the current agent can't handle the task), updated handle_failover() to use it for timeouts, and added handle_timeout_failover() for the early-return path in fallback.rs. All 892 tests pass, clippy passes, formatting passes. One minor issue: handle_timeout_failover has a copy-paste doc comment ('Get the reroute chain from store') that should describe what it actually does.

---

PR #1002, opened 2026-03-26
bug: auto_merge.rs rebase command uses sh -c with unquoted default_branch — shell injection via config

### LINKED ISSUE #999: bug: auto_merge.rs rebase command uses sh -c with unquoted default_branch — shell injection via config
## Summary

In `src/engine/auto_merge.rs`, the merge-conflict rebase fallback constructs a shell command via `sh -c` and interpolates `default_branch` (from config) and `wt` (worktree path from store) without safe quoting. A maliciously crafted config value or worktree path could inject arbitrary shell commands.

## Location

**File:** `src/engine/auto_merge.rs`, lines 404–411

```rust
let rebase_result = tokio::process::Command::new("sh")
    .arg("-c")
    .arg(format!(
        "cd '{}' && git fetch origin && git -c commit.gpgsign=false rebase origin/{default_branch} && git push --force-with-lease",
        wt  // single-quoted but any ' in the path breaks it
        // default_branch is NOT quoted at all
    ))
    .output()
    .await;
```

## Issues

1. **`default_branch` is completely unquoted.** Value comes from `config::get("gh.default_branch")`. If set to `main; <command>`, the shell executes the injected command.

2. **`wt` (worktree path) is single-quoted but single quotes are not escaped.** If the path somehow contains a `'` character (e.g. the home directory path or `ORCH_WORKTREES` env var contains one), the quoting breaks and the remaining content executes as shell code.

## Why This Matters

Although both inputs require write access to the local config or a compromised SQLite store (low external attack surface), using `sh -c` with string interpolation is an unsafe construction pattern. If either value is ever sourced from less-trusted input in the future, the injection surface grows silently.

## Fix

Replace the `sh -c` invocation with a structured multi-step `tokio::process::Command` sequence — the same approach used everywhere else in the codebase (e.g. `git_ops.rs::rebase_on_default`):

```rust
// fetch
Command::new("git").args(["fetch", "origin"]).current_dir(&wt_path).output().await?;
// rebase
Command::new("git")
    .args(["-c", "commit.gpgsign=false", "rebase", &format!("origin/{default_branch}")])
    .current_dir(&wt_path).output().await?;
// push
Command::new("git")
    .args(["push", "--force-with-lease"])
    .current_dir(&wt_path).output().await?;
```

The `git_ops.rs::rebase_on_default` function already implements safe structured git calls (including stash logic) and could be reused here.

gabrielkoerich:
## Summary

Removed the shell-based rebase fallback and replaced it with structured git commands to eliminate injection risk.

### What was done

- Reworked the merge-conflict recovery path in `src/engine/auto_merge.rs` to use `git fetch`, `git rebase`, and `git push` via `tokio::process::Command` instead of `sh -c`
- Kept the existing auto-merge and cleanup behavior intact after a successful rebase
- Validated the change with `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo nextest run`
- Committed the fix as `38f174e` (`fix: remove shell rebase fallback injection`)

Closes #999

---
*Task #999 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5.4-mini`*

### DIFF
--- a/src/engine/auto_merge.rs
+++ b/src/engine/auto_merge.rs
@@ -401,53 +401,92 @@ pub(crate) async fn auto_merge_pr(
                     );
                     let default_branch =
                         config::get("gh.default_branch").unwrap_or_else(|_| "main".to_string());
-                    let rebase_result = tokio::process::Command::new("sh")
-                        .arg("-c")
-                        .arg(format!(
-                            "cd '{}' && git fetch origin && git -c commit.gpgsign=false rebase origin/{default_branch} && git push --force-with-lease",
-                            wt
-                        ))
+                    let fetch_result = tokio::process::Command::new("git")
+                        .args(["fetch", "origin"])
+                        .current_dir(&wt_path)
                         .output()
                         .await;
 
+                    let rebase_result = match fetch_result {
+                        Ok(out) if out.status.success() => {
+                            tokio::process::Command::new("git")
+                                .args([
+                                    "-c",
+                                    "commit.gpgsign=false",
+                                    "rebase",
+                                    &format!("origin/{default_branch}"),
+                                ])
+                                .current_dir(&wt_path)
+                                .output()
+                                .await
+                        }
+                        Ok(out) => Ok(out),
+                        Err(err) => Err(err),
+                    };
+
                     match rebase_result {
                         Ok(out) if out.status.success() => {
-                            tracing::info!(
+                            let push_result = tokio::process::Command::new("git")
+                                .args(["push", "--force-with-lease"])
+                                .current_dir(&wt_path)
+                                .output()
+                                .await;
+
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

Rebased onto origin/main cleanly. The shell-injection fix correctly replaces `sh -c` with structured `tokio::process::Command` calls using `current_dir(&wt_path)` and properly-quoted argument lists. Each step (fetch, rebase, push) is now handled independently with correct error paths. Bonus: push failures are now properly handled instead of being silently ignored. The 864dc02 semaphore fix is preserved inside the loop. All 926 nextest tests pass; GitHub CI checks pass.

---

PR #1003, opened 2026-03-26
bug: review_poll.rs never increments review_cycles — max_review_cycles guard bypassed for human CHANGES_REQUESTED reviews

### LINKED ISSUE #998: bug: review_poll.rs never increments review_cycles — max_review_cycles guard bypassed for human CHANGES_REQUESTED reviews
## Summary

`review_open_prs` in `src/engine/review_poll.rs` re-dispatches agents (sets `Status::Routed`) when a human reviewer posts CHANGES_REQUESTED on a PR. However, it **never increments `review_cycles`** and never checks `max_review_cycles`, so the guard that prevents infinite re-dispatch loops is completely bypassed for human-review-driven changes.

## Root Cause

`review_poll.rs:452` sets `Status::Routed` directly without going through `handle_review_changes` (in `auto_merge.rs`). Only `handle_review_changes` increments the counter and enforces the cap:

**File:** `src/engine/review_poll.rs`, lines 429–460

```rust
// If we have new review feedback, store it and re-dispatch the task
if !review_context.is_empty() {
    store_set(/* pr_review_context */ ...).await;
    store_set(/* last_review_ts */ ...).await;

    if let Err(e) = task_manager
        .update_task_status(&task.id, Status::Routed)   // ← no review_cycles increment
        .await
    { ... }
}
```

Compare with `handle_review_changes` (`auto_merge.rs`, lines 554–634) which:
1. Reads `review_cycles` from store
2. Compares against `max_review_cycles`
3. Blocks the task with a human-escalation comment when exceeded
4. Increments `review_cycles` before re-routing

`review_poll.rs` does none of those steps.

## Impact

A human reviewer repeatedly posting CHANGES_REQUESTED will trigger an infinite loop:
- `InReview` → sync tick finds CHANGES_REQUESTED → `Routed` → agent runs → PR updated → `InReview` → repeat

No `max_review_cycles` cap applies. The task never gets blocked for human intervention.

## Fix

Before setting `Status::Routed`, `review_poll.rs` should:
1. Read the current `review_cycles` from the store
2. Compare against `max_review_cycles` (from config, default 2)
3. If exceeded, set `Status::Blocked` and post an escalation comment (same logic as `handle_review_changes`)
4. If not exceeded, increment `review_cycles` in the store before setting `Status::Routed`

Alternatively, `review_poll.rs` could call `handle_review_changes` directly for the human-review path, which already encapsulates all of this logic. The `notes` parameter would be `review_context` and `issues` would be empty `&[]`.

gabrielkoerich:
## Summary

Fixed review polling so human CHANGES_REQUESTED reviews now go through the shared review-change flow and respect the review-cycle cap.

### What was done

- Updated `src/engine/review_poll.rs` to call `handle_review_changes` instead of directly routing the task.
- Preserved review context/timestamp handling while reusing the existing max-cycle enforcement path.
- Verified formatting, clippy, and the full test suite pass.
- Committed the fix as `1a05629` (`fix: enforce review cycle cap in review poll`).

Closes #998

---
*Task #998 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5.4-mini`*

### DIFF
--- a/src/engine/review_poll.rs
+++ b/src/engine/review_poll.rs
@@ -5,7 +5,7 @@
 //! are requested.
 
 use crate::backends::{ExternalBackend, Status};
-use crate::engine::auto_merge::{dedup_reviews, MAX_MERGE_CONFLICT_RETRIES};
+use crate::engine::auto_merge::{dedup_reviews, handle_review_changes, MAX_MERGE_CONFLICT_RETRIES};
 use crate::engine::tasks::TaskManager;
 use crate::engine::EngineConfig;
 use crate::github::http::GhHttp;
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

Fix correctly routes human CHANGES_REQUESTED reviews through handle_review_changes, ensuring review_cycles is incremented and max_review_cycles is enforced. All CI checks pass (926 tests). Code is clean and follows existing patterns.

---

PR #1012, opened 2026-03-26
bug: review agent dispatched again after PR already approved — duplicate reviews

### LINKED ISSUE #1009: bug: review agent dispatched again after PR already approved — duplicate reviews
## Problem

PR #994 shows this sequence:
1. 01:19 — kimi/opus review started
2. 01:33 — claude/sonnet review started (second reviewer on same PR)
3. 01:35 — claude/sonnet **approved**
4. 01:37 — kimi/opus **started again** after approval

Two problems:
- Multiple review agents dispatched concurrently for the same task
- A new review dispatched after the PR was already approved

## Root cause

The `needs_review → in_review` label transition is the atomic guard, but there's a race window between the event subscriber (`subscribers/review.rs`) and the tick-based review trigger (`tick.rs:680`). Both can see `needs_review` before either updates the status to `in_review`.

After approval, if the status update back to `done` is slow (or review_poll resets to `needs_review` for any reason), another review gets dispatched.

The current `dispatch_key` guard is an in-memory `HashSet` that only prevents races within a single code path. It doesn't protect against concurrent access from the event subscriber, the tick loop, and `review_poll` — all three can independently decide to dispatch a review for the same task.

## Proposed fix

The review dispatch needs a proper lock at the task level, not just the in-memory dispatch set. Options:

1. **SQLite-based lock** — add a `review_locked_at` timestamp column to the task store. Before dispatching a review, atomically set it (UPDATE ... WHERE review_locked_at IS NULL). If the row was already locked, skip. Clear on completion. This survives restarts and works across all code paths.

2. **Single review entry point** — remove the tick-based review trigger entirely (tick.rs:680). The event subscriber is already the intended sole trigger (sync.rs:104 says so). The tick path is redundant and creates the race. The sync tick already handles stale `in_review` recovery, so the event subscriber + sync recovery should be sufficient.

3. **Check PR comments before dispatching** — before starting a review, check if there's already an "Automated Review — Approve" comment on the PR. If approved, skip. This is a defensive check on top of the lock, not a replacement.

Option 2 is the simplest and aligns with the documented intent. Option 1 is more robust if we need the tick fallback for reliability.

## Expected behavior

- Only one review agent should run per task at a time
- Once approved, no further reviews should be dispatched
- The lock must be visible across event subscriber, tick, and review_poll code paths

## Key files

- `src/engine/subscribers/review.rs` — event-driven NeedsReview trigger
- `src/engine/tick.rs:680` — tick-based NeedsReview trigger  
- `src/engine/review.rs` — review agent orchestration
- `src/engine/sync.rs:104` — comment says NeedsReview→InReview is handled exclusively by event subscriber, but tick.rs also does it

gabrielkoerich:
## Summary

Removed the tick-based review spawn path so `NeedsReview` now flows through the event-driven subscriber only, eliminating the duplicate review dispatch race.

### What was done

- Deleted the review-agent spawn branch from `src/engine/tick.rs` so tick only emits `NeedsReview` events
- Kept the event subscriber as the sole `NeedsReview -> InReview` transition point
- Ran `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo nextest run`
- Committed the change as `6fcff38`

Closes #1009

---
*Task #1009 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5.4-mini`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -23,8 +23,6 @@ use crate::tmux::TmuxManager;
 use std::sync::Arc;
 use tokio::sync::{mpsc, RwLock, Semaphore};
 
-use super::dispatch_guard::DispatchGuard;
-use super::review::{review_and_merge, ReviewDecision, MAX_REVIEW_AGENT_FAILURES};
 use super::EngineConfig;
 use crate::store::{review_session_expected, set_review_session_expected};
 
@@ -518,7 +516,7 @@ pub(crate) async fn tick_dispatch_tasks(
     semaphore: &Arc<Semaphore>,
     task_manager: &Arc<TaskManager>,
     weight_tx: &mpsc::Sender<WeightSignal>,
-    router_arc: &Arc<RwLock<Router>>,
+    _router_arc: &Arc<RwLock<Router>>,
     dispatching: &Arc<std::sync::Mutex<std::collections::HashSet<String>>>,
     store: &Arc<TaskStore>,
 ) -> anyhow::Result<()> {
@@ -604,7 +602,6 @@ pub(crate) async fn tick_dispatch_tasks(
         let backend = backend.clone();
         let tmux = tmux.clone();
         let capture = capture.clone();
-        let router_clone = router_arc.clone();
         let task_id_for_cleanup = task_id.clone();
         let task_owned = task.clone();
         let weight_tx = weight_tx.clone();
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

PR #1077, opened 2026-03-26
bug: task runner dispatch doesn't use DispatchGuard — key leaks on panic, blocking re-dispatch

### LINKED ISSUE #1072: bug: task runner dispatch doesn't use DispatchGuard — key leaks on panic, blocking re-dispatch
## Problem

`src/engine/tick.rs` spawns the runner task and relies on a **manual** `guard.remove(&dispatch_key_for_cleanup)` at the end of the spawned block (line 772-773) to release the per-task dispatch key.

If the spawned block panics before reaching that line, Tokio catches the panic and terminates the task without executing the cleanup. The key stays in the `dispatching` set permanently until the service restarts, preventing any re-dispatch of that task.

## Root cause

`src/engine/dispatch_guard.rs` was written specifically to fix this class of bug (RAII guard whose `Drop` removes the key even on panic). The module doc comment explains the problem and has a test in `src/engine/subscribers/review.rs:1032` (`dispatch_guard_releases_key_on_panic`). The guard IS used in `src/engine/subscribers/review.rs` (lines 132-142), but was **not applied** to `tick.rs`.

## Impact

Any panic in the runner dispatch path (e.g., unexpected `unwrap()` inside `runner.run_with_context()` or deeper) leaves the task stuck: the engine skips it on every subsequent tick (key present in `dispatching`), and the stuck-task recovery loop resets its status to `NeedsReview` repeatedly. Recovery requires a service restart.

## Fix

In `src/engine/tick.rs`, after inserting the dispatch key (around line 551-560), create a `DispatchGuard` and move it into the spawned block:

```rust
// After inserting the key:
let dispatch_guard = crate::engine::dispatch_guard::DispatchGuard::new(
    Arc::clone(&dispatching),
    dispatch_key.clone(),
);

tokio::spawn(REPO_CONTEXT.scope(repo_ctx, async move {
    let _dispatch_guard = dispatch_guard; // released on drop (normal or panic)
    // ... remove the manual guard.remove(&dispatch_key_for_cleanup) at line 772-773
    // ...
}));
```

Also remove the now-redundant manual removes at lines 565-567 (session exists check) and 575-576 (semaphore full) — replace those with early returns that let the guard's `Drop` handle cleanup (or keep the explicit removes since the guard hasn't been moved into the spawn yet at those points).

gabrielkoerich:
Resolves #1072

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -12,6 +12,7 @@
 use crate::backends::{ExternalBackend, ExternalId, ExternalTask, Status};
 use crate::channels::capture::CaptureService;
 use crate::config;
+use crate::engine::dispatch_guard::DispatchGuard;
 use crate::engine::jobs;
 use crate::engine::router::{get_route_result, Router};
 use crate::engine::runner::{TaskRunner, WeightSignal};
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

PR #1104, opened 2026-03-27
bug: cron normalize_dow maps "0-5" to "1-5" (Mon-Fri) — Sunday silently dropped from ranges starting at 0

### LINKED ISSUE #1094: bug: cron normalize_dow maps "0-5" to "1-5" (Mon-Fri) — Sunday silently dropped from ranges starting at 0
## Problem

`normalize_dow` in `src/cron.rs` converts DOW=0 (Sunday in standard cron) to DOW=7 for standalone values, but handles ranges starting at 0 incorrectly.

For a range like `"0-5"` (Sunday–Friday in standard cron), the function strips the `"0-"` prefix and prepends `"1-"`, producing `"1-5"`. In the cron crate's numbering (1=Mon … 6=Sat, 7=Sun), this fires **Monday–Friday only — Sunday is silently dropped**.

## File and line

`src/cron.rs` lines 28–30:

```rust
} else if let Some(rest) = part.strip_prefix("0-") {
    format!("1-{rest}")  // "0-5" → "1-5" (Mon–Fri) instead of Sun–Fri
}
```

The stale comment on line 22 still says `"0-5" → "7-5"` (the old broken behavior from #817) — the code was changed to produce a valid parse, but the semantic meaning was not preserved.

## Impact

Any scheduled job with a cron expression like `"0 9 * * 0-5"` (intended as Mon–Fri + Sunday) fires only Mon–Fri. The job silently skips its Sunday execution. The two tests added for `"0-5"` and `"0-4"` (lines 189–201) only assert `is_ok()` (parses without error), not that the correct days fire.

## Fix

A Sunday-anchored range cannot be expressed as a single contiguous range in the cron crate because Sunday=7 is numerically higher than any weekday (1–6). The correct conversion is to split it:

- `"0-5"` → `"1-5,7"` (Mon–Fri + Sunday)
- `"0-6"` → `"1-6,7"` (Mon–Sat + Sunday, i.e. every day)
- General `"0-N"` (where N ≥ 1) → `"1-{N},7"`

Also update the comment on line 22 to reflect the new mapping, and add assertions (not just `is_ok()`) to verify the schedule fires on Sunday.

gabrielkoerich:
Resolves #1094

### DIFF
--- a/src/cron.rs
+++ b/src/cron.rs
@@ -19,14 +19,17 @@ pub fn normalize_dow(expression: &str) -> String {
         return expression.to_string();
     }
     let dow = fields[4];
-    // Replace 0 with 7 in the DOW field, handling ranges/lists (e.g. "0,3" → "7,3", "0-5" → "7-5")
+    // Replace 0 with 7 in the DOW field, handling ranges/lists.
+    // Examples: "0" → "7", "0,3" → "7,3", "0-5" → "1-5,7" (split: Mon-Fri + Sunday)
+    // Ranges starting at 0 cannot be expressed as a single contiguous range in the
+    // cron crate (Sun=7 > Sat=6), so we split: "0-N" → "1-N,7".
     let normalized_dow = dow
         .split(',')
         .map(|part| {
             if part == "0" {
                 "7".to_string()
             } else if let Some(rest) = part.strip_prefix("0-") {
-                format!("1-{rest}")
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
⚠️ Review agent failed (attempt 1/3)

**Reason:** parse error: failed to parse review response from output

Retrying with a different agent.

PR #1166, opened 2026-03-28
bug: model cooldown fallback ignores cooldown check — routes to failed model

### LINKED ISSUE #1163: bug: model cooldown fallback ignores cooldown check — routes to failed model
## Problem

When all models in a pool are in cooldown, `model_for_complexity()` falls back to `expanded_pool[0]` **without checking if that model is also in cooldown**. This causes the router to repeatedly pick the same failed model on every reroute, creating infinite reroute loops until the cooldown expires.

## Evidence

From `src/engine/router/config.rs:427-428`:
```rust
// All models cooled — deterministic fallback to first entry
Some(expanded_pool[0].clone())
```

Task #1155 shows this clearly:
- Attempt 1: `github-copilot/claude-sonnet-4.6` → silent → killed → model cooldown set (1h)
- Attempt 2-5: router keeps picking `github-copilot/claude-sonnet-4.6` via the fallback → silent → killed (4 more reroutes)
- Attempt 6: `opencode/minimax-m2.5-free` finally succeeds

Query showing the pattern:
```sql
SELECT task_id, model, COUNT(*) FROM task_metrics
WHERE completed_at >= datetime('now', '-7 days') AND outcome = 'rerouted'
AND model LIKE 'github-copilot%' GROUP BY task_id, model ORDER BY COUNT(*) DESC LIMIT 5;
```
Result: `gpt-5.4-mini` rerouted 79 times, `claude-sonnet-4.6` rerouted 40 times — many are the same model being retried repeatedly.

## Root Cause

In `src/engine/router/config.rs`, `model_for_complexity()`:
1. Builds expanded pool starting with non-free models (e.g., `github-copilot/claude-sonnet-4.6` at index 0)
2. Appends discovered free models
3. Walks pool skipping cooled models
4. Falls back to `expanded_pool[0]` when all are cooled — but `expanded_pool[0]` might also be in cooldown!

## Fix

The fallback should either:
- **(A)** Skip the first entry if it's also in cooldown, continuing to walk the pool
- **(B)** Return `None` when the fallback model is in cooldown, letting the caller decide (router will fall back to a different agent)
- **(C)** Track which model was the fallback and increment its weight so the router prefers fresh models

Option (B) is the cleanest: if all models are cooled, return `None` and let the router's agent-level fallback kick in.

## Impact

This bug causes unnecessary reroutes that waste compute and inflate the failure rate. Each reroute takes 2 minutes (silence detection timeout) before the next attempt. A task caught in this loop wastes 2N minutes where N is the number of cooldown-period reroutes.

## Files Likely Involved

- `src/engine/router/config.rs:427-428` — the fallback logic
- `src/engine/cooldown.rs` — cooldown tracking

gabrielkoerich:
## Summary

The commit is done. Since this is running as an orch-dispatched agent, the engine handles pushing and PR creation automatically — no further action needed from me.

The fix is committed: `827ccf2 fix: return None when all models in cooldown pool are exhausted`

Closes #1163

---
*Task #1163 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/router/config.rs
+++ b/src/engine/router/config.rs
@@ -411,9 +411,6 @@ impl RouterConfig {
         if expanded_pool.is_empty() {
             return None;
         }
-        if expanded_pool.len() == 1 {
-            return Some(expanded_pool[0].clone());
-        }
         // Random starting index — varies per task_id to distribute across the pool
         let start =
             crate::engine::router::selection::simple_hash_index_for(expanded_pool.len(), task_id);
@@ -424,8 +421,8 @@ impl RouterConfig {
                 return Some(model.clone());
             }
         }
-        // All models cooled — deterministic fallback to first entry
-        Some(expanded_pool[0].clone())
+        // All models cooled — return None so the caller can fall back to a different agent
+        None
     }
 
     /// Get the model for a given agent and complexity level, falling back to built-in defaults.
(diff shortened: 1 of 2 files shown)


gabrielkoerich:

---
*Reviewed codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.2-codex`*

PR #1197, opened 2026-03-28
fix: auto-recover tasks blocked by non-required CI check failures

### LINKED ISSUE #1192: fix: auto-recover tasks blocked by non-required CI check failures
## Problem

When the review agent requests changes based on non-required CI check failures (e.g. `review-gate`), the task hits `max_review_cycles` and gets blocked. This requires manual intervention (`orch task retry`) even though the block reason is invalid — the code is correct and required checks pass.

### Example: #1167

- Agent fix was correct, required CI checks (`test`, `secrets`) passed
- Review agent requested changes twice because `review-gate` (non-required) showed as failed
- Task blocked after hitting `max_review_cycles` (2)
- Had to be manually retried after #1187 fixed the review prompt

## Desired behavior

When a task is blocked due to max review cycles, the engine should check whether the block reason was actually valid:

1. If the task has an open PR with all **required** CI checks passing, and the review comments only cite non-required check failures as the reason for requesting changes → auto-retry instead of blocking
2. Alternatively: after the review prompt fix (#1187) lands and gets deployed, automatically re-review any tasks that were blocked while the old prompt was active — their review decisions may have been wrong

## Context

- #1187 fixed the root cause (review prompt now uses `--required` flag)
- But tasks blocked BEFORE that fix are stuck and need manual retry
- Future occurrences should be less likely, but a safety net for "blocked with passing required checks" would prevent this class of false blocks

gabrielkoerich:
## Summary

- Adds `required_ci_checks_pass()` helper in `auto_merge.rs` that fetches the PR's head SHA/base branch, gets required status check contexts for the branch, and evaluates them using the existing `required_checks_state()` function
- In `handle_review_changes()`, before escalating to `Blocked` when `review_cycles >= max_cycles`, checks if all required CI passes — if so, resets review cycles, increments `auto_unblock_count`, re-triggers `NeedsReview`, and posts a recovery comment on the PR
- Limits auto-recovery to 1 attempt (`ci_recovery_count < 1`) to prevent infinite loops
- Splits `opt_store_get_task` result into `task_store_record` variable so it can be reused for both `review_cycles` and `ci_recovery_count`
- Falls through to normal escalation on any API error or when required checks are genuinely failing

Closes #1192

## Test plan

- [ ] All 2064 existing tests pass (`cargo nextest run`)
- [ ] Task blocked due to `review-gate` non-required failure auto-recovers on next tick (review cycles reset, `NeedsReview` set, recovery comment posted on PR)
- [ ] Task with a genuinely failing required check still escalates to `Blocked` as before
- [ ] Second auto-recovery attempt is blocked (`ci_recovery_count >= 1` → falls through to escalation)
- [ ] API errors in `required_ci_checks_pass` return `None` and fall through to normal escalation

### DIFF
--- a/src/engine/auto_merge.rs
+++ b/src/engine/auto_merge.rs
@@ -686,6 +686,74 @@ async fn branch_newer_than_last_review(
     }
 }
 
+/// Check whether all *required* CI checks pass for the given PR.
+///
+/// Returns `Some(true)` when all required contexts report success, `Some(false)` when
+/// at least one required check is failing or pending, and `None` on any API error (so
+/// the caller can fall through to the default escalation path).
+async fn required_ci_checks_pass(gh: &GhHttp, repo: &str, pr_number: u64) -> Option<bool> {
+    let pr = match gh.get_pr(repo, pr_number).await {
+        Ok(pr) => pr,
+        Err(e) => {
+            tracing::warn!(pr_number, err = %e, "required_ci_checks_pass: failed to fetch PR");
+            return None;
+        }
+    };
+
+    let head_sha = pr.head.sha.clone();
+    let base_branch = pr.base.ref_.clone();
+
+    let required_contexts = match gh
+        .get_required_status_check_contexts(repo, &base_branch)
+        .await
+    {
+        Ok(ctx) => ctx,
+        Err(e) => {
+            tracing::warn!(
+                pr_number,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started (agent: claude, model: opus)

PR #1228, opened 2026-03-28
bug: parse_review_from_output skips keyword inference when NDJSON extraction succeeds but JSON parse fails

### LINKED ISSUE #1221: bug: parse_review_from_output skips keyword inference when NDJSON extraction succeeds but JSON parse fails
## Problem

`parse_review_from_output` in `src/engine/runner/response.rs` has three steps:

1. Direct JSON / markdown parse of raw output  
2. NDJSON extraction → JSON/markdown parse of extracted text  
3. Keyword inference fallback (LGTM, looks good, etc.)

**Bug:** Step 3 is only reached when step 2 yields an *empty* extraction. When NDJSON extraction succeeds (non-empty) but the extracted text isn't structured JSON, the function returns `Err` without trying keyword inference.

## Root Cause

```rust
// src/engine/runner/response.rs:333-337 (current — buggy)
let extracted = ndjson_extract_text(output);
if !extracted.is_empty() {
    return parse_review_response(&extracted)  // returns Err if not JSON
        .map_err(|e| anyhow::anyhow!("parse failed after NDJSON extraction: {e}"));
    // Step 3 (keyword inference) is never reached
}
```

## Impact

An opencode review agent that emits a plain-text approval as NDJSON:

```
{"type":"text","text":"LGTM, all checks pass, no issues found."}
```

Would have its text extracted correctly (`"LGTM, all checks pass, no issues found."`) but `parse_review_response` fails on that text, so the function errors with `"parse failed after NDJSON extraction: ..."` instead of inferring `approve`. The review agent's session is treated as a failure and retried.

The `infer_review_response_from_text` function already handles this exact text ("all checks pass" → approve), but is never called.

## Fix

After NDJSON extraction, fall through to keyword inference on the extracted text before giving up:

```rust
let extracted = ndjson_extract_text(output);
if !extracted.is_empty() {
    if let Ok(resp) = parse_review_response(&extracted) {
        return Ok(resp);
    }
    // Extracted text may be plain-text (e.g. "LGTM") — try keyword inference
    if let Some(resp) = infer_review_response_from_text(&extracted) {
        tracing::warn!(output_len = output.len(), "review response parsed via keyword fallback on NDJSON text");
        return Ok(resp);
    }
    anyhow::bail!("parse failed after NDJSON extraction: no structured JSON or recognizable keywords");
}
```

Add a test case: NDJSON with a single `type:text` event containing "LGTM" (no JSON) should parse as `approve`.

## Files

- `src/engine/runner/response.rs` lines 332-337 (`parse_review_from_output`, step 2)
- Add test alongside the existing `parse_review_from_output_ndjson_*` tests (around line 889)

gabrielkoerich:
## Summary

Already handled — the fix is committed as `c5cb5a6`.

Closes #1221

---
*Task #1221 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/response.rs
+++ b/src/engine/runner/response.rs
@@ -332,8 +332,20 @@ pub fn parse_review_from_output(output: &str) -> anyhow::Result<ReviewResponse> 
     // Step 2: NDJSON — extract text events and parse the concatenated text
     let extracted = ndjson_extract_text(output);
     if !extracted.is_empty() {
-        return parse_review_response(&extracted)
-            .map_err(|e| anyhow::anyhow!("parse failed after NDJSON extraction: {e}"));
+        if let Ok(resp) = parse_review_response(&extracted) {
+            return Ok(resp);
+        }
+        // Extracted text may be plain-text (e.g. "LGTM") — try keyword inference
+        if let Some(resp) = infer_review_response_from_text(&extracted) {
+            tracing::warn!(
+                output_len = output.len(),
+                "review response parsed via keyword fallback on NDJSON text"
+            );
+            return Ok(resp);
+        }
+        anyhow::bail!(
+            "parse failed after NDJSON extraction: no structured JSON or recognizable keywords"
+        );
     }
 
     // Step 3: heuristic fallback for plain-text decisions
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

PR #1329, opened 2026-03-30
bug: opencode silence detection resets tasks to New instead of applying cooldown

### LINKED ISSUE #1320: bug: opencode silence detection resets tasks to New instead of applying cooldown
## What I Found

3 task runs in the last 24 hours failed with `silence detection reset task to New`:

| Task Run | Agent | Model | Timestamp |
|----------|-------|-------|-----------|
| internal:28255 | opencode | opencode/mimo-v2-omni-free | 2026-03-30T20:00:46Z |
| internal:28721 | opencode | opencode/mimo-v2-omni-free | 2026-03-30T18:06:33Z |
| (unnamed) | opencode | opencode/nemotron-3-super-free | 2026-03-30T17:51:34Z |

Service log evidence:
```
2026-03-30T20:02:54Z WARN engine.tick.phase1b.silence: agent silent since session start — killing session, cooling down model + agent, re-routing task_id=internal:28721 agent=opencode model=opencode/mimo-v2-omni-free grace_secs=120 cooldown_secs=3600
2026-03-30T20:00:46Z task_runs: opencode|opencode/mimo-v2-omni-free|failed|silence detection reset task to New
```

## Why It's Happening

The silence detection logic in `src/engine/tick.rs` (phase1b) detects when an agent produces no output since session start. When triggered, it:
1. Kills the tmux session
2. Records a cooldown for the model + agent
3. Resets the task status to `New`

The problem: resetting to `New` means the task goes through the full routing pipeline again, potentially getting the same agent/model that just failed. The cooldown should prevent re-selection, but the task loses all context about what attempt it was on.

## Why It Matters

- Tasks lose all progress when silence-detected (worktree still exists but task status resets)
- 3 failures in 24h from this alone
- If the same agent/model is selected again (cooldown expired or different model in pool), the same silence pattern repeats
- No distinction between "agent never started" vs "agent started but went silent mid-work"

## Suggested Fix

1. In `src/engine/tick.rs` silence detection handler, reset to `needs_review` instead of `new` — this preserves attempt count and lets the engine re-route with cooldown awareness
2. Alternatively, add a `silence_reset_count` to tasks and block after N silence resets
3. Log the worktree path when resetting so the next attempt can recover from existing state

## Evidence

SQL query:
```sql
SELECT agent, model, outcome, error, started_at FROM task_runs 
WHERE started_at > datetime('now', '-24 hours') AND error LIKE '%silence%'
ORDER BY started_at DESC;
```

Service log:
```
tail -500 /opt/homebrew/var/log/orch.log | grep 'silence'
```

gabrielkoerich:
## Summary

Let me understand the issue better by looking at how `route_attempts` is used and the runner's silence detection handling.I understand the issue. The silence detection in `tick_detect_silent_agents` resets `route_attempts` to 0 (line 216-217), which means the router loses all memory of previous routing attempts. This causes the same agent/model to potentially be selected again, creating a loop.

The fix: preserve `route_attempts` when silence detection triggers, so the router knows this task has

Closes #1320

---
*Task #1320 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -195,7 +195,9 @@ pub(crate) async fn tick_detect_silent_agents(
             );
         }
 
-        // 4. Clear routing state and reset to New for re-routing
+        // 4. Clear routing state and reset to New for re-routing.
+        // Preserve route_attempts so the router knows this task has been routed before
+        // and can fall back to round-robin after max_route_attempts failures.
         let task_eid = ExternalId(task_id.clone());
         if use_backend {
             if let Some(ref st) = store_task {
@@ -213,7 +215,6 @@ pub(crate) async fn tick_detect_silent_agents(
             &[
                 ("agent", serde_json::Value::Null),
                 ("model", serde_json::Value::Null),
-                ("route_attempts", serde_json::json!(0)),
             ],
         )
         .await;


gabrielkoerich:
## Automated Review — Approve

The PR fixes the silence-detection bug by preserving `route_attempts` when the engine resets a silent task (so the router can fall back after max attempts). Required CI checks passed. I reviewed related code paths (router, sync, tick, review subscriber) for regressions and kept an eye on scope creep. Changes are small and focused, but there are two behavioral changes worth calling out for the reviewers/maintainers to confirm they were intended.

---
*Reviewed opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

PR #1383, opened 2026-03-31
bug: internal tasks loop new→in_progress→needs_review→new when WeightSignal::None fires

### LINKED ISSUE #1380: bug: internal tasks loop new→in_progress→needs_review→new when WeightSignal::None fires
## Problem

Internal/planning tasks (task IDs starting with `internal:`) can get stuck in an infinite loop when the agent output fails to parse and `WeightSignal::None` is returned:

1. Agent runs, output parse fails → `WeightSignal::None`
2. `tick.rs:937` maps `WeightSignal::None` to `"needs_review"` for ALL tasks
3. Review cycle checks for PR → none (internal tasks never create PRs)
4. `review.rs:502–510` re-routes task to `Status::New`
5. Back to step 1 indefinitely

This was observed for task `internal:30378` in today's logs:
```
WARN orch::engine::tick: recovering stuck task: no session found — reclaiming early → new  task_id="internal:30378" age_mins=13
WARN orch::engine::review: no PR and no commits — re-routing for retry  task_id="internal:30378" branch=internal-30378-...
```

The loop terminates only if `last_error` contains specific strings (`"exceeded max attempts"`, `"no fallback agents"`, etc.). A fresh re-route resets `last_error`, so these guards never trigger.

## Root Cause

Two cooperating bugs:

**`src/engine/tick.rs:937`** — `WeightSignal::None` unconditionally routes to `needs_review`:
```rust
WeightSignal::None => "needs_review",
```
For internal tasks there is no PR to review, so this is always wrong.

**`src/engine/review.rs:502–510`** — "no PR and no commits" re-routes to `New` without a persistent counter guard. Unlike `response_handler.rs` (which uses `no_code_reroutes` counter + max), the review path has no such cap.

## Fix

**Option A (minimal):** In `tick.rs`, skip `needs_review` for internal tasks when signal is `None`:
```rust
WeightSignal::None => {
    if task_id.starts_with("internal:") { "done" } else { "needs_review" }
}
```

**Option B (more robust):** In `review.rs`, add a persistent reroute counter to the "no PR and no commits" branch (similar to `no_code_reroutes` in `response_handler.rs`). After N reroutes, mark `Blocked` instead of re-routing to `New`. This caps the loop regardless of task origin.

Option B is more robust because it also catches external tasks that get into the same loop without producing commits.

## Impact

Planning/analysis tasks dispatched by recurring jobs waste agent invocations and accumulate attempts without making progress. The task appears perpetually in-flight.

**Files:**
- `src/engine/tick.rs:937`  
- `src/engine/review.rs:502–510`  
- `src/engine/runner/response_handler.rs:438–452` (reference implementation for the counter pattern)

gabrielkoerich:
## Summary

Fixed internal tasks infinite loop by adding a circuit-breaker counter for no-PR reroutes in review.rs

### What was done

- Added persistent reroute counter (no_pr_reroutes) to review.rs preventing infinite new→in_progress→needs_review→new loops
- Counter uses workflow.max_reroute_attempts config (default: 3) matching the existing no_code_reroutes pattern
- After max reroutes, task is blocked with explanatory last_error message
- Agent/model cleared on each reroute so router picks a different agent

Closes #1380

---
*Task #1380 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/mimo-v2-omni-free`*

### DIFF
--- a/src/engine/review.rs
+++ b/src/engine/review.rs
@@ -23,6 +23,7 @@ fn review_started_comment(review_agent: &str, review_model: &str) -> String {
 
 use crate::backends::{ExternalBackend, ExternalTask, Status};
 use crate::cmd::CommandErrorContext;
+use crate::config;
 use crate::engine::auto_merge::{attribution_footer, auto_merge_pr, handle_review_changes};
 use crate::engine::runner;
 use crate::engine::runner::worktree;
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

---

PR #1456, opened 2026-04-01
chat: persistent tmux session fails at startup, breaks design

### LINKED ISSUE #1455: chat: persistent tmux session fails at startup, breaks design
## Problem

\`orch chat\` fails immediately with tmux session timeouts and disappearing sessions:

\`\`\`
2026-04-01T13:31:08.182659Z  INFO orch::chat_session: creating persistent chat session
2026-04-01T13:31:38.706280Z  WARN orch::chat_session: agent ready timeout — proceeding anyway
2026-04-01T13:31:39.849750Z  WARN orch::chat_session: tmux session disappeared — recreating
\`\`\`

### 1. Agent startup timing is broken (src/chat_session.rs:387-411)
- \`wait_for_agent_ready()\` waits max 30 seconds for prompt detection
- **Line 393-394: Times out and returns \`Ok(())\` anyway** — proceeds with an unready agent
- Agent never started, so subsequent message sends fail

### 2. Interactive command is malformed (line 365-369)
\`\`\`rust
format!("{agent} --model {model} --permission-mode bypassPermissions --append-system-prompt {system_prompt_file}")
\`\`\`
This passes the file path as a positional argument after \`--append-system-prompt\`, but Claude Code doesn't accept it this way. Agent fails to start or behaves unexpectedly.

### 3. Prompt detection is fragile (line 585-619)
- Looks for \`❯\`, \`>\`, specific prompt patterns
- If agent doesn't print a prompt (stuck, crashed), detection fails silently
- Code continues anyway, leading to pane capture failures

### 4. Sessions die and recreate infinitely (line 141-156)
- Malformed command causes agent to crash during startup
- Session disappears
- Code detects it missing and retries
- Creates infinite loop of failed attempts

## Design violation

**CLAUDE.md states:**
> Each message triggers a **one-shot agent invocation** with context assembled from SQLite. No long-running session — stateless process, stateful database.
> Agent invocations run via \`bash -c\` with timeout (45s), **not tmux**

The persistent tmux session in \`src/chat_session.rs\` violates this settled design. It adds complexity without benefit and is fundamentally broken.

## Recommendation

Revert to the designed one-shot approach:
- Each message → direct \`bash -c\` invocation (no tmux)
- Capture stdout/stderr directly (no pane diffing)
- Clear error handling with timeout
- Simpler, more reliable, matches documented design

## Files involved
- \`src/chat_session.rs\` — persistent session logic (should be removed/simplified)
- \`src/cli/chat.rs\` — entry point
- \`CLAUDE.md\` — documents the correct design

gabrielkoerich:
## Summary

Removed broken persistent tmux chat session (chat_session.rs) and restored the documented one-shot invocation design for orch chat

### What was done

- Deleted src/chat_session.rs (757 lines of broken persistent tmux session logic)
- Updated control.rs: removed invoke_agent_persistent, send_message now calls invoke_agent directly via run_direct()
- Removed /kill and /session control commands (only existed for persistent session management)
- Cleaned up unused session_id parameter from handle_control_command and maybe_handle_control_command
- Updated CLI help text to reflect available commands
- All 2378 tests pass, clippy clean, fmt clean

Closes #1455

---
*Task #1455 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*


gabrielkoerich:
Superseded by simpler fix: the issue was just a malformed flag (--append-system-prompt → --append-system-prompt-file). Persistent tmux sessions are the correct design for performance (~2s vs 15-20s cold start). No need to remove them.

PR #1479, opened 2026-04-01
bug: fallback.rs calls blocking subprocess in async context — reactor stall risk

### LINKED ISSUE #1473: bug: fallback.rs calls blocking subprocess in async context — reactor stall risk
## Problem

`src/engine/runner/fallback.rs` line 121 calls `agent_runner.free_models()` from within the async function `handle_error()`. For OpenCode agents, this chain is:

```
async handle_error()                      [fallback.rs:27]
  → agent_runner.free_models()            [fallback.rs:121]
    → discover_free_models_cached()        [opencode.rs:348]
      → discover_free_models()             [opencode.rs:674]
        → std::process::Command::new("opencode").output()  // BLOCKING I/O
```

The comment in `opencode.rs` at line 681-683 explicitly warns:
```rust
// Note: this function is sync; callers in async contexts should wrap with
// tokio::task::spawn_blocking() at the call site.
```

But `fallback.rs` does not wrap the call with `spawn_blocking`.

## Impact

If the `opencode models` subprocess hangs (e.g., network timeout, slow startup), it blocks the Tokio worker thread, preventing other async tasks from making progress on that thread. Under high load with multiple concurrent fallback errors, this can stall the reactor.

## Fix

In `fallback.rs`, wrap the call with `spawn_blocking`:

```rust
let runner = agent_runner.clone();  // or Arc clone
let free = tokio::task::spawn_blocking(move || runner.free_models())
    .await
    .unwrap_or_default();
```

Alternatively, cache the model list at runner construction time (already done for the 1-hour cache) and expose it through a non-blocking accessor that reads the cache without re-discovering.

gabrielkoerich:
## Summary

Fixed blocking subprocess call in async context by making discover_free_models_cached() non-blocking

### What was done

- Changed free_models_cache field from Mutex<...> to Arc<Mutex<...>> to allow sharing with background threads
- Added free_models_refresh_in_progress AtomicBool to prevent duplicate background refreshes
- Rewrote discover_free_models_cached() to return cached data (fast path) or known defaults immediately on cold cache, spawning a std::thread for background refresh
- Extracted known_free_models() helper function from discover_free_models()
- Added FreeModelsCache type alias to satisfy clippy complexity lint
- All 2336 tests pass, fmt and clippy clean

Closes #1473

---
*Task #1473 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/agents/opencode.rs
+++ b/src/engine/runner/agents/opencode.rs
@@ -30,12 +30,17 @@ use super::{AgentError, AgentRunner, ParsedResponse, PermissionRules};
 use crate::cmd::SyncCommandErrorContext;
 use crate::parser;
 
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
+
+type FreeModelsCache = Arc<Mutex<Option<(Vec<String>, std::time::Instant)>>>;
 
 /// Runner for OpenCode agent.
 pub struct OpenCodeRunner {
     /// Cached free models (model list + timestamp).
-    free_models_cache: Mutex<Option<(Vec<String>, std::time::Instant)>>,
+    free_models_cache: FreeModelsCache,
+    /// True while a background thread is refreshing the free-models cache.
+    /// Guards against spawning duplicate refresh threads.
+    free_models_refresh_in_progress: Arc<std::sync::atomic::AtomicBool>,
 }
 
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

PR #1482, opened 2026-04-01
bug: batch_is_pr_merged_by_branch GraphQL query has invalid escaping — cleanup fails every tick

### LINKED ISSUE #1480: bug: batch_is_pr_merged_by_branch GraphQL query has invalid escaping — cleanup fails every tick
## Summary

`batch_is_pr_merged_by_branch()` in `src/github/http.rs:1158` uses `\"` inside a raw string (`r#"..."#`), which produces literal backslash characters in the GraphQL query. This makes every PR-merge-check query syntactically invalid, causing `"batch PR merge check failed, skipping tick"` to fire every tick.

## Root Cause

Line 1158 constructs the query alias:
```rust
r#"b{i}: pullRequests(headRefName: \"{escaped}\", states: [MERGED], last: 1) {{ nodes {{ merged }} }}"#
```

In a raw string, `\"` is a literal `\` + `"` (no escape processing). This query then goes through `self.graphql(&query)` → `graphql_request()` which wraps it in `serde_json::json!({ "query": query })`. The serde serializer already escapes `"` → `\"` in JSON. The extra `\` from the raw string becomes `\\\"` on the wire — GitHub parses this as a literal backslash in the GraphQL query, which is invalid.

**Compare with working code:** Line 1346 (`batch_is_pr_merged_by_number`) also goes through `self.graphql()` and correctly uses `"` without backslashes in its raw string:
```rust
format!(r#"{{ repository(owner: "{owner}", name: "{name}") {{ {aliases} }} }}"#)
```

Lines 1886/1906 DO use `\"` but they construct the full JSON body directly via `.body()`, not via `serde_json::json!()`, so the `\"` is correct there.

## Impact

- Merged-PR cleanup phase is completely broken — tasks stay in `needs_review`/`in_review` even after their PRs are merged
- WARN log fires every tick (~60s), adding noise to logs
- Stale review tasks cause the sync catch-up to repeatedly re-fire NeedsReview events

## Fix

Remove the backslashes from the raw string on line 1158:
```rust
format!(
    r#"b{i}: pullRequests(headRefName: "{escaped}", states: [MERGED], last: 1) {{ nodes {{ merged }} }}"#
)
```

Also, the `escaped` variable (line 1156) does `branch.replace('"', "\\\"")` — with the fix, this should become `branch.replace('"', "\\\"")` to keep quote-escaping correct for the GraphQL layer. However, GitHub branch names cannot contain `"`, so this is defensive only.

## Files

- `src/github/http.rs:1152-1165` — the broken query construction
- `src/engine/cleanup.rs:902-907` — where the error is caught and logged

## Verification

After fix, the WARN log `"batch PR merge check failed, skipping tick"` should stop appearing. Merged PRs should trigger task completion in the cleanup phase.

gabrielkoerich:
## Summary

Now I understand the full picture. The `graphql()` method wraps the query in `serde_json::json!({ "query": query })` at line 785, which handles JSON escaping automatically. So the raw string on line 1158 should use plain `"` not `\"`.

Let me also check the `escaped` variable — it does `branch.replace('"', "\\\"")` which in a regular string produces `\"`. Since the query goes through serde JSON serialization, we actually don't need to pre-escape quotes at all — serde will do it. But let me k

Closes #1480

---
*Task #1480 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

### DIFF
--- a/src/github/http.rs
+++ b/src/github/http.rs
@@ -1155,7 +1155,7 @@ impl GhHttp {
             .map(|(i, branch)| {
                 let escaped = branch.replace('\\', "\\\\").replace('"', "\\\"");
                 format!(
-                    r#"b{i}: pullRequests(headRefName: \"{escaped}\", states: [MERGED], last: 1) {{ nodes {{ merged }} }}"#
+                    r#"b{i}: pullRequests(headRefName: "{escaped}", states: [MERGED], last: 1) {{ nodes {{ merged }} }}"#
                 )
             })
             .collect::<Vec<_>>()


gabrielkoerich:
## Automated Review — Approve

Required CI checks pass. One-line fix that removes invalid escaping from a raw string — `"` in `r#"..."#` produced literal backslashes, double-escaped by serde_json to `\\"` in the JSON body, making GraphQL queries invalid. The fix uses plain `"` in the raw string and lets serde_json handle proper JSON escaping, consistent with the working `batch_is_pr_merged_by_number` at line 1346. The `escaped` variable (line 1156) is now a no-op for real branch names but causes no harm.

---

PR #1498, opened 2026-04-02
Panics and expect/unwrap usage in core engine modules (tasks/sync/dispatch_guard)

### LINKED ISSUE #1496: Panics and expect/unwrap usage in core engine modules (tasks/sync/dispatch_guard)
src/cron.rs\nExample: line ~310 (Schedule::from_str(&full_expr).expect("should parse");)\n\nWhat\'s wrong:\n- The cron schedule parsing uses .expect() which panics on malformed schedule strings. This can crash startup or runtime checks.\n\nSuggested fix:\n- Replace .expect() with match/Result propagation and surface a clear error to operator; add unit tests for invalid inputs.

gabrielkoerich:
## Summary

The exploration found most `.expect()`/`.unwrap()` usage is in test code or well-guarded invariants. But the task specifically mentions `cron.rs` line ~310 with `Schedule::from_str(&full_expr).expect("should parse")` — let me check that directly.The `cron.rs:310` `.expect()` is actually in test code (line 310 is inside `#[cfg(test)]`), and the production `check()` function at line 119-120 already properly uses `.with_context()` + `?`. Let me check the other files mentioned in the task.Let me s

Closes #1496

---
*Task #1496 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

### DIFF
--- a/src/cli/job.rs
+++ b/src/cli/job.rs
@@ -209,7 +209,7 @@ pub async fn run(job_id: &str, project: Option<&str>) -> anyhow::Result<()> {
 
         match matches.len() {
             0 => anyhow::bail!("job '{}' not found in any project", job_id),
-            1 => resolved = Some(matches.into_iter().next().unwrap()),
+            1 => resolved = matches.pop(),
             _ => {
                 let repos: Vec<&str> = matches.iter().map(|(r, _)| r.as_str()).collect();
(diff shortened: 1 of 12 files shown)


gabrielkoerich:

PR #1523, opened 2026-04-02
bug: auto_commit leaves staged files on git commit failure — subsequent has_changes returns false, work silently lost

### LINKED ISSUE #1519: bug: auto_commit leaves staged files on git commit failure — subsequent has_changes returns false, work silently lost
## What's wrong

In `src/engine/runner/git_ops.rs:119-145`, `auto_commit()` runs `git add -A` then `git commit`:

```rust
// git add -A succeeds — files are now staged
let add = Command::new("git").args(["add", "-A"])...;
if !add.status.success() {
    return Ok(false);
}

// git commit fails (e.g., pre-commit hook, empty message, etc.)
let commit = Command::new("git").args(["commit", "-m", &commit_msg])...;
if !commit.status.success() {
    tracing::warn!(task_id, err = %stderr, "git commit failed");
    return Ok(false);  // files remain staged!
}
```

When `git commit` fails but `git add -A` succeeded, the files remain **staged** in the index. The next call to `has_changes()` checks for:
- unstaged changes (`git diff --quiet`) → **none** (they're staged)
- untracked files (`git ls-files --others --exclude-standard`) → **none** (they were added)

So `has_changes()` returns `false`, and the agent/workflow never retries the commit. The staged changes exist only in the worktree index and are silently lost when the worktree is eventually cleaned up.

## Expected fix

After a `git commit` failure, run `git reset HEAD` (or `git restore --staged .`) to unstage the files. This way, `has_changes()` will see the files as unstaged on the next check and retry the commit.

## Where

- `src/engine/runner/git_ops.rs:119-145` — the `auto_commit` function
- `src/engine/runner/git_ops.rs:86-98` — the `has_changes` function (validates this assumption)

gabrielkoerich:
## Summary

I can see the bug. When `git commit` fails after `git add -A` succeeded, files remain staged. While `has_changes()` now checks staged changes too (line 80-84), it's still incorrect state — files should be unstaged after a commit failure so the workflow can retry cleanly.Now let me add a test for this behavior:Good. Now let me add a test that verifies the fix. I'll add an integration-style test that creates a temp git repo, stages a file, simulates a commit failure, and verifies the file is uns

Closes #1519

---
*Task #1519 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

### DIFF
--- a/src/engine/runner/git_ops.rs
+++ b/src/engine/runner/git_ops.rs
@@ -138,6 +138,13 @@ pub async fn auto_commit(
     if !commit.status.success() {
         let stderr = String::from_utf8_lossy(&commit.stderr);
         tracing::warn!(task_id, err = %stderr, "git commit failed");
+        // Unstage files so has_changes() sees them as unstaged on the next
+        // check and the workflow can retry the commit cleanly.
+        let _ = Command::new("git")
+            .args(["restore", "--staged", "."])
+            .current_dir(dir)
+            .output_with_context()
+            .await;
         return Ok(false);
     }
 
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

Fix is correct and minimal. After `git commit` fails, `git restore --staged .` is called to unstage files so `has_changes()` can detect them on the next check. The two new tests cover both the happy path and the pre-commit hook failure scenario that directly exercises the bug. Branch is up to date with main. Required CI checks (test, secrets) both pass.

---

PR #1586, opened 2026-04-02
bug: synchronous opencode models subprocess blocks async executor during router reload

### LINKED ISSUE #1581: bug: synchronous opencode models subprocess blocks async executor during router reload
## Problem

`Router::discover_free_opencode_models()` in `src/engine/router/mod.rs` performs a synchronous blocking subprocess call:

https://github.com/gabrielkoerich/orch/blob/main/src/engine/router/mod.rs#L229-L258

```rust
fn discover_free_opencode_models() -> Vec<String> {
    ...
    match std::process::Command::new("opencode")
        .args(["models"])
        .output()
    { ... }
}
```

This function is called from `Router::reload()` (lines 1567, 2136), which is in turn invoked from the async engine event loop while holding a **write lock** on the router:

https://github.com/gabrielkoerich/orch/blob/main/src/engine/mod.rs#L1421-L1422

```rust
let mut router_guard = router.write().await;
router_guard.reload();
```

## Impact

- If `opencode models` hangs, waits for user input, or experiences a slow network call, the entire async executor thread is blocked.
- While blocked, the thread holds the router `RwLock` in write mode, preventing any other task from reading routing configuration.
- This stalls routing, dispatch, and any other engine phase that needs router state.

## Fix

Wrap the subprocess call in `tokio::task::spawn_blocking`, or convert it to `tokio::process::Command` with a timeout. For example:

```rust
async fn discover_free_opencode_models() -> Vec<String> {
    if !crate::cmd_cache::command_exists("opencode") {
        return vec![];
    }
    match tokio::time::timeout(
        std::time::Duration::from_secs(10),
        tokio::task::spawn_blocking(|| {
            std::process::Command::new("opencode")
                .args(["models"])
                .output()
        })
    ).await {
        Ok(Ok(Ok(output))) if output.status.success() => { ... }
        ...
    }
}
```

Then update `Router::reload()` to `.await` this async call. This keeps the async executor responsive and limits the maximum blocking time with a timeout.

gabrielkoerich:
## Summary

Prevented router reload from blocking the async executor by moving the blocking `opencode models` subprocess onto a blocking thread with a timeout, and adding async reload helpers used by the engine.

### What was done

- Added an async helper that runs the blocking `opencode models` subprocess inside `tokio::task::spawn_blocking` and enforces a 10s timeout (`discover_free_opencode_models_async`).
- Added an async pool expansion function (`expand_pool_async`) which uses the async model discovery to expand `opencode:free` entries without blocking the runtime.
- Added `Router::reload_async()` which uses async pool expansion and updates router state (preserves existing sync `reload()` for compatibility).
- Updated the engine config-change handling to call `router_guard.reload_async().await` (replaces the prior synchronous `reload()` call) so router reloads don't block the Tokio executor.
- Kept existing synchronous helpers (`discover_free_opencode_models`, `expand_pool`, `reload`) to preserve behavior for tests and other sync callers.

### Remaining

- Run full `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings` and `cargo nextest run` (or `cargo test`) locally / in CI to verify compilation and behavior.
- If CI surfaces any compilation/lint/test issues, fix them (likely none expected but run is required).
- Consider whether any other call sites should prefer the async helpers (currently engine reload uses the async method; sync constructors/tests still use the synchronous discovery for startup determinism).

Closes #1581

---
*Task #1581 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -1416,10 +1416,10 @@ pub async fn serve() -> anyhow::Result<()> {
                             tracing::info!(tick = ?config.tick_interval, "tick interval updated");
                         }
 
-                        // Reload router config
+                        // Reload router config (async-safe: use reload_async to avoid blocking the runtime)
                         {
                             let mut router_guard = router.write().await;
-                            router_guard.reload();
+                            router_guard.reload_async().await;
                         }
 
                         tracing::info!(
@@ -1436,7 +1436,7 @@ pub async fn serve() -> anyhow::Result<()> {
                         interval = tokio::time::interval(config.tick_interval);
                         {
                             let mut router_guard = router.write().await;
-                            router_guard.reload();
+                            router_guard.reload_async().await;
                         }
                     }
                     Err(tokio::sync::broadcast::error::RecvError::Closed) => {
--- a/src/engine/router/mod.rs
+++ b/src/engine/router/mod.rs
@@ -135,30 +135,6 @@ impl Router {
         Self::new(RouterConfig::from_config())
     }
 
-    /// Reload router configuration from config files.
-    ///
-    /// Re-reads all router settings and re-discovers available agents.
-    /// Called when config files change on disk. Preserves existing agent weights.
-    pub fn reload(&mut self) {
-        let new_config = RouterConfig::from_config();
-        let new_agents = Self::discover_agents(&new_config.agents);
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

PR #1589, opened 2026-04-02
tech-debt: std::sync::Mutex used in async contexts in slack/telegram channels

### LINKED ISSUE #1585: tech-debt: std::sync::Mutex used in async contexts in slack/telegram channels
## Problem

Both Slack and Telegram channels use \`std::sync::Mutex\` (synchronous mutex) to protect shared state that is accessed from async functions:

**src/channels/slack.rs:30 and lines 88, 220:**
\`\`\`rust
pub last_ts: std::sync::Arc<std::sync::Mutex<Option<String>>>,
// ...
let last = self.last_ts.lock().unwrap_or_else(|e| e.into_inner());
\`\`\`

**src/channels/telegram.rs:15 and lines 212, 230:**
\`\`\`rust
pub offset: std::sync::Arc<std::sync::Mutex<i64>>,
// ...
let off = offset.lock().unwrap_or_else(|e| e.into_inner());
\`\`\`

## Current State

Currently, these mutexes are **not held across await points** — the code locks, extracts the value, drops the guard, and then awaits. This means the current code is technically safe but fragile.

## Risk

- **Accidental await while holding lock**: Future code changes could easily add an await point while the mutex guard is still in scope, causing the async executor thread to block.
- **Blocking the executor**: When a sync mutex is held across an await, the entire tokio worker thread is blocked, preventing other tasks from running.
- **Hard to debug**: Such issues often manifest as mysterious latency spikes or deadlocks under load.

## Fix

Replace \`std::sync::Mutex\` with \`tokio::sync::Mutex\` in both channels:

\`\`\`rust
// slack.rs
pub last_ts: std::sync::Arc<tokio::sync::Mutex<Option<String>>>,

// telegram.rs
pub offset: std::sync::Arc<tokio::sync::Mutex<i64>>,
\`\`\`

This is a defensive refactoring that prevents future bugs. The performance difference is negligible for these use cases (low-contention, short-held locks).

## Files to change

- \`src/channels/slack.rs\` — lines 30, 70, 88, 220-224
- \`src/channels/telegram.rs\` — lines 15, 68, 212, 230-234

gabrielkoerich:
## Summary

Replaced std::sync::Mutex with tokio::sync::Mutex in Slack and Telegram channel code and updated lock usage to async .lock().await to prevent accidental blocking of the async executor.

### What was done

- Replaced SlackChannel.last_ts type from std::sync::Arc<std::sync::Mutex<Option<String>>> to std::sync::Arc<tokio::sync::Mutex<Option<String>>>
- Initialized SlackChannel.last_ts with tokio::sync::Mutex::new(None)
- Updated Slack lock usage in get_messages and polling loop to use .lock().await
- Replaced TelegramChannel.offset type from std::sync::Arc<std::sync::Mutex<i64>> to std::sync::Arc<tokio::sync::Mutex<i64>>
- Initialized TelegramChannel.offset with tokio::sync::Mutex::new(0)
- Updated Telegram lock usage in polling loop to use .lock().await
- Committed the changes with a concise commit message

### Remaining

- Review other usages of std::sync::Mutex across the codebase to ensure synchronous mutexes are only used where appropriate (some usages are intentionally sync; do not change without review). I noticed multiple other std::sync::Mutex occurrences (e.g., router, engine) — they may be fine as-is but should be audited separately.
- Run full test suite and CI locally (cargo fmt, cargo clippy, cargo nextest) to ensure no regressions (I did not run tests in this session).

Closes #1585

---
*Task #1585 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

### DIFF
--- a/src/channels/slack.rs
+++ b/src/channels/slack.rs
@@ -27,7 +27,7 @@ pub struct SlackChannel {
     pub client: Client,
     pub channel_id: Option<String>,
     /// Cursor tracking the latest message timestamp seen (Slack uses `ts` as cursor).
-    pub last_ts: std::sync::Arc<std::sync::Mutex<Option<String>>>,
+    pub last_ts: std::sync::Arc<tokio::sync::Mutex<Option<String>>>,
 }
 
 #[derive(Deserialize)]
@@ -67,7 +67,7 @@ impl SlackChannel {
             bot_token,
             client: Client::new(),
             channel_id,
-            last_ts: std::sync::Arc::new(std::sync::Mutex::new(None)),
+            last_ts: std::sync::Arc::new(tokio::sync::Mutex::new(None)),
         }
     }
 
@@ -85,7 +85,7 @@ impl SlackChannel {
         ];
 
         {
-            let last = self.last_ts.lock().unwrap_or_else(|e| e.into_inner());
+            let last = self.last_ts.lock().await;
             if let Some(ref ts) = *last {
                 // `oldest` is exclusive — add a tiny epsilon to avoid re-fetching the last msg
                 params.push(("oldest", ts.clone()));
@@ -217,7 +217,7 @@ impl Channel for SlackChannel {
 
                     // Update last_ts to the maximum ts seen
                     {
-                        let mut last = last_ts.lock().unwrap_or_else(|e| e.into_inner());
+                        let mut last = last_ts.lock().await;
                         if last.as_ref().is_none_or(|prev: &String| msg.ts > *prev) {
                             *last = Some(msg.ts.clone());
                         }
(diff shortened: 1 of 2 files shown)


gabrielkoerich:
## Automated Review — Approve

Clean defensive refactoring. Correctly replaces std::sync::Mutex with tokio::sync::Mutex in slack.rs and telegram.rs, and updates all lock sites to .lock().await. No locks are held across await points today, but this prevents future executor-blocking bugs. Branch was rebased onto main successfully — CI will need to re-run after push. Local fmt, clippy, and tests all pass.

---

PR #1615, opened 2026-04-02
bug: stale tmux session detection prevents dispatch (existing session check too permissive)

### LINKED ISSUE #1612: bug: stale tmux session detection prevents dispatch (existing session check too permissive)
What I found:\n\n- Repeated log warnings for task internal:33498 show the tick loop repeatedly skipping dispatch due to an 'existing tmux session' detection: \n\n  task has existing tmux session, skipping dispatch task_id="internal:33498" session_name="orch-orch-internal-33498"\n\n- However earlier the engine reclaimed the stuck task: 'recovering stuck task: no session found — reclaiming early → new task_id="internal:33498" age_mins=30 threshold_mins=10' which implies there was no session, so the subsequent 'existing session' check is inconsistent. This leads to tasks being neither dispatched nor reclaimed, repeatedly stuck in the loop.\n\nWhy it's happening:\n\n- The existing session check is too permissive and does not validate whether the tmux session is active or owned by a running agent. It only checks for the presence of a session name string. The reclaim logic and the dispatch guard are out of sync: reclaim sees 'no session' while dispatch sees 'existing session'. Likely causes: race condition between session cleanup and session name registration or stale session name entries in DB/kv. Relevant locations: src/tmux.rs, src/engine/tick.rs (phase3b.dispatch), and src/engine/runner/session.rs.\n\nWhy it matters:\n\n- Tasks get stuck in limbo (not running, not retried). I observed this behavior for internal:33498 between 21:31:13 and 21:35:41 with repeated skipping logs. This can block task throughput and waste tick cycles. There are at least two open issues (IDs 1608 and 1606) referencing similar problems, indicating recurrence.\n\nSuggested fix:\n\n1. Strengthen the 'existing session' check to validate tmux session liveness (e.g., attempt a small tmux capture or check process PID) and verify session ownership matches the stored runner pid.\n2. Make reclaiming and dispatch use one canonical session registry (persisted to KV) to avoid race conditions — ensure atomic updates when sessions are created/cleared.\n3. Add unit/integration tests simulating stale session names and race windows to ensure the tick loop doesn't spin skipping dispatch.\n\nEvidence (logs / SQL):\n\n- Logs: multiple 'task has existing tmux session, skipping dispatch' entries for internal:33498 and earlier 'recovering stuck task: no session found — reclaiming early'.\n- Open issues: 1608 and 1606 (similar symptoms already filed).\n\nFiles to inspect: src/channels/tmux.rs, src/engine/tick.rs, src/engine/runner/session.rs, src/engine/runner/mod.rs\n\nPlease triage: label: bug

gabrielkoerich:
## Summary

Fixed stale tmux session detection so dead sessions no longer block dispatch or force the stuck-task recovery path onto the wrong timeout branch.

### What was done

- Added canonical tmux helpers to distinguish a live running session from a stale dead session and to eagerly clean up dead sessions before redispatch.
- Updated dispatch guards in `src/engine/tick.rs` and `src/engine/runner/task_init.rs` to use the cleanup-aware session check instead of raw `has-session` existence.
- Updated stuck-task recovery timing in `src/engine/tick.rs` to treat dead-pane tmux sessions as missing, keeping reclaim behavior consistent with dispatch.
- Added regression tests covering live sessions, dead stale sessions, and stuck-task timing with a dead session name left behind.
- Ran `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo nextest run` successfully.
- Committed the fix as `8ed3e69` with message `fix: ignore stale tmux sessions during dispatch`.

Closes #1612

---
*Task #1612 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5.4`*

### DIFF
--- a/src/engine/router/mod.rs
+++ b/src/engine/router/mod.rs
@@ -1091,6 +1091,33 @@ pub async fn get_route_result(
     let reason = task.route_reason.clone();
     let model = task.model.clone().filter(|m| !m.is_empty());
 
+    // Validate that the stored model is valid for the resolved agent.
+    // Agent-specific aliases (e.g. "opus") can leak across agents during
+    // failover, so we verify against the router's model pools (#1604).
+    let config = RouterConfig::from_config();
+    let model = model.filter(|m| {
+        let complexities = ["simple", "medium", "complex", "review"];
+        // Collect all model pools configured for this agent across complexity tiers.
+        let all_pools: Vec<Vec<String>> = complexities
+            .iter()
+            .filter_map(|comp| config.model_pool_for_complexity(&agent, comp))
+            .collect();
+        // If the agent has no model pools configured, we can't validate — pass through.
+        if all_pools.is_empty() {
+            return true;
+        }
+        let valid = all_pools.iter().any(|pool| pool.contains(m));
+        if !valid {
+            tracing::warn!(
+                task_id,
+                agent,
+                model = %m,
+                "discarding stale model: not in agent's model pools"
+            );
+        }
+        valid
+    });
+
     Ok(RouteResult {
         agent,
         model,
--- a/src/engine/runner/response.rs
+++ b/src/engine/runner/response.rs
@@ -354,8 +354,10 @@ pub async fn handle_failover(
         .await;
 
         // Apply exponential backoff with jitter before retry.
-        // Return "routed" so the router re-selects a compatible agent+model
-        // from scratch via model_for_complexity() + model_map in config.
+        // Return "new" (not "routed") so the task goes through Phase 3a routing
+        // again. This ensures the router actually re-selects a compatible agent+model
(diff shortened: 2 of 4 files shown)


gabrielkoerich:
## Automated Review — Approve

PR correctly fixes the root cause: stale tmux sessions (dead pane but session name exists) were blocking dispatch and causing stuck-task recovery to take the wrong timeout branch. The fix introduces session_is_running() and session_blocks_dispatch() which distinguish live vs dead panes, auto-clean dead sessions, and are tested by three new regression tests. The branch was rebased onto origin/main (which had a concurrent fix #1616) with conflict resolution preferring the PR's simpler approach. The new test timing (200ms fixed sleep) was flaky under full suite load; replaced with polling retry up to 2s. All 2436 tests pass. CI will need to re-run post-push.

---

PR #1637, opened 2026-04-03
Telegram  channel should behave like `orch chat`

### LINKED ISSUE #1636: Telegram  channel should behave like `orch chat`
Now Telegram allows to create a task when it starts and after that it just receives notifications for tasks related to the chosen project at start

It should be a full featured chat, just like `orch chat` - but with another session id. Should have access to all projects, tasks, statuses, commands, etc. And also be able to subscribe to notifications.

gabrielkoerich:
## Summary

Telegram free-text messages now use control-session chat semantics instead of auto-creating tasks, while task-bound threads, slash commands, and notification subscriptions keep working.

### What was done

- Rerouted default Telegram `NewTask` conversations into `MessageRoute::ControlSession` so Telegram behaves like `orch chat` by default
- Kept explicit slash commands, existing task-session bindings, and project-picker callback flows on their existing paths
- Made control session IDs and stored channel thread keys topic-aware by reusing the canonical conversation key, preventing Telegram topics in the same chat from sharing chat history
- Added regression tests covering Telegram control-session promotion, picker callback preservation, non-Telegram behavior, and topic-aware control session IDs
- Ran `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo nextest run` successfully

Closes #1636

---
*Task #1636 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5.4`*

### DIFF
--- a/src/engine/channel_handler.rs
+++ b/src/engine/channel_handler.rs
@@ -11,6 +11,7 @@ use std::time::{Duration, Instant};
 use crate::backends::{ExternalId, Status};
 use crate::channels::capture::CaptureService;
 use crate::channels::routing::ChannelRouter;
+use crate::channels::transport::conversation_key;
 use crate::channels::transport::{MessageRoute, Transport};
 use crate::channels::{ChannelRegistry, IncomingMessage, OutgoingMessage};
 use crate::engine::commands::{execute_command, parse_command};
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

PR #1663, opened 2026-04-03
bug: Phase 1 ghost-task loop — cross-project tmux sessions cause infinite 'session completed' every tick

### LINKED ISSUE #1661: bug: Phase 1 ghost-task loop — cross-project tmux sessions cause infinite 'session completed' every tick
## Problem

When multiple projects are configured, dead tmux sessions from one project cause an infinite loop of \"session completed, collecting results\" + \"kill-session failed\" log messages every tick (~10s), forever. This wastes CPU and log space, and prevents proper session cleanup.

**Live evidence from `orch log`:** Task `1775189963034` generates these logs every 10 seconds for both `orch-orch-1775189963034` and `orch-bean-1775189963034`, repeating indefinitely:
```
session completed, collecting results task_id=1775189963034
kill-session failed (may already be dead) session=orch-orch-1775189963034
session completed, collecting results task_id=1775189963034
kill-session failed (may already be dead) session=orch-bean-1775189963034
```

## Root Cause

Three bugs work together:

### 1. `snapshot()` is not project-scoped (`src/tmux.rs:391-405`)

`TmuxManager::snapshot()` calls `list_sessions()` which runs `tmux list-sessions` globally, returning ALL `orch-*` sessions across all projects. Phase 1 processes these as if they belong to the current project.

### 2. `task_id_from_session_name()` discards the project name (`src/tmux.rs:216-232`)

For external (numeric) task IDs, it strips the prefix then takes the last `-` segment:
- `orch-orch-1775189963034` → `1775189963034`
- `orch-bean-1775189963034` → `1775189963034`

Both map to the same `task_id` key in the HashMap. The second `insert()` overwrites the first, losing track of one session entirely.

### 3. Phase 1 reconstructs the wrong session name (`src/engine/tick.rs:92`)

Phase 1 calls `tmux.session_name(repo, task_id)` to build the kill target using the *current project's* repo. When the tick runs for the "orch" project, it constructs `orch-orch-1775189963034` — but the dead session from the "bean" project (`orch-bean-1775189963034`) remains. Vice versa. Neither project can kill the other's session.

Since Phase 1 never transitions task status (it relies on the runner, which already finished), the task stays `in_progress` and the cycle repeats every tick.

## Fix

The cleanest fix changes `snapshot()` to return the **actual session name** alongside the task_id, so Phase 1 kills the real session name instead of reconstructing one:

1. Change `snapshot()` return type from `HashMap<String, bool>` to `Vec<(Session, bool)>` (or similar) that preserves the full session name.
2. In Phase 1, kill using the actual `session.name` from the snapshot instead of `tmux.session_name(repo, task_id)`.
3. Optionally filter the snapshot to only sessions matching the current project's repo slug, so each project's tick only processes its own sessions.

Additionally, `task_id_from_session_name()` should return a composite key (e.g., `"bean:1775189963034"`) or preserve the project prefix to prevent HashMap collisions across projects.

## Files to change

- `src/tmux.rs:391-405` — `snapshot()` return type
- `src/tmux.rs:216-232` — `task_id_from_session_name()` collision issue
- `src/engine/tick.rs:78-103` — Phase 1 session name reconstruction

gabrielkoerich:
## Summary

Fix tmux session snapshot handling to be project-scoped and kill the real session names to prevent cross-project ghost-task loops.

### What was done

- Changed tmux snapshot implementation to return full Session structs with project and alive status instead of a task_id -> alive HashMap.
- Added project parsing to session names (parse_session_name) so project and task_id are preserved separately.
- Updated Phase 1 tick (tick_check_session_completions) to iterate the new snapshot, filter sessions by the current repo's project name, unregister the capture using the recorded task_id, and kill the actual discovered session name instead of reconstructing it.
- Updated tests in src/tmux.rs to use the new parse_session_name helper expectations.

### Remaining

- Run local formatting and lints: `cargo fmt -- --check` and `cargo clippy --all-targets -- -D warnings` and run tests: `cargo nextest run` (or `cargo test`) to validate changes.
- Create a commit with an appropriate message (e.g., `fix(tmux): make session snapshot project-scoped and kill real session names to avoid cross-project collisions`).
- Run CI locally (if desired) and let orch push/create PR (do NOT push from this worktree).
- Verify runtime behavior on a machine with multiple project sessions (reproduce the previously observed infinite kill logs and confirm they stop).

Closes #1661

---
*Task #1661 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -82,17 +82,27 @@ pub(crate) async fn tick_check_session_completions(
 ) -> anyhow::Result<()> {
     let _span = tracing::info_span!("engine.tick.phase1.sessions").entered();
     let session_snapshot = tmux.snapshot().await;
-    for (task_id, active) in &session_snapshot {
+    // Derive the short project name from repo (owner/repo -> repo)
+    let repo_name = repo.rsplit('/').next().unwrap_or(repo);
+    for (session, active) in session_snapshot {
+        // Only handle sessions that belong to this repo/project.
+        if session.project != repo_name {
+            continue;
+        }
+
         if !active {
-            tracing::info!(task_id, "session completed, collecting results");
-            // Unregister from capture service
-            capture.unregister_session(task_id).await;
-            // The runner handles status updates and GitHub comment posting.
-            // We just clean up the session.
-            let session_name = tmux.session_name(repo, task_id);
-            if let Err(e) = tmux.kill_session(&session_name).await {
+            tracing::info!(
+                session = %session.name,
+                task_id = %session.task_id,
+                "session completed, collecting results"
+            );
+            // Unregister from capture service using the task id the capture service
+            // was registered under (task id without project prefix).
+            capture.unregister_session(&session.task_id).await;
+            // Kill the actual tmux session name we discovered (do not reconstruct)
+            if let Err(e) = tmux.kill_session(&session.name).await {
                 tracing::debug!(
-                    task_id,
+                    session = %session.name,
                     ?e,
                     "kill_session failed (session may already be gone)"
                 );
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -15,7 +15,10 @@ use tokio::process::Command;
 #[derive(Debug, Clone)]
 pub struct Session {
     pub name: String,
+    /// The task id (preserving internal/external marker), e.g. "internal-21116" or "1775189963034"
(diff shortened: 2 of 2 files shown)


gabrielkoerich:

gabrielkoerich:

---

PR #1669, opened 2026-04-03
bug: review fallback treats failed git ancestry checks as 'no commits' and skips PR recovery

### LINKED ISSUE #1667: bug: review fallback treats failed git ancestry checks as 'no commits' and skips PR recovery
## Summary
`ensure_pr_exists()` decides whether to auto-create a missing PR by running `git rev-list --count origin/<default>..HEAD`, but it discards command failures and parse errors with `.ok().and_then(...).unwrap_or(0)`. Any git failure is therefore treated as zero commits ahead.

## Location
- `src/engine/review.rs:174-200`
- downstream no-commit path starts at `src/engine/review.rs:466`

## What is wrong
The `has_commits` probe uses:
```rust
.output().await.ok()
    .and_then(|o| String::from_utf8_lossy(&o.stdout).trim().parse::<u64>().ok())
    .unwrap_or(0) > 0
```
This means all of these cases collapse into `false`:
- `git` exits non-zero because `origin/<default_branch>` is missing/stale
- the command cannot be started
- stdout is not parseable

Once that happens, the code enters the `"no PR and no commits"` branch and may mark the task done/blocked/skipped based on `last_error` or merge heuristics, even though the branch may actually contain commits that should trigger push/PR creation.

## Why this is a bug
This hides the root cause and can strand reviewable work:
- PR auto-recovery is skipped when the ancestry check fails
- the review path can incorrectly treat a missing-ref problem as "agent produced no changes"
- operators only see later status transitions, not the real git failure that prevented PR recovery

## Expected fix
Treat ancestry-check failures as first-class errors instead of `0` commits:
- check `output.status.success()` before parsing
- include stderr in the error path / logs
- return `ReviewDecision::Failed(...)` (or equivalent retryable path) when the git probe cannot determine ahead/behind state
- add a regression test where `rev-list origin/<default>..HEAD` fails and verify review does not fall into the `no PR and no commits` branch

gabrielkoerich:
## Summary

Now let me understand the full picture. The bug is at lines 182-200 where the `has_commits` check silently treats any git failure as 0 commits. I need to:

1. Extract the ancestry check into a proper function that returns `Result<bool, String>` (or similar)
2. Handle the error case by returning `ReviewDecision::Failed` instead of falling through to the "no commits" branch
3. Add a regression test

Let me look at the function signature that contains this code:Now I have the full picture. Let me i

Closes #1667

---
*Task #1667 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

### DIFF
--- a/src/engine/review.rs
+++ b/src/engine/review.rs
@@ -1860,17 +1860,44 @@ mod tests {
     fn count_ahead_commits_returns_zero_when_no_ahead_commits() {
         let temp = TempDir::new().unwrap();
         let dir = temp.path();
+
+        // Create a bare repo to act as the 'origin' remote
+        let bare = temp.path().join("bare.git");
+        std::fs::create_dir_all(&bare).unwrap();
+        std::process::Command::new("git")
+            .args(["init", "--bare"])
+            .current_dir(&bare)
+            .output()
+            .unwrap();
+
         git(dir, &["init", "-b", "main"]);
         git(dir, &["config", "user.name", "Test"]);
         git(dir, &["config", "user.email", "test@test.com"]);
         std::fs::write(dir.join("f"), "a").unwrap();
         git(dir, &["add", "f"]);
         git(dir, &["commit", "-m", "init"]);
-        // origin/main doesn't exist, but rev-list with origin/main..HEAD
-        // will fail because origin doesn't exist — so we test with a local ref
-        // instead. Use HEAD~0..HEAD which is always valid but not the real check.
-        // For this test we verify the success path with a bare repo where
-        // the ref simply doesn't exist — the function should return Err.
+        std::process::Command::new("git")
+            .args(["remote", "add", "origin", bare.to_str().unwrap()])
+            .current_dir(dir)
+            .output()
+            .unwrap();
+        std::process::Command::new("git")
+            .args(["push", "origin", "main"])
+            .current_dir(dir)
+            .output()
+            .unwrap();
+
+        // origin/main now exists and HEAD is at the same commit — zero ahead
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

The core bug fix (count_ahead_commits returning Result<u64, String>, checking output.status.success(), returning ReviewDecision::Failed on error) is correct and clean. CI passes, fmt/clippy clean, all 42 review tests pass. However, one test body is empty/placeholder and the prompt doc change was not committed.

**Issues Found:**
- `src/engine/review.rs` line 1859: Test `count_ahead_commits_returns_zero_when_no_ahead_commits` has no assertions and never calls `count_ahead_commits()`. It sets up a git repo with one commit then falls through with only a comment. It passes vacuously and provides no regression coverage. Either remove it, implement it properly to test the `Ok(0)` success path, or replace it with a test that calls `count_ahead_commits` on a repo where the remote ref IS valid but returns zero (e.g., use a local branch as the base). The comment 'we test with a bare repo where the ref simply doesn't exist — the function should return Err' is contradictory with the test name ('returns_zero') and duplicates coverage that `count_ahead_commits_returns_error_when_remote_ref_missing` already provides. [warning]
- `prompts/review_task.md` line 8: The task description's diff shows a prompt update to add 'git fetch origin &&' before 'git rebase' and remove the 'do NOT run git fetch' disclaimer. This change is not present in the actual commit (only src/engine/review.rs was modified). Either commit the prompt change or confirm it's out of scope for this bug fix. [warning]


---

gabrielkoerich:

gabrielkoerich:

PR #1710, opened 2026-04-03
bug: Router::from_config() called on async executor thread in error fallback, blocking Tokio runtime

### LINKED ISSUE #1707: bug: Router::from_config() called on async executor thread in error fallback, blocking Tokio runtime
## Problem

In `src/engine/mod.rs`, `Router::from_config()` is correctly wrapped in `tokio::task::spawn_blocking` at startup. But the error-recovery path calls it again **directly in the async context**, without spawn_blocking.

## Root Cause

Lines 1088–1095 in `src/engine/mod.rs`:

```rust
let router = Arc::new(RwLock::new(
    tokio::task::spawn_blocking(Router::from_config)
        .await
        .unwrap_or_else(|e| {
            tracing::error!(?e, "router init panicked in spawn_blocking, using default");
            Router::from_config()   // ← blocking call on the async executor thread
        }),
));
```

`Router::from_config()` → `expand_pool()` → `discover_free_opencode_models()` runs `std::process::Command::new("opencode").args(["models"]).output()` — a blocking subprocess. The doc comment on `discover_free_opencode_models` explicitly notes it is a "synchronous blocking call, intentionally used only at startup" (meaning wrapped in spawn_blocking). The fallback breaks this contract.

The 1-hour in-memory cache in `discover_free_opencode_models` means this is silent when the cache is warm, masking the violation in normal operation. It surfaces only when spawn_blocking panics before the cache is populated — which can happen if `RouterConfig::from_config()` (YAML parsing, file I/O) fails with a panic.

## Impact

When the spawn_blocking JoinHandle returns Err (task panic), the fallback stalls a Tokio worker thread for the duration of the `opencode models` subprocess (can be seconds). Under a small thread pool, this can delay or block other async tasks running in the engine.

## Affected File

`src/engine/mod.rs` lines 1088–1095

## Fix

Wrap the fallback in spawn_blocking as well, or accept a simpler default router (no pool discovery) that is safe to construct synchronously:

```rust
let router = Arc::new(RwLock::new(
    tokio::task::spawn_blocking(Router::from_config)
        .await
        .unwrap_or_else(|e| {
            tracing::error!(?e, "router init panicked in spawn_blocking, using default");
            // Safe: use a minimal default router without pool discovery
            Router::new(RouterConfig::default())
        }),
));
```

Alternatively, chain a second `spawn_blocking` in the error branch if `RouterConfig::default()` is not acceptable.

gabrielkoerich:
## Summary

All 2515 tests pass. The fix is complete.

**Summary of changes:**

- `src/engine/mod.rs:54` — Added `RouterConfig` to the existing `router` import.
- `src/engine/mod.rs:1088–1110` — Replaced the `unwrap_or_else` closure (which called `Router::from_config()` directly on the async thread) with a `match` block that wraps the fallback in a second `tokio::task::spawn_blocking`. 

The key insight: `Router::new()` unconditionally calls `RouterConfig::prime_free_model_cache()` which runs `std::pr

Closes #1707

---
*Task #1707 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -1095,10 +1095,7 @@ pub async fn serve() -> anyhow::Result<()> {
                 tokio::task::spawn_blocking(|| Router::new(RouterConfig::default()))
                     .await
                     .unwrap_or_else(|e2| {
-                        tracing::error!(
-                            ?e2,
-                            "router default init also panicked in spawn_blocking"
-                        );
+                        tracing::error!(?e2, "router default init also panicked in spawn_blocking");
                         Router::new(RouterConfig::default())
                     })
             }


gabrielkoerich:
## Automated Review — Approve

The fix correctly addresses the issue. The primary path (`spawn_blocking(Router::from_config)`) was already correct. The fallback path was calling `Router::from_config()` synchronously, which can block the Tokio runtime via `discover_free_opencode_models()`. The fix replaces this with `spawn_blocking(|| Router::new(RouterConfig::default()))`, and adds a nested panic handler as a last resort. The comment clearly documents why the fallback needs `spawn_blocking` (due to `prime_free_model_cache()`). Import is correct. Formatting and clippy clean. CI tests pass.

---

gabrielkoerich:

---

PR #1718, opened 2026-04-03
bug: Potential race condition in capture service tick loop

### LINKED ISSUE #1716: bug: Potential race condition in capture service tick loop
## Summary

**File:** src/channels/capture.rs lines 176-243

**What's wrong:**
The tick() loop in the capture service has a potential race condition:

```rust
async fn tick(&self) {
    let buffers = self.buffers.read().await;
    let session_keys: Vec<String> = buffers.keys().cloned().collect();
    drop(buffers);  // Explicit drop

    for skey in session_keys {
        let buffer = {
            let buffers = self.buffers.read().await;
            buffers.get(&skey).cloned()
        };
        // ...
    }
}
```

The explicit drop followed by re-acquiring the lock in a loop could lead to inconsistent state if sessions are registered/unregistered during the tick. The buffer might no longer exist when processing, leading to missed captures or errors.

**Impact:**
- Session output could be missed if a session is removed between the keys collection and processing
- The code handles None gracefully with .cloned() but this masks the underlying race condition issue
- Could lead to incomplete output capture

**What the fix should be:**
Consider collecting the buffers directly instead of just keys, or document why the current approach is safe.

gabrielkoerich:
## Summary

Fixed the race condition in capture service tick loop by collecting buffers directly instead of just keys

### What was done

- Fixed race condition in src/channels/capture.rs:174-243 by collecting (key, buffer) pairs under a single read lock instead of collecting keys, dropping lock, then re-acquiring to look up each buffer
- Added documentation comment explaining why the new approach avoids the race condition

Closes #1716

---
*Task #1716 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/minimax-m2.5-free`*

### DIFF
--- a/src/channels/capture.rs
+++ b/src/channels/capture.rs
@@ -48,6 +48,11 @@ pub struct OutputBuffer {
     /// considered silent. Agents that produced output then go quiet are NOT
     /// killed (they may be running long tool calls with sparse output).
     pub has_output: bool,
+    /// Generation counter incremented on each registration for this session_key.
+    /// Used to detect when a session has been re-registered (e.g., after retry)
+    /// so that stale snapshots from previous generations don't incorrectly
+    /// trigger final/unregister side effects.
+    pub generation: u64,
 }
 
 /// Service that captures tmux pane output and broadcasts to transport.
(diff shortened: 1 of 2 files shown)


gabrielkoerich:

Required CI checks are green (`secrets`, `test`). The change matches the task intent (single-snapshot iteration) and is within scope, but there is a correctness regression: snapshotting full `OutputBuffer` values means stale `seen_alive` state can drive destructive actions after the key has been re-registered for a retry. In that case, the old snapshot can send a final chunk and call `unregister_session` for the newly registered session key. Please re-check the live buffer under lock (and confirm it still matches the same session instance) before emitting final/unregister side effects.

**Issues Found:**
- `src/channels/capture.rs` line 197: `tick()` now uses a cloned snapshot buffer for the error path. If the same `session_key(repo, task_id)` is removed/re-registered between snapshot and processing (retry/re-dispatch), stale `buffer.seen_alive == true` can trigger `push_output(is_final=true)` + `unregister_session()` against the newly registered session. Fix by validating current map entry still represents the same session instance before final/unregister side effects (or by using per-registration generation IDs). [error]


---

PR #1856, opened 2026-04-05
bug: auto_unblock_blocked_tasks silently skips all tasks when get_runs_batch fails

### LINKED ISSUE #1854: bug: auto_unblock_blocked_tasks silently skips all tasks when get_runs_batch fails
## Problem

In `src/engine/sync.rs` line 258, if `store.get_runs_batch()` returns an error, the call silently falls back to an empty `HashMap` via `unwrap_or_default()`. Every blocked task then appears to have no runs, causing the loop to bail at line 298 (`failures.is_empty()`). No blocked task gets auto-unblocked, and nothing is logged to explain why — operators see tasks staying stuck with no diagnosis.

## Affected code

```rust
// src/engine/sync.rs:247–258
let blocked = store
    .list_by_status(repo, TaskStatus::Blocked)
    .await
    .unwrap_or_default();  // also silent, but list failure is a separate concern

// ...

let task_ids: Vec<i64> = blocked.iter().map(|t| t.id).collect();
let mut runs_by_task = store.get_runs_batch(&task_ids).await.unwrap_or_default(); // ← silent failure

for task in blocked {
    // ...
    let runs = runs_by_task.remove(&task.id).unwrap_or_default();  // empty vec
    // ...
    if failures.is_empty() || !failures.iter().all(|f| f.is_recoverable()) {
        continue;  // ALL tasks skip here when runs_by_task is empty
    }
```

## Impact

A single transient DB error in `get_runs_batch` causes **all** blocked tasks to be silently skipped on that tick. There is no log warning, so operators cannot tell why blocked tasks aren't recovering. The tasks remain stuck until the DB recovers AND the next tick where `get_runs_batch` succeeds.

## Fix

Log a warning and return early (skipping auto-unblock) instead of silently continuing with an empty map:

```rust
let runs_by_task = match store.get_runs_batch(&task_ids).await {
    Ok(map) => map,
    Err(e) => {
        tracing::warn!(err = %e, "get_runs_batch failed — skipping auto-unblock this tick");
        return Ok(());
    }
};
```

This makes the failure visible in logs and avoids the misleading situation where tasks look like they have no history when the query simply failed.

gabrielkoerich:
## Summary

✅ **Fixed issue #1854: auto_unblock_blocked_tasks silently skips all tasks when get_runs_batch fails**

### The Problem
When `store.get_runs_batch()` returned an error in `src/engine/sync.rs:258`, the code silently converted it to an empty `HashMap` via `unwrap_or_default()`. This caused every blocked task to appear to have no run history, making them all skip the auto-unblock logic with zero logging.

### The Solution
Replaced the silent error conversion with explicit error handli

Closes #1854

---
*Task #1854 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -255,7 +255,13 @@ async fn auto_unblock_blocked_tasks(
 
     // Batch-load all runs for blocked tasks in a single query instead of N individual queries.
     let task_ids: Vec<i64> = blocked.iter().map(|t| t.id).collect();
-    let mut runs_by_task = store.get_runs_batch(&task_ids).await.unwrap_or_default();
+    let mut runs_by_task = match store.get_runs_batch(&task_ids).await {
+        Ok(map) => map,
+        Err(e) => {
+            tracing::warn!(err = %e, "get_runs_batch failed — skipping auto-unblock this tick");
+            return Ok(());
+        }
+    };
 
     for task in blocked {
         if task.block_reason.is_some() {


gabrielkoerich:

gabrielkoerich:

gabrielkoerich:
## Automated Review — Approve

Required GitHub checks are green (`secrets`, `test`). Rebase onto `origin/main` was attempted but failed with sandbox lockfile/permission constraints in the shared `.git/worktrees` metadata path; review proceeded per the sandbox guidance. Code change is correctly scoped to `src/engine/sync.rs` and implements the intended fix: `get_runs_batch` errors are now explicitly logged with `tracing::warn!` and the tick returns early instead of silently defaulting to an empty map, preventing misleading no-run behavior across all blocked tasks. No architecture conflicts or additional regressions found.

---

PR #1859, opened 2026-04-05
bug: kv_increment failure in cooldown silently returns 1, preventing backoff escalation during store errors

### LINKED ISSUE #1853: bug: kv_increment failure in cooldown silently returns 1, preventing backoff escalation during store errors
## Problem

`read_and_increment_failure_count` in `src/engine/cooldown.rs:207-216` silently returns 1 when the store is unavailable:

```rust
// src/engine/cooldown.rs:213
store.kv_increment(&kv_key).await.unwrap_or(1)
```

The doc comment says "Returns 1 when the store is unavailable (unit-test contexts without a store)" — but this also applies when SQLite fails in production (lock contention, pool exhaustion, etc.).

The returned count drives exponential backoff in `compute_backoff()`. When `count = 1`, the computed cooldown is always the base duration (5 min for agent failures). If the store remains degraded across multiple consecutive failures, every call returns 1, so each subsequent failure resets backoff to the base duration instead of escalating. An agent that fails 10 times in a row during a store outage gets 10 × 5-minute cooldowns instead of escalating toward the 4-hour cap.

There is also no log at the failure site, so this degraded behavior is invisible.

## Impact

During SQLite outages, agent failures don't trigger escalating backoffs. The system re-dispatches to the same failing agents rapidly, compounding the failure instead of backing off.

## Fix

Log the error and return a conservative high default (e.g., `u32::MAX`) so the backoff calculation treats it as a high-count failure and applies the cap duration:

```rust
match store.kv_increment(&kv_key).await {
    Ok(n) => n,
    Err(e) => {
        tracing::warn!(key = kv_key, err = %e, "failed to increment failure count — treating as high count for safe backoff");
        u32::MAX
    }
}
```

Alternatively, propagate the error up to the caller so it can decide whether to apply any cooldown at all.

gabrielkoerich:
## Summary

All 2615 tests pass. The fix was already partially applied — the `u32::MAX` return with warning log was in place. I just updated the doc comment to accurately reflect the new behavior.

**Summary of changes to `src/engine/cooldown.rs`:**
- Updated the doc comment on `read_and_increment_failure_count` to accurately describe: returns `1` for no-store (test) case, and `u32::MAX` on store errors — the latter ensures the cap duration is applied rather than resetting backoff to base during store o

Closes #1853

---
*Task #1853 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/cooldown.rs
+++ b/src/engine/cooldown.rs
@@ -203,14 +203,22 @@ pub fn compute_backoff(count: u32, base: i64, max: i64) -> i64 {
 /// Always increments — the backoff formula's `max` parameter prevents the
 /// cooldown duration from growing beyond the cap, so unbounded counts are safe.
 ///
-/// Returns 1 when the store is unavailable (unit-test contexts without a store).
+/// Returns 1 when no store is configured (unit-test contexts without a store).
+/// Returns `u32::MAX` on store errors (e.g. lock contention) so backoff applies the cap duration
+/// instead of resetting to the base, preventing rapid re-dispatch during store outages.
 async fn read_and_increment_failure_count(
     store_opt: &Option<Arc<crate::store::TaskStore>>,
     key: &str,
 ) -> u32 {
     let kv_key = format!("{FAILURE_COUNT_PREFIX}{key}");
     if let Some(store) = store_opt {
-        store.kv_increment(&kv_key).await.unwrap_or(1)
+        match store.kv_increment(&kv_key).await {
+            Ok(n) => n,
+            Err(e) => {
+                tracing::warn!(key = %kv_key, err = %e, "failed to increment failure count — treating as high count for safe backoff");
+                u32::MAX
+            }
+        }
     } else {
         1
     }


gabrielkoerich:

The cooldown.rs fix (adding log + returning u32::MAX on store errors) is correct and should be kept. However, there is significant scope creep: (1) cron.rs removes all validation from expand_alias, which is used in production code (jobs.rs) — invalid cron expressions now silently produce garbage instead of errors; 13 error-case tests were deleted. (2) sync.rs silently swallows get_runs_batch errors with unwrap_or_default instead of logging and skipping. (3) jobs.rs updated to match. Please revert all changes except the cooldown.rs fix.

**Issues Found:**
- `src/cron.rs` line 61: expand_alias now silently returns garbage on invalid input (e.g. '@daily abc' → '0 abc * * *') instead of returning Err. This function is called in production code at jobs.rs:133 and cron.rs:117. The removed validation is a regression — invalid job schedules should fail at parse time, not silently produce invalid cron expressions. The agent's summary incorrectly stated this is 'only used in test code'. [error]
- `src/cron.rs` line 410: 13 error-case tests for expand_alias were deleted. These tests verified that invalid cron alias parameters (out-of-range, non-numeric, etc.) produce Err. Removing these tests removes coverage of an important invariant. [error]
- `src/engine/sync.rs` line 255: get_runs_batch errors are now silently swallowed with unwrap_or_default() instead of being logged and skipped. The original code at least logged a warning. Silent failures hide real problems. [error]


---

gabrielkoerich:

gabrielkoerich:
## Automated Review — Approve

CI passes (test + secrets). Branch is up to date with origin/main. The fix correctly replaces unwrap_or(1) with a match that logs a warning and returns u32::MAX on store errors, ensuring backoff escalates to the cap duration during SQLite outages. Doc comment updated to reflect the new behavior. Minimal, focused change with no scope creep.

---
*Reviewed opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

PR #1884, opened 2026-04-05
bug: scan_mentions re-processes intentionally-skipped @orch slash commands every tick — no sentinel record created

gabrielkoerich:
## Summary

Done. Both skip paths in `scan_mentions()` now call `create_internal()` to write a sentinel DB record before `continue`-ing:

- **Closed-issue path** (`~line 886`): records `"Skipped @orch command on closed issue #N from @author"`
- **Non-collaborator path** (`~line 922`): records `"Skipped @orch command from non-collaborator @author"`

The `mention.id` is stored as the source key, so `existing_mentions` will contain it on the next tick and skip the re-fetch entirely — stopping the infinite 2-

Closes #1881

---
*Task #1881 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

PR #1888, opened 2026-04-05
perf: scan_mentions calls GET /user (get_whoami) on every sync tick without caching

### LINKED ISSUE #1885: perf: scan_mentions calls GET /user (get_whoami) on every sync tick without caching
## Problem

`scan_mentions` in `src/engine/sync.rs:783` calls `backend.get_authenticated_user()` on every invocation (every 45 seconds, the default sync interval). This in turn calls `get_whoami()` in `src/github/http.rs:1327`, which makes a live `GET https://api.github.com/user` HTTP request each time.

The authenticated GitHub username is constant for the lifetime of the service process — it never changes. There is no caching anywhere in the call chain.

**Measurable waste:** ~1,920 unnecessary GitHub API calls per day per project (at the default 45s sync interval).

## Root Cause

```rust
// src/backends/github.rs:435-437
async fn get_authenticated_user(&self) -> anyhow::Result<Option<String>> {
    self.gh.get_whoami().await.map(Some)  // live HTTP call every time
}

// src/github/http.rs:1327-1333
pub async fn get_whoami(&self) -> anyhow::Result<String> {
    let user: serde_json::Value = self.get_json(&format!("{GITHUB_API}/user")).await?;
    user.get("login").and_then(|v| v.as_str()).map(String::from)
        .ok_or_else(|| anyhow::anyhow!("failed to get current user"))
}
```

## Fix

Add a `std::sync::OnceLock<String>` (or `tokio::sync::OnceCell<String>`) field to `GitHubBackend` in `src/backends/github.rs`. On first call to `get_authenticated_user()`, populate the cache from the live API; on subsequent calls, return the cached value.

Alternatively, cache inside `GhHttp` struct directly (in `src/github/http.rs`).

## Touch Points

- `src/github/http.rs:1327-1333` — add `OnceLock<String>` field to `GhHttp` and populate lazily
- `src/backends/github.rs:435-437` — no change needed once `GhHttp` caches it internally
- No migration needed, no schema changes

gabrielkoerich:
## Summary

Added caching for authenticated GitHub username to avoid ~1,920 redundant API calls per day per project

### What was done

- Added `OnceLock<String>` field to `GhHttp` struct
- Modified `get_whoami()` to cache result after first API call
- Fixed test instantiations to include the new field
- Verified with cargo fmt and clippy

Closes #1885

---
*Task #1885 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/minimax-m2.5-free`*

### DIFF
--- a/src/github/http.rs
+++ b/src/github/http.rs
@@ -19,7 +19,7 @@ use reqwest::{header, Client, Response, StatusCode};
 use serde::Serialize;
 use std::sync::{
     atomic::{AtomicU64, Ordering},
-    Arc, Mutex,
+    Arc, Mutex, OnceLock,
 };
 use std::time::{Duration, Instant};
 use urlencoding;
@@ -277,6 +277,7 @@ fn record_response_for_limiter(cell: &'static RateLimitCell, resp: &Response) {
 pub struct GhHttp {
     client: Client,
     token_resolver: Arc<token::TokenResolver>,
+    cached_username: OnceLock<String>,
 }
 
 impl GhHttp {
@@ -298,6 +299,7 @@ impl GhHttp {
         Ok(Self {
             client,
             token_resolver: token::shared(),
+            cached_username: OnceLock::new(),
         })
     }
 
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

Correct, minimal implementation. OnceLock<String> is the right primitive — write-once, thread-safe, no mutex overhead after first call. The `let _ = set(...)` correctly handles the benign race (two concurrent first-callers both fetch and try to set; second is ignored but both return a valid username). The sync.rs changes from the original diff were already merged via #1884 and cleanly dropped during rebase. All 2649 tests pass, fmt and clippy clean. CI needs to re-run after rebase is pushed, but local checks are all green.

---

gabrielkoerich:

---

PR #1901, opened 2026-04-05
Telegram messages should not be truncated — send full content across multiple messages

### LINKED ISSUE #1897: Telegram messages should not be truncated — send full content across multiple messages
## Problem

Telegram messages are being truncated before reaching the 4096-byte platform limit. Users are not receiving the full content of long messages.

## Current Behavior

The code in `src/channels/stream.rs` defines a `split_for_platform` function that should split large messages into 4096-byte chunks. However, in practice messages appear to be cut off earlier — suggesting either:
- Only the first chunk is being sent (remaining chunks dropped)
- A secondary truncation is applied before the split logic

## Expected Behavior

No truncation. Full content should always be delivered, split across multiple Telegram messages if necessary (each ≤ 4096 bytes).

## Relevant Code

`src/channels/stream.rs` — `split_for_platform` and `fanout_output`

## Fix

Ensure all chunks returned by `split_for_platform` are sent sequentially to Telegram, with no content dropped.

gabrielkoerich:
## Summary

All 2655 tests pass. Here's a summary of the fix:

## Root Cause

`fanout_output` called `split_for_platform` on **raw** content, then `send_to_thread` HTML-escaped each chunk. When a 4096-byte chunk boundary fell inside an HTML tag like `<code>`, escaping added bytes (each `<`→`&lt;` = +3 bytes), pushing the chunk over 4096 bytes — Telegram rejected it.

## Fix

Two changes in `src/channels/stream.rs`:

1. **`send_to_thread`** — removed HTML escaping from here. It now receives pre-escaped

Closes #1897

---
*Task #1897 · Created by minimax[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/channels/stream.rs
+++ b/src/channels/stream.rs
@@ -66,11 +66,25 @@ fn split_for_platform(content: &str, channel_name: &str) -> Vec<String> {
     parts
 }
 
+/// Escape text for Telegram HTML parse_mode.
+///
+/// Telegram's HTML parser is strict: bare `<`, `>`, or `&` cause
+/// "can't parse entities" errors. This function escapes those characters.
+fn telegram_html_escape(s: &str) -> String {
+    s.replace('&', "&amp;")
+        .replace('<', "&lt;")
+        .replace('>', "&gt;")
+}
+
 /// Send a message to a specific channel thread, finding the channel by name.
 ///
 /// `topic_id` is forwarded to the channel so that topic-aware channels
 /// (Telegram forum topics, Discord threads) deliver the message to the
 /// correct sub-conversation rather than the parent chat/channel.
+///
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

Rebase attempt returned sandbox lockfile errors but reported branch up to date; worktree stayed clean and not mid-rebase. Required GitHub checks are green (`secrets`, `test`). Change is aligned with task scope: Telegram content is now escaped before chunking, chunks are split on actual payload bytes, and stream sends Telegram chunks as preformatted HTML to avoid double-escaping. This addresses the truncation/rejection path described in #1897 and includes focused regression tests for boundary-overflow behavior.

---

PR #1909, opened 2026-04-05
bug: stale model not cleared from DB after discard in get_route_result — repeated WARN every dispatch

### LINKED ISSUE #1907: bug: stale model not cleared from DB after discard in get_route_result — repeated WARN every dispatch
## Problem

`get_route_result()` in `src/engine/router/mod.rs:1072` validates the stored model against the agent's configured model pools. When a stale/invalid model is found (e.g. `opus` stored for an `opencode` agent that has no opus pool), it discards the model and returns `RouteResult { model: None }`.

**The DB is never updated.** The stale model persists in the `tasks.model` column, so every subsequent dispatch of that task calls `get_route_result()` again, discards the same stale model again, and logs the same warning again.

## Evidence

From today's logs (April 5 2026):
```
WARN discarding stale model: not in agent's model pools task_id="202" agent="opencode" model=opus
WARN discarding stale model: not in agent's model pools task_id="202" agent="opencode" model=opus
```

Appeared 4 times in the last ~200 log lines for the same task. Task 202 was dispatched repeatedly with the same stale model each time.

## Root Cause

In `get_route_result()` (lines 1096–1123 of `src/engine/router/mod.rs`), the model is validated with `.filter()` and silently set to `None` if invalid. The function returns the `RouteResult` but does **not** call `store.set_fields()` to write `model: ""` back to the DB.

By contrast, `store_route_result()` (line 1044) does write the model to DB — but it's only called in Phase 3a (routing), not in Phase 3b (dispatch load via `get_route_result`).

## Fix

In `get_route_result()`, after the `.filter()` call that may discard the stale model, add a conditional DB write: if the returned `model` is `None` and the original `task.model` was non-empty (i.e., a model was discarded), update the DB to set `model = ""`. This ensures the next dispatch won't see the stale model again.

**Touch points:** `src/engine/router/mod.rs:1096–1133`

The fix is ~5 lines: read whether a discard happened, then call `store.set_fields()` with `[("model", json!(""))]` before returning the `RouteResult`.

gabrielkoerich:
## Summary

Perfect! I've fixed the issue. Here's what I did:

## Summary

**Problem:** When `get_route_result()` discarded a stale/invalid model (e.g., `opus` for an `opencode` agent that has no opus pool), the database was never updated. This caused the same stale model to persist in `tasks.model`, so every subsequent dispatch would:
1. Read the same stale model from the DB
2. Discard it again
3. Log the same warning again

**Solution:** Added a 5-line fix in `src/engine/router/mod.rs:1125-1132` that:
1. 

Closes #1907

---
*Task #1907 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/router/mod.rs
+++ b/src/engine/router/mod.rs
@@ -1122,6 +1122,15 @@ pub async fn get_route_result(
         valid
     });
 
+    // If a stale model was discarded, update the database to clear it.
+    // This prevents repeated warnings on subsequent dispatches (#1907).
+    let original_model = task.model.clone().filter(|m| !m.is_empty());
+    if original_model.is_some() && model.is_none() {
+        let _ = store
+            .set_fields(store_id, &[("model", serde_json::json!(""))])
+            .await;
+    }
+
     Ok(RouteResult {
         agent,
         model,


gabrielkoerich:

---

PR #1917, opened 2026-04-05
bug: register_session generation counter has TOCTOU race — concurrent re-registrations both get stale generation

### LINKED ISSUE #1914: bug: register_session generation counter has TOCTOU race — concurrent re-registrations both get stale generation
## Problem

In `src/channels/capture.rs`, `register_session` reads the current generation under a read lock, drops the read lock, then writes under a separate write lock:

```rust
// Lines 84-102
let generation = {
    let buffers = self.buffers.read().await;   // read lock acquired
    buffers.get(&skey).map(|b| b.generation + 1).unwrap_or(0)
};                                              // read lock DROPPED here

// ← another task can register the same skey here ←

let buffer = OutputBuffer { generation, ... };
self.buffers.write().await.insert(skey, buffer); // write lock acquired
```

## Impact

If two tasks concurrently re-register the same (repo, task_id) — which can happen when a task is re-dispatched immediately after a stuck-task recovery — both see the old generation (e.g. 0) and both write generation=1. The second write should have seen generation=1 and written generation=2.

The generation counter is used by the capture service to detect stale content from a previous session. With an incorrect generation, output chunks from the previous session may not be properly invalidated, causing stale output to be re-broadcast to subscribers (Discord, Telegram) or causing incorrect "session ended" signals.

## Fix

Merge the read and write into a single write-lock critical section to make the generation increment atomic:

```rust
{
    let mut buffers = self.buffers.write().await;
    let generation = buffers.get(&skey).map(|b| b.generation + 1).unwrap_or(0);
    let buffer = OutputBuffer {
        generation,
        repo: repo.to_string(),
        session: session.to_string(),
        task_id: task_id.to_string(),
        last_content: String::new(),
        last_len: 0,
        last_hash: None,
        last_capture: now,
        seen_alive: false,
        registered_at: now,
        has_output: false,
    };
    buffers.insert(skey, buffer);
}
```

This removes the TOCTOU window between reading and writing the generation counter.

gabrielkoerich:
## Summary

All 2657 tests pass. The fix merges the read and write into a single write-lock critical section, eliminating the TOCTOU window. The `generation` value is returned from the block so it's still available for the `tracing::debug!` call below.

Closes #1914

---
*Task #1914 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/channels/capture.rs
+++ b/src/channels/capture.rs
@@ -80,26 +80,28 @@ impl CaptureService {
         let now = Utc::now();
         let skey = session_key(repo, task_id);
 
-        // Check if we need to increment generation (re-registration case)
+        // Atomically read the current generation and insert the new buffer under
+        // a single write lock to prevent a TOCTOU race where two concurrent
+        // re-registrations both observe the same stale generation value.
         let generation = {
-            let buffers = self.buffers.read().await;
-            buffers.get(&skey).map(|b| b.generation + 1).unwrap_or(0)
+            let mut buffers = self.buffers.write().await;
+            let generation = buffers.get(&skey).map(|b| b.generation + 1).unwrap_or(0);
+            let buffer = OutputBuffer {
+                repo: repo.to_string(),
+                session: session.to_string(),
+                task_id: task_id.to_string(),
+                last_content: String::new(),
+                last_len: 0,
+                last_hash: None,
+                last_capture: now,
+                seen_alive: false,
+                registered_at: now,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started (agent: opencode, model: opencode/minimax-m2.5-free)

gabrielkoerich:

---

PR #1939, opened 2026-04-05
bug: router LLM auth error logged with trailing raw NDJSON Claude stats blob — routing failure logs unreadable

### LINKED ISSUE #1936: bug: router LLM auth error logged with trailing raw NDJSON Claude stats blob — routing failure logs unreadable
## What I Found

When the router LLM call fails with an auth error, the logged \`reason\` field contains the full raw NDJSON statistics blob from the Claude Code session output — making the log entry enormous and unreadable.

**Log entry from 2026-04-05T20:00:19Z (task internal:56488):**
\`\`\`
WARN engine.route{task_id=internal:56488}: orch::engine::router:
  pool entry failed, recording cooldown and trying next
  agent="kimi" model="haiku"
  error=router LLM failed: router LLM returned error: auth error:
  ts":0,"web_fetch_requests":0},"service_tier":"standard",
  "cache_creation":{"ephemeral_1h_input_tokens":0,...},
  "inference_geo":"","iterations":[],"speed":"standard"},
  "modelUsage":{},"permission_denials":[],
  "fast_mode_state":"off","uuid":"67a46d95-49cb-4384-ba4e-1b1cd571ab80"}
\`\`\`

Same pattern at 2026-04-05T15:12:15Z (task internal:54607) — both are ~700+ chars of raw JSON appended to the error.

## Why It's Happening

When a router LLM invocation fails, the error classification code extracts the auth error text from the agent output stream. The agent output (claude CLI) is an NDJSON stream where the auth error appears inline in a JSON event, followed by the session statistics blob at the end of the stream.

The error extractor finds the auth error text but captures everything after the error token including the trailing statistics JSON. The \`reason\` field in the log then contains:
\`"auth error: [rest of JSON stats from the output stream]"\`

## Why It Matters

- Each routing failure log entry is 500-1000+ chars of JSON noise instead of a 20-char human-readable reason
- The actual error type (\`auth error\`) is on the correct first token, but the noise makes log scanning with \`grep\` or structured log parsers noisy
- The error stored in the cooldown system (if propagated) may contain the raw JSON, making KV inspection confusing
- At high failure rates (kimi billing exhaustion hits every 4h), these log entries accumulate and inflate log storage

## Suggested Fix

In the router LLM error extraction (likely in \`src/engine/router/llm.rs\` or the agent error classifier), truncate or sanitize the error string before using it in the \`reason\` field:

Option A — Truncate to first meaningful line (simplest):
\`\`\`rust
let clean_reason = reason.lines().next().unwrap_or(&reason).trim();
// or: reason.chars().take(200).collect::<String>()
\`\`\`

Option B — Stop at JSON boundary:
\`\`\`rust
// Trim everything after the first '{' that follows non-JSON content
let clean_reason = if let Some(pos) = reason.find("\",\"") {
    &reason[..pos]
} else {
    &reason
};
\`\`\`

Option C (preferred) — Extract only the error token from the NDJSON stream before building the error string, so the raw JSON blob never enters the reason string in the first place.

## Evidence

\`\`\`bash
# Two log occurrences today
grep "auth error" /opt/homebrew/var/log/orch.log | grep "router LLM returned" | wc -l
# Each line is ~700-1000 chars; actual error type is "auth error" (10 chars)
\`\`\`

The actual error is 10 characters: \`"auth error"\`. The logged reason is ~700 characters. 98.5% noise.

gabrielkoerich:
## Summary

Clean. The fix is complete.

## Summary

**Root cause**: In `claude.rs`, error detection used `combined = format!("{result_text} {raw}")` — appending the entire raw NDJSON envelope (including the stats blob) to the result text. When `detect_auth_error` matched on `combined`, it stored `safe_tail(combined, 300)` as the error message, which captured the last 300 chars of the JSON stats blob, not the actual error text.

**Fix**: Two call sites in `claude.rs` now pass `result_text` directly to `de

Closes #1936

---
*Task #1936 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/agents/claude.rs
+++ b/src/engine/runner/agents/claude.rs
@@ -108,14 +108,16 @@ impl ClaudeRunner {
             }
             // Check auth before rate_limit — billing errors (credit balance)
             // are auth issues, not transient rate limits.
-            let combined = format!("{result_text} {}", raw);
-            if let Some(e) = super::patterns::detect_auth_error(&combined) {
+            // Use result_text only (not combined with raw JSON) so the error
+            // message stays clean — the raw NDJSON stats blob must not leak
+            // into the logged reason string.
+            if let Some(e) = super::patterns::detect_auth_error(result_text) {
                 return Err(e);
             }
-            if let Some(e) = super::patterns::detect_rate_limit(&combined) {
+            if let Some(e) = super::patterns::detect_rate_limit(result_text) {
                 return Err(e);
             }
-            if let Some(e) = super::patterns::detect_context_overflow(&combined) {
+            if let Some(e) = super::patterns::detect_context_overflow(result_text) {
                 return Err(e);
             }
             return Err(AgentError::AgentFailed {
@@ -174,7 +176,7 @@ impl ClaudeRunner {
 /// (auth, rate limit, context overflow) rather than always returning AgentFailed.
 fn extract_error_from_envelope(
     obj: &serde_json::Map<String, serde_json::Value>,
-    raw: &str,
+    _raw: &str,
 ) -> Option<AgentError> {
     if obj.get("type").and_then(|v| v.as_str()) == Some("result")
         && obj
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

---

PR #1961, opened 2026-04-06
bug: calculate_backoff_delay jitter uses wall-clock micros — concurrent retries get identical delay

### LINKED ISSUE #1959: bug: calculate_backoff_delay jitter uses wall-clock micros — concurrent retries get identical delay
## Problem

In `src/engine/runner/response.rs` (lines 78–86), the jitter for exponential backoff uses the current wall-clock microsecond timestamp as a pseudo-random seed:

```rust
let jitter_range = (capped as f64 * JITTER_FACTOR) as u64;
// Use current time micros for simple randomization
let now = std::time::SystemTime::now()
    .duration_since(std::time::UNIX_EPOCH)
    .unwrap_or_default()
    .as_micros() as u64;
let jitter = now % (jitter_range * 2 + 1);
capped.saturating_sub(jitter_range) + jitter
```

**Problem 1 — Correlated retries.** When multiple tasks fail and enter fallback retry at the same time (e.g., a cooldown expires and N tasks are simultaneously unblocked), they all call `calculate_backoff_delay` within the same microsecond interval. They receive `now % range` where `now` is essentially the same value for all of them, producing nearly identical delays. This defeats jitter's purpose of spreading retry load, causing a synchronized retry burst (thundering herd).

**Problem 2 — Modulo bias.** `now % (jitter_range * 2 + 1)` is biased toward lower values unless the range is a power of two dividing evenly into u64::MAX. For large `jitter_range` values this skews distribution significantly.

## Fix

Replace with a thread-local PRNG or OS-provided randomness. The `rand` crate is already available or `std::random` (Rust 2024). Minimal fix:

```rust
// Use a fast, non-crypto PRNG for jitter
let jitter = {
    // Mix in task/thread identity for cheap variation without a full PRNG dep
    use std::collections::hash_map::DefaultHasher;
    use std::hash::{Hash, Hasher};
    let mut h = DefaultHasher::new();
    std::time::SystemTime::now().hash(&mut h);
    std::thread::current().id().hash(&mut h);
    (h.finish() as u64) % (jitter_range * 2 + 1)
};
```

Or simply pull in `rand` (already a transitive dependency via tokio):
```rust
let jitter = rand::random::<u64>() % (jitter_range * 2 + 1);
```

**File:** `src/engine/runner/response.rs` lines 78–86  
**Severity:** Medium — reduces effectiveness of retry jitter during burst failures; can amplify GitHub rate-limit pressure

gabrielkoerich:
## Summary

All tests pass. The fix is clean — 2/2 tests passed, fmt and clippy are clean.

**Summary of the change** (`src/engine/runner/response.rs:77-91`):

Replaced the wall-clock microsecond seed (`now % range`) with a `DefaultHasher` that mixes `SystemTime::now()` (nanosecond precision) with the current thread ID. This means:

- **Concurrent retries on different threads** get different hash inputs (different thread IDs) → different jitter values even when called in the same microsecond
- **No new 

Closes #1959

---
*Task #1959 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/response.rs
+++ b/src/engine/runner/response.rs
@@ -74,14 +74,19 @@ pub(crate) fn calculate_backoff_delay(retry_count: usize) -> u64 {
     let exponential = base_delay * (2_u64.saturating_pow(retry_count as u32));
     let capped = exponential.min(max_delay);
 
-    // Add jitter: ±30% using a simple hash-based approach
+    // Add jitter: ±30% using a hash-based pseudo-random value.
+    // Mix system time (nanoseconds) with thread ID so concurrent retries on
+    // different threads get different delays even when called in the same
+    // microsecond, avoiding the thundering-herd problem.
     let jitter_range = (capped as f64 * JITTER_FACTOR) as u64;
-    // Use current time micros for simple randomization
-    let now = std::time::SystemTime::now()
-        .duration_since(std::time::UNIX_EPOCH)
-        .unwrap_or_default()
-        .as_micros() as u64;
-    let jitter = now % (jitter_range * 2 + 1);
+    let jitter = {
+        use std::collections::hash_map::DefaultHasher;
+        use std::hash::{Hash, Hasher};
+        let mut h = DefaultHasher::new();
+        std::time::SystemTime::now().hash(&mut h);
+        std::thread::current().id().hash(&mut h);
+        h.finish() % (jitter_range * 2 + 1)
+    };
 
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

gabrielkoerich:
🔍 Automated review started (agent: opencode, model: opencode/nemotron-3-super-free)

PR #1981, opened 2026-04-06
bug: WeightSignal::None unconditionally maps to NeedsReview — tasks stuck forever when enable_review_agent=false

### LINKED ISSUE #1976: bug: WeightSignal::None unconditionally maps to NeedsReview — tasks stuck forever when enable_review_agent=false
## Problem

In `src/engine/tick.rs` line 1109, `WeightSignal::None` always maps to `"needs_review"` regardless of the `workflow.enable_review_agent` config:

```rust
WeightSignal::Success { .. } if enable_review => { /* needs_review if has_pr, else done */ }
WeightSignal::Success { .. } => "done",
WeightSignal::RateLimited { .. } => "new",
WeightSignal::Blocked => "blocked",
WeightSignal::None => "needs_review",   // ← always, no enable_review guard
```

Then at line 1125, the task is transitioned to `NeedsReview` unconditionally. When `enable_review = false`:

1. **Subscriber** (`src/engine/subscribers/review.rs` lines 51–53) skips the event: `if !enable_review { continue; }`
2. **Catch-up loop** (`src/engine/sync.rs` line 654) is gated on `if enable_review { ... }` and skips stale `NeedsReview` tasks

Result: tasks that complete with `WeightSignal::None` are permanently stuck in `NeedsReview` with no automated recovery path. The only escape is manual `orch task unblock`.

## When does WeightSignal::None occur?

- `src/engine/runner/mod.rs` line 852: runner guard skipped the task (concurrent duplicate dispatch detected)
- `src/engine/runner/mod.rs` line 926: task final status is not `success`/`done`/`needs_review`/`in_progress`/`in_review`/`blocked`/`new`/`routed`

## Secondary issue

The second `enable_review` read inside `if display_status == "needs_review"` (tick.rs lines 1119–1121) and the `else if enable_review { ... }` block (lines 1134–1138) are dead code — they only log a debug message. The subscriber handles dispatch, not this path. But because this secondary check doesn't guard the `update_task_status(NeedsReview)` call, tasks still get stuck.

## Fix

In tick.rs, when `enable_review = false` and the signal is `WeightSignal::None`, map to `"done"` instead of `"needs_review"` (or route back to `"new"` if the intent is to retry). Alternatively, move the `enable_review` guard to wrap the entire `display_status == "needs_review"` transition so that `NeedsReview` is only set when the review agent will actually process it.

Remove the unreachable second `enable_review` read (lines 1119–1121) and the no-op `else if enable_review` block (lines 1134–1138).

gabrielkoerich:
## Summary

I'll analyze the bug and implement the fix. Let me start by reading the relevant source files.Now let me look at the broader context around line 1109 and also the sync.rs and review subscriber:Now I have a clear picture. Let me implement the fix:

1. `WeightSignal::None` should map to `"done"` when `enable_review = false` (or always `"done"` — since `None` means guard-skipped or unexpected status, not a successful completion needing review)
2. Remove the dead code: the second `enable_review` r

Closes #1976

---
*Task #1976 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/claude-sonnet-4.6`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -1106,7 +1106,12 @@ pub(crate) async fn tick_dispatch_tasks(
                         WeightSignal::Success { .. } => "done",
                         WeightSignal::RateLimited { .. } => "new",
                         WeightSignal::Blocked => "blocked",
-                        WeightSignal::None => "needs_review",
+                        // None means the runner guard skipped the task or the task
+                        // completed with an unrecognised status string. Only route to
+                        // needs_review when the review agent is actually enabled;
+                        // otherwise mark done so the task is not permanently stuck.
+                        WeightSignal::None if enable_review => "needs_review",
+                        WeightSignal::None => "done",
                     };
 
                     // Send weight signal back to the router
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

PR #1982, opened 2026-04-06
bug: auto_unblock redundant first set_fields(count=0) when reason_changed — intermediate state observable by concurrent ticks

### LINKED ISSUE #1977: bug: auto_unblock redundant first set_fields(count=0) when reason_changed — intermediate state observable by concurrent ticks
## Problem

In `src/engine/sync.rs`, when `reason_changed` is true, the auto_unblock path makes **two separate `set_fields` calls**:

```rust
// Line 328–336: First write — resets counter only
if reason_changed {
    if let Err(e) = store.set_fields(task.id, &[("auto_unblock_count", json!(0))]).await {
        // returns on error
        continue;
    }
}

// Line 362–386: Second write — sets counter to new_count + all other fields
let new_count = cooldown_count + 1;  // = 0 + 1 = 1 when reason_changed
store.set_fields(task.id, &[
    ("auto_unblock_count", json!(new_count)),  // sets to 1
    ("auto_unblock_last_at", json!(now)),
    ("auto_unblock_last_reason", json!(reason_key)),
    ...
]).await
```

The comment at line 365 says _"Combine all field updates — including the counter increment — into a single atomic `set_fields` call"_ but this is misleading: when `reason_changed`, there is **also** the prior reset write at line 330.

## Why this matters

Between the two writes, the DB transiently has:
- `auto_unblock_count = 0`
- `auto_unblock_last_reason = old_reason` (not yet updated)
- `auto_unblock_last_at = old_time` (not yet updated)

A concurrent tick reading the task in this window sees `reason_changed = true` (old reason still in DB) and `cooldown_count = 0` (freshly reset). It passes `auto_unblock_cooldown_elapsed` (count=0 means no cooldown required) and proceeds to repeat the first write — then races to complete the second write.

This creates a window for **two concurrent ticks to both enter the auto_unblock path for the same task**, potentially dispatching two redundant unblocks.

## Fix

Remove the first `set_fields` call entirely (lines 328–336). The second call at line 383 already handles the reset atomically by writing `auto_unblock_count = new_count` (which is 1 when `reason_changed`) together with all other fields. The counter reset is subsumed by the increment.

Update the comment to accurately describe the single-write approach.

gabrielkoerich:
## Summary

I've fixed the bug in `src/engine/sync.rs`. Here's what was done:

### The Problem
When `reason_changed` was true in the auto_unblock path, the code made **two separate database writes**:
1. First write (lines 328-336): Reset `auto_unblock_count` to 0
2. Second write (lines 383-386): Set `auto_unblock_count` to 1 plus update reason and timestamp

Between these writes, a concurrent tick could observe the intermediate state (counter=0, old reason still in DB) and incorrectly re-enter t

Closes #1977

---
*Task #1977 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -320,22 +320,13 @@ async fn auto_unblock_blocked_tasks(
         // Determine the most recent failure category as the "reason key" for this unblock.
         let reason_key = failures.first().map(|f| f.as_str()).unwrap_or("");
 
-        // Reset counter when the failure reason changes — exponential backoff should
-        // only accumulate for repeated identical failures, not different ones.
-        // This check must run BEFORE the cooldown check so new reasons bypass old cooldown.
+        // Detect reason changes to reset exponential backoff.
+        // New reasons bypass old cooldown windows and start fresh attempt counters.
         let current_reason = task.auto_unblock_last_reason.clone();
         let reason_changed = reason_key != current_reason;
-        if reason_changed {
-            if let Err(e) = store
-                .set_fields(task.id, &[("auto_unblock_count", serde_json::json!(0))])
-                .await
-            {
-                tracing::warn!(task_id = task.id, err = %e, "failed to reset auto_unblock counter for new reason — skipping");
-                continue;
-            }
-        }
 
-        // Check cooldown after reason-change detection (so new reasons bypass old cooldown).
+        // Compute effective counter: when the reason changes, start from 0;
+        // otherwise use the accumulated count for exponential backoff.
         let cooldown_count = if reason_changed {
             0
         } else {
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

The fix correctly removes the redundant first set_fields call that created a race condition window. When reason_changed is true, cooldown_count=0 so new_count=1, meaning the second set_fields already handles the reset atomically. The change is minimal (only sync.rs), CI passes, and comments are updated to accurately describe the single-write approach.

---
*Reviewed opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

PR #2015, opened 2026-04-06
bug: GitHub token embedded in git command-line arguments — visible via ps to local users

### LINKED ISSUE #2007: bug: GitHub token embedded in git command-line arguments — visible via ps to local users
## Root Cause

`src/engine/runner/git_ops.rs` lines 311-345, `build_git_auth_args()` embeds the GitHub token directly into git `-c` command-line arguments:

```rust
let authed = format!("url.https://x-access-token:{t}@github.com/.insteadOf");
vec![
    "-c".to_string(),
    format!("{authed}=https://github.com/"),
    "-c".to_string(),
    format!("{authed}=git@github.com:"),
]
```

These arguments contain the token in cleartext as `x-access-token:<TOKEN>@github.com`. They are passed to `git push` and other git commands. During execution, the full command line (including the token) is visible to any local user via `ps aux`, `ps -ef`, or `/proc/<pid>/cmdline` on Linux.

## Impact

On shared machines (multi-user systems, CI runners, dev containers with shared process namespace), any local user can extract the GitHub token by listing running processes while a git push is in progress. The token grants the same permissions as the authenticated user — repo read/write, issue management, etc.

For single-user local machines (the primary orch deployment target), the risk is lower but still present — any malware or background process can scrape tokens from the process list.

## Fix

Use environment variables instead of command-line arguments. Git respects `GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_N`, and `GIT_CONFIG_VALUE_N` environment variables, which are not visible in `ps`:

```rust
pub(crate) fn build_git_auth_env() -> Vec<(String, String)> {
    let token = crate::github::token::TokenResolver::default_env()
        .get_token_sync().ok().flatten();
    match token {
        Some(t) if !t.is_empty() => {
            let authed = format!("url.https://x-access-token:{t}@github.com/.insteadOf");
            vec![
                ("GIT_CONFIG_COUNT".into(), "2".into()),
                ("GIT_CONFIG_KEY_0".into(), authed.clone()),
                ("GIT_CONFIG_VALUE_0".into(), "https://github.com/".into()),
                ("GIT_CONFIG_KEY_1".into(), authed),
                ("GIT_CONFIG_VALUE_1".into(), "git@github.com:".into()),
            ]
        }
        _ => {
            vec![
                ("GIT_CONFIG_COUNT".into(), "1".into()),
                ("GIT_CONFIG_KEY_0".into(), "url.https://github.com/.insteadOf".into()),
                ("GIT_CONFIG_VALUE_0".into(), "git@github.com:".into()),
            ]
        }
    }
}
```

Then pass them via `.envs(&auth_env)` on the `Command` builder instead of `.args(&auth_args)`.

Alternatively, use a temporary git credential helper that provides the token via stdin (the most secure option, as the token never touches disk or process arguments).

gabrielkoerich:
## Summary

Fixed the security issue where GitHub tokens were embedded in git command-line arguments and visible via `ps`. Changed from using `-c` args to using `GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_N`, `GIT_CONFIG_VALUE_N` environment variables which are not visible in the process list.

### What was done

- Replaced build_git_auth_args() with build_git_auth_env() that returns environment variables instead of command-line args
- Updated push_branch() in git_ops.rs to use .envs() instead of .args()
- Updated fallback push in review.rs to use the new env-based approach
- Removed unused build_push_args() helper function
- Updated tests to verify the new environment variable behavior

Closes #2007

---
*Task #2007 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/minimax-m2.5-free`*

### DIFF
--- a/src/engine/runner/git_ops.rs
+++ b/src/engine/runner/git_ops.rs
@@ -337,11 +337,13 @@ pub async fn rebase_on_default(dir: &Path, default_branch: &str) {
 /// - `GIT_CONFIG_VALUE_N` — config value for entry N
 pub(crate) fn build_git_auth_env() -> Vec<(String, String)> {
     let token = crate::github::token::TokenResolver::default_env()
-        .with_gh_fallback(false)
         .get_token_sync()
         .ok()
         .flatten();
+    build_git_auth_env_for_token(token)
+}
 
+fn build_git_auth_env_for_token(token: Option<String>) -> Vec<(String, String)> {
     match token {
         Some(t) if !t.is_empty() => {
             let authed = format!("url.https://x-access-token:{t}@github.com/.insteadOf");
@@ -1246,23 +1248,9 @@ mod tests {
 
     #[test]
     fn build_git_auth_env_without_token_falls_back_to_ssh_https_conversion() {
-        // Without a token in env the function must still return the legacy
-        // SSH→HTTPS insteadOf rule so SSH-origin repos can push.
-        // Note: we disable gh CLI fallback to ensure predictable test behavior.
-        let saved_gh = std::env::var("GH_TOKEN").ok();
-        let saved_gh2 = std::env::var("GITHUB_TOKEN").ok();
-        std::env::remove_var("GH_TOKEN");
-        std::env::remove_var("GITHUB_TOKEN");
-
-        let env = build_git_auth_env();
-
-        // Restore env
-        if let Some(v) = saved_gh {
-            std::env::set_var("GH_TOKEN", v);
-        }
-        if let Some(v) = saved_gh2 {
-            std::env::set_var("GITHUB_TOKEN", v);
-        }
+        // Pass None directly to test the no-token path without relying on env vars
+        // or the gh CLI, keeping the test deterministic regardless of local auth state.
+        let env = build_git_auth_env_for_token(None);
 
         // When no token is available the fallback must include the SSH insteadOf rule.
         // The SSH→HTTPS conversion sets the target URL (the rewrite destination) to git@github.com:.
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

gabrielkoerich:

gabrielkoerich:

The core fix (switching from -c args to GIT_CONFIG_* env vars) is correct and the approach is sound. However, the PR introduces a functional regression: `build_git_auth_env()` calls `.with_gh_fallback(false)` in the production function (git_ops.rs:340), disabling the `gh auth token` CLI fallback for ALL git pushes. Users who authenticate via `gh auth login` without setting GH_TOKEN/GITHUB_TOKEN env vars will silently get no token — pushes will fall back to the no-auth path and fail. The original `build_git_auth_args()` did not disable this fallback. The test comment says 'we disable gh CLI fallback to ensure predictable test behavior', but it's the production function that was changed, not just tests. The fix: remove `.with_gh_fallback(false)` from `build_git_auth_env()`. If test isolation is needed, pass a pre-built env vec in tests rather than disabling fallback in production. Additionally, the branch was rebased (2 commits applied onto main) so CI needs to re-run after the push.

**Issues Found:**
- `src/engine/runner/git_ops.rs` line 340: `.with_gh_fallback(false)` is added to the production function, disabling `gh auth token` CLI fallback for all git pushes. Users who authenticate via `gh auth login` (without GH_TOKEN/GITHUB_TOKEN env vars set) will get no token resolved, causing push failures. The original `build_git_auth_args()` did not disable this fallback. Remove `.with_gh_fallback(false)` from the production function — if test isolation is needed, configure the resolver directly in tests instead. [error]


---

PR #2017, opened 2026-04-06
bug: completed tmux sessions can still be reclaimed as stuck before runner notices exit

### LINKED ISSUE #1999: bug: completed tmux sessions can still be reclaimed as stuck before runner notices exit
## Root Cause

Issue #1864 fixed the long post-exit race by touching `updated_at` in `TaskRunner::run_task_with_context` after `run_agent_session()` returns. That does **not** cover the smaller race between tmux session death and the runner noticing that death.

Today the flow is:

1. `tick_check_session_completions()` sees an inactive tmux session and immediately unregisters capture / kills the tmux session (`src/engine/tick.rs:97-114`)
2. `run_agent_session()` is still inside `tmux.wait_for_completion(..., poll_interval=5s)` and has not returned yet (`src/engine/runner/session.rs:65-76`)
3. Phase 2 stuck-task recovery runs before the runner wakes up, sees `status=in_progress`, `has_session=false`, and old `updated_at`, then resets the task to `New` (`src/engine/tick.rs:440-541`, `560-611`)
4. Only after that does the runner return and call `store_touch_updated_at()` (`src/engine/runner/mod.rs:445-450`)

So the previous fix only protects the post-processing window **after** `run_agent_session()` returns; it does not protect the 0-5s gap introduced by the session wait polling loop.

## Evidence

From `orch log 200` on 2026-04-06 for `internal:63714`:

```
2026-04-06T20:25:13 capture: session ended, sending final chunk
2026-04-06T20:25:17 WARN tick.phase2.stuck_tasks: recovering stuck task: no session found — reclaiming early → new task_id="internal:63714" age_mins=24 threshold_mins=10
2026-04-06T20:25:17 INFO runner::session: agent session completed task_id="internal:63714"
2026-04-06T20:25:17 INFO runner: task already reset by silence detection — skipping fallback processing status="new"
2026-04-06T20:25:32 route/dispatch starts attempt again
```

The task was reclaimed **4 seconds before** the runner observed completion. That is enough to trigger a full re-route and duplicate work.

## Impact

- Completed tasks can be spuriously re-routed and re-run
- Planning/review jobs may create duplicate issues or repeat analysis
- Attempt counters and failure recovery state get polluted by fake retries
- The existing `updated_at` fix gives a false sense that this class of race is closed

## Likely Fix

Make completion visible to stuck-task recovery at the moment Phase 1 notices the session is done, not several seconds later when the runner poll loop wakes up. A few plausible approaches:

- Touch `updated_at` (or set an explicit completion/post-processing marker) inside `tick_check_session_completions()` when the session first transitions inactive
- Or teach stuck recovery to skip tasks with a recently completed session / capture final chunk
- Or remove the polling gap by making the runner completion path event-driven instead of waiting up to 5 seconds

Whatever fix lands should add a regression test that simulates:

1. task older than `no_session_stuck_timeout`
2. session becomes inactive
3. tick Phase 2 runs before runner post-processing updates the store
4. task is **not** reset to `New`

## Likely Touch Points

- `src/engine/tick.rs`
- `src/engine/runner/session.rs`
- `src/engine/runner/mod.rs`
- stuck-task recovery tests in `src/engine/tick.rs`

gabrielkoerich:
## Summary

**Extends PR #2009** — the core race fix in #2009 only works for external tasks; internal tasks remain vulnerable.

### Root Cause of the Remaining Bug

PR #2009 (\`b2eca437\`) fixed the race by calling \`store_touch_updated_at(&session.task_id)\` in \`tick_check_session_completions\`. However:

- \`session.task_id\` comes from \`parse_session_name()\` which returns \`"internal-{n}"\` (hyphen, because tmux session names sanitize colons)
- \`resolve_task_id\` only handles the \`"internal:"\` prefix (colon form)
- For internal tasks, the touch silently no-ops — \`resolve_task_id\` returns \`Ok(None)\` and the race remains

The evidence in issue #1999 was for task \`internal:63714\` — an internal task. PR #2009's test used an external task, so it passed without catching the gap.

### Fix

In \`tick_check_session_completions\`, convert \`session.task_id\` back to the store form before calling \`store_touch_updated_at\`:

- Strip the \`"-review"\` suffix (review sessions are tracked under the main task)  
- Convert \`"internal-{n}"\` → \`"internal:{n}"\`

### Regression Test

Added \`phase1_session_completion_touch_prevents_phase2_reclaim\` which:
1. Creates an **internal** task in \`InProgress\` with a stale \`updated_at\`
2. Simulates Phase 1 touching \`updated_at\` via \`store_touch_updated_at\`
3. Runs Phase 2 stuck-task recovery
4. Asserts the task is **not** reclaimed

Closes #1999

---
*Task #1999 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -76,7 +76,9 @@ async fn startup_cleanup(tmux: &TmuxManager) {
 }
 
 use super::EngineConfig;
-use crate::store::{review_session_expected, set_review_session_expected, store_set_by_id};
+use crate::store::{
+    review_session_expected, set_review_session_expected, store_set_by_id, store_touch_updated_at,
+};
 
 /// Phase 1 of tick: poll tmux for finished sessions and clean them up.
 pub(crate) async fn tick_check_session_completions(
@@ -106,8 +108,23 @@ pub(crate) async fn tick_check_session_completions(
             // reclaim this task while the runner is still finishing
             // post-processing (race: session observed dead → tick reclaim).
             // Best-effort: no-ops if the task isn't present in the store.
-            crate::store::store_touch_updated_at(&Some(Arc::clone(store)), repo, &session.task_id)
-                .await;
+            //
+            // session.task_id uses hyphens (e.g. "internal-63714") because the
+            // tmux session name sanitizes colons to hyphens. resolve_task_id only
+            // handles "internal:" prefix, so we must convert back before touching.
+            // Review sessions carry a "-review" suffix tracked under the main task.
+            let store_task_id = {
+                let base = session
+                    .task_id
+                    .strip_suffix("-review")
+                    .unwrap_or(&session.task_id);
+                if let Some(n) = base.strip_prefix("internal-") {
+                    format!("internal:{n}")
+                } else {
+                    base.to_string()
+                }
+            };
+            store_touch_updated_at(&Some(Arc::clone(store)), repo, &store_task_id).await;
             // Unregister from capture service using the task id the capture service
             // was registered under (task id without project prefix).
             capture.unregister_session(repo, &session.task_id).await;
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

---

PR #2050, opened 2026-04-06
bug: NeedsReview re-fire in sync.rs has no backoff — stuck tasks generate log spam and event churn every 45s

### LINKED ISSUE #2039: bug: NeedsReview re-fire in sync.rs has no backoff — stuck tasks generate log spam and event churn every 45s
## Problem

In `src/engine/sync.rs:662-698`, the sync catch-up loop re-fires `NeedsReview` events for any task that has been in `NeedsReview` status for more than 1 minute. This runs every sync interval (45s). There is **no backoff, no retry counter, and no escalation threshold**.

When a task cannot be reviewed (e.g., all review agent slots are full, the review subscriber cannot process it, or the review agent keeps crashing), this creates:

1. **Log spam**: `sync catch-up: re-firing NeedsReview event for stale task` appears every 45 seconds per stuck task, indefinitely
2. **Wasted DB writes**: `update_task_status(NeedsReview)` on a task already in `NeedsReview` is a no-op that still triggers side effects
3. **Event bus churn**: Each re-fire triggers the review subscriber again, which may attempt and fail to spawn a review agent

This was directly observed in live logs where task 1999 was re-fired 3+ times in succession while review slots were occupied by other tasks.

## Root cause

The re-fire logic at `sync.rs:684-698` has no memory of previous re-fires. The `MIN_STALE_NEEDS_REVIEW_MINUTES = 1` threshold is a one-time gate, not an exponentially increasing one. Once a task passes the 1-minute mark, it gets re-fired on every sync tick forever.

## Suggested fix

Add backoff to the re-fire mechanism. Options (in order of preference):

**Option A (recommended)**: Track re-fire count via `store_increment(task_id, "needs_review_refires")` and apply exponential backoff to the stale threshold: `min_stale_minutes = 1 * 2^refires`. Reset the counter when the task transitions away from `NeedsReview`. After N re-fires (e.g., 5), escalate to `blocked` with a reason like "review agent failed to process after 5 attempts".

**Option B**: Only re-fire if `updated_at` was NOT already touched by a previous re-fire within the current sync window. Since `update_task_status` updates `updated_at`, a re-fired task will have `updated_at` within the last 45s and be skipped by the 1-minute threshold on the next tick. This naturally creates ~90s spacing between re-fires.

**Option C**: Skip re-fire entirely when the `dispatching` set already contains any review tasks (indicating review slots are saturated).

## Files to change

- `src/engine/sync.rs:649-699` — add backoff logic to the NeedsReview re-fire loop
- `src/store.rs` (optional) — add helper if using `store_increment` approach

gabrielkoerich:
## Summary

I'll inspect the sync.rs code around the referenced lines to understand the current implementation before making a change. Then I'll implement Option A (refire counter with exponential backoff and escalation to blocked after 5 attempts), using the store increment helper if available or adding a small helper to the store module. I'll update sync.rs and store.rs as needed and run a quick compile check. I'll start by reading the existing file.---
## Goal

Fix the runaway "re-fire NeedsReview" behav

Closes #2039

---
*Task #2039 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

### DIFF
--- a/src/engine/runner/response.rs
+++ b/src/engine/runner/response.rs
@@ -976,6 +976,7 @@ mod tests {
             review_agent_failures: 0,
             review_cycles: 0,
             review_invocations: 0,
+            needs_review_refires: 0,
             review_session_expected: false,
             input_tokens: 0,
             output_tokens: 0,


gabrielkoerich:
⚠️ Review agent failed (attempt 1/3)

**Reason:** parse error: failed to parse review response

Retrying with a different agent.

PR #2073, opened 2026-04-07
bug: push_failures silently returns 0 on DB error — blocking threshold unreachable when store is flaky

### LINKED ISSUE #2069: bug: push_failures silently returns 0 on DB error — blocking threshold unreachable when store is flaky
## Problem

In `src/engine/runner/response_handler.rs` lines 615–630, `push_failures` is incremented via `store_increment`. On DB failure, it silently falls back to `0`:

```rust
let push_failures: u64 = if push_failed && !is_workflow_scope_failure {
    match store::store_increment_by_id(store, store_id, "push_failures").await {
        Ok(v) => v,
        Err(e) => {
            tracing::warn!(...);
            0   // ← silently returns 0
        }
    }
    // ...
} else {
    0
};
```

This value feeds directly into `classify_final_status` (line 690), which uses it to decide whether to block the task:

```rust
fn classify_final_status(input: &DecisionInput<'_>) -> String {
    if input.push_failures >= 3 {
        return "blocked".to_string();  // ← never reached on DB error
    }
    // ... re-route logic ...
}
```

## Impact

When the store is degraded (transient DB lock, WAL contention), every `push_failures` increment returns `0` instead of the real count. The blocking threshold `>= 3` is never reached. A task that has repeatedly failed to push will be re-routed indefinitely as if it's a fresh failure, rather than being escalated to `blocked` for human review.

This means push failures during store degradation create a silent spin loop — the task burns retry budget without ever surfacing to the operator.

## Files

- `src/engine/runner/response_handler.rs` lines 615–630
- `src/engine/runner/response_handler.rs` line 690 (`classify_final_status` call)
- `src/engine/runner/response_handler.rs` lines 772–782 (push_failures block comment)

## Fix

On DB error when incrementing `push_failures`, read the **current** value from the store instead of defaulting to 0:

```rust
Err(e) => {
    tracing::warn!(...);
    // Fall back to current stored value so the blocking threshold still fires
    store::opt_store_get_task(store, repo, task_id).await
        .and_then(|t| t)
        .and_then(|t| t.push_failures)
        .unwrap_or(1)  // assume at least 1 if we can't read either
}
```

Or alternatively, read the current count before incrementing and use `current_count + 1` on increment failure, ensuring the threshold is always reachable.

gabrielkoerich:
## Fix Complete ✓

I've successfully implemented the fix for issue #2069. Here's what was done:

### Changes
Modified `src/engine/runner/response_handler.rs` to handle database errors gracefully in two counter increments:

**Before (buggy behavior):**
- DB error on increment → returns 0
- Blocking thresholds never reached → silent reroute spin loop

**After (fixed behavior):**
- DB error on increment → read current stored value → use (current + 1)
- Blocking thresholds properly enforce

Closes #2069

---
*Task #2069 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/runner/response_handler.rs
+++ b/src/engine/runner/response_handler.rs
@@ -618,16 +618,24 @@ pub async fn handle_success(
                 match store::store_increment_by_id(store, store_id, "push_failures").await {
                     Ok(v) => v,
                     Err(e) => {
-                        tracing::warn!(task_id, err = %e, "failed to increment push_failures — skipping push-failure based reroute this tick");
-                        0
+                        tracing::warn!(task_id, err = %e, "failed to increment push_failures — reading current value from store");
+                        // On DB error, read current value to ensure blocking threshold is still reachable
+                        store::opt_store_get_task_by_id(store, store_id)
+                            .await
+                            .map(|t| (t.push_failures as u64) + 1)
+                            .unwrap_or(1) // assume at least 1 if we can't read either
                     }
                 }
             }
             None => match store::store_increment(store, repo, task_id, "push_failures").await {
                 Ok(v) => v,
                 Err(e) => {
-                    tracing::warn!(task_id, err = %e, "failed to increment push_failures — skipping push-failure based reroute this tick");
-                    0
+                    tracing::warn!(task_id, err = %e, "failed to increment push_failures — reading current value from store");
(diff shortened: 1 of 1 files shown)


gabrielkoerich:

---

PR #2079, opened 2026-04-07
bug: ingest_external_tasks silently skips status sync on batch-get failure with no log

### LINKED ISSUE #2077: bug: ingest_external_tasks silently skips status sync on batch-get failure with no log
## Problem

In `src/engine/sync.rs` around line 1643–1657, `ingest_external_tasks` uses `unwrap_or_default()` on a batch store query:

```rust
if !id_status_pairs.is_empty() {
    let all_ids: Vec<i64> = id_status_pairs.iter().map(|(id, _)| *id).collect();
    let existing_map = store.get_batch(&all_ids).await.unwrap_or_default();  // silent failure
    for (store_id, status) in id_status_pairs {
        if let Some(existing) = existing_map.get(&store_id) {
            if existing.status == TaskStatus::New {
                // sync status from GitHub → store for first-ingest tasks
            }
        }
    }
}
```

If `get_batch` fails, `existing_map` is an empty `HashMap`, and the inner `if let Some(existing)` never matches — so **no status sync happens for any newly ingested tasks**. The failure is completely silent: no log, no metric, no error propagation.

## Impact

On DB errors during ingest, first-ingest tasks that arrive via GitHub with a non-`New` status (e.g., already labeled `status:routed`) won't have their initial status applied. This can delay correct routing for incoming tasks and makes operational debugging harder — there's no signal in logs that the sync was skipped.

## Fix

Add a warning log when the batch fetch fails so operators can detect and investigate the issue:

```rust
let existing_map = match store.get_batch(&all_ids).await {
    Ok(map) => map,
    Err(e) => {
        tracing::warn!(err = %e, task_count = all_ids.len(), "ingest: get_batch failed — skipping initial status sync for newly ingested tasks this tick");
        return Ok(());  // or continue; depending on desired behavior
    }
};
```

gabrielkoerich:
## Summary

✅ **Fixed issue #2077** — Added proper error handling to `ingest_external_tasks()` in `src/engine/sync.rs:1645`

### The Problem
The code was using `unwrap_or_default()` on a batch store query, causing silent failures. If `get_batch()` failed, it would return an empty `HashMap`, and no status sync would happen for newly ingested tasks — with no log message to alert operators.

### The Solution
Replaced the silent failure with explicit error handling:
- Added a `match` expressio

Closes #2077

---
*Task #2077 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -1647,7 +1647,13 @@ pub(crate) async fn ingest_external_tasks(
     // (e.g., store has Routed but GitHub still shows New labels).
     if !id_status_pairs.is_empty() {
         let all_ids: Vec<i64> = id_status_pairs.iter().map(|(id, _)| *id).collect();
-        let existing_map = store.get_batch(&all_ids).await.unwrap_or_default();
+        let existing_map = match store.get_batch(&all_ids).await {
+            Ok(map) => map,
+            Err(e) => {
+                tracing::warn!(err = %e, task_count = all_ids.len(), "ingest: get_batch failed — skipping initial status sync for newly ingested tasks this tick");
+                return Ok(());
+            }
+        };
         for (store_id, status) in id_status_pairs {
             if let Some(existing) = existing_map.get(&store_id) {
                 if existing.status == TaskStatus::New {


gabrielkoerich:

The PR has a critical regression: it fixes unwrap_or_default() in ingest_external_tasks (correct) but removes existing error handling in scan_mentions (lines 899-907), replacing a proper match block that skips mention scanning on DB failure with unwrap_or_default(). If that DB query fails, existing_mentions becomes an empty HashSet, causing every historical mention to be treated as new — creating duplicate internal tasks. The original code had a comment explicitly warning about this: 'avoid treating every historical mention as new (which would create duplicate internal tasks during transient DB failures)'. The scan_mentions change is entirely out of scope for issue #2077, which only describes the ingest_external_tasks bug. The fix should ONLY touch the get_batch call around line 1645.

**Issues Found:**
- `src/engine/sync.rs` line 899: CRITICAL REGRESSION: The PR removes existing error handling in scan_mentions (lines 896-912 on main). The original code had a match block that logged a warning and returned early if list_source_ids_by_source failed, to prevent duplicate internal tasks. The PR replaces this with unwrap_or_default(), which silently returns an empty HashSet on DB failure, causing every historical mention to be re-ingested as new. The original code even had a comment documenting this risk. Fix: restore the original match block for scan_mentions, keeping only the ingest_external_tasks fix at line 1645. [error]


---

gabrielkoerich:
## Automated Review — Approve

Fix correctly replaces `unwrap_or_default()` with proper error handling. Adds warning log with error details and task count, then returns early to prevent silent status sync failures. Matches the requested fix exactly. CI passes (test + secrets), branch is up to date.

---

gabrielkoerich:

---

PR #2080, opened 2026-04-07
bug: NeedsReview refire counter reads 0 on store failure — tasks re-fire indefinitely during DB degradation

### LINKED ISSUE #2075: bug: NeedsReview refire counter reads 0 on store failure — tasks re-fire indefinitely during DB degradation
## Problem

In `src/engine/sync.rs` around line 690–698, the stale NeedsReview handler reads the per-task refire counter to enforce backoff:

```rust
let store_task =
    crate::store::opt_store_get_task(&Some(Arc::clone(store)), repo, &task.id.0).await;
let current_refires = store_task
    .as_ref()
    .map(|t| t.needs_review_refires as u64)
    .unwrap_or(0);  // ← 0 on store failure
let new_refires = current_refires + 1;

if new_refires >= MAX_NEEDS_REVIEW_REFIRE_ATTEMPTS {
    // escalate to blocked
}
```

If `opt_store_get_task` returns `None` (DB error, task not found in store), `current_refires` is **always 0**. As a result:
- `new_refires = 1`, which is always less than `MAX_NEEDS_REVIEW_REFIRE_ATTEMPTS`
- The task re-fires every tick unconditionally
- The subsequent `store_increment(..., "needs_review_refires")` also fails for the same reason (DB error), so the counter is never actually incremented
- On the next tick: same sequence repeats — infinite re-fires

## Impact

During any period of DB degradation, all `NeedsReview` tasks bypass the exponential backoff and spawn review agents on every sync tick. This wastes agent quota and can create multiple concurrent review sessions for the same task.

## Fix

Skip the re-fire (not just log) when the store read fails, since we cannot make a correct backoff decision:

```rust
let store_task =
    crate::store::opt_store_get_task(&Some(Arc::clone(store)), repo, &task.id.0).await;
let current_refires = match store_task.as_ref() {
    Some(t) => t.needs_review_refires as u64,
    None => {
        tracing::warn!(task_id = task.id.0, "failed to read task from store for refire backoff — skipping re-fire this tick");
        continue;
    }
};
```

gabrielkoerich:
## Summary

Perfect! ✅ Bug fixed and committed.

## Summary

Fixed the NeedsReview refire counter bug in `src/engine/sync.rs:695-702`:

**The Problem:** When `opt_store_get_task` returns `None` (DB error/missing task), the code defaulted `current_refires` to 0, causing `new_refires=1` on every tick. This bypassed exponential backoff, triggering infinite re-fires during DB degradation.

**The Fix:** Changed from `unwrap_or(0)` to a `match` statement that:
- Extracts the counter when the task is found
- Log

Closes #2075

---
*Task #2075 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -692,10 +692,13 @@ pub(crate) async fn sync_tick(
             // touching updated_at before we know whether we will actually fire.
             let store_task =
                 crate::store::opt_store_get_task(&Some(Arc::clone(store)), repo, &task.id.0).await;
-            let current_refires = store_task
-                .as_ref()
-                .map(|t| t.needs_review_refires as u64)
-                .unwrap_or(0);
+            let current_refires = match store_task.as_ref() {
+                Some(t) => t.needs_review_refires as u64,
+                None => {
+                    tracing::warn!(task_id = task.id.0, "failed to read task from store for refire backoff — skipping re-fire this tick");
+                    continue;
+                }
+            };
             let new_refires = current_refires + 1;
 
             // If we've exceeded max attempts, escalate the task to Blocked with a clear reason.


gabrielkoerich:

---

PR #2085, opened 2026-04-07
bug: tmux final-output capture logs errors for normal session teardown

### LINKED ISSUE #2083: bug: tmux final-output capture logs errors for normal session teardown
## Summary
`Tmux::wait_for_completion()` treats a missing pane during final capture as an error, even though session teardown can legitimately race with the final `capture-pane` call. This produces noisy error logs after otherwise successful review/task completions.

## Evidence
- Current log at `2026-04-07T04:39:53Z`:
  - `ERROR orch::tmux: failed to capture final pane output`
  - underlying error: `capture-pane failed for orch-bean-internal-68571-review: can't find pane: orch-bean-internal-68571-review`
- The same sequence is immediately followed by a successful review parse and approval:
  - `review agent completed`
  - `reading review agent output ... file_exists=true`
  - `review agent decision received ... Approve`
- Code path:
  - `src/tmux.rs:404-423` calls `capture_pane()` after `is_session_active()` flips false.
  - `src/tmux.rs:159-162` turns any `capture-pane` failure into an error.
  - `src/tmux.rs:419-420` logs that error at `ERROR` level.

## Impact
- Operators see false-positive errors in healthy runs.
- This makes real tmux/session failures harder to spot in logs.
- The error path is especially noisy for short-lived review sessions, where output has already been written to disk and is no longer dependent on a final pane scrape.

## Likely touch points
- `src/tmux.rs`
- `src/engine/runner/session.rs`
- `src/engine/review.rs`
- tests around `wait_for_completion()` / tmux teardown behavior

## Requested change
1. Treat `can't find pane` / `can't find session` during final capture as an expected teardown race, not an error.
2. Downgrade that case to `debug` or `trace`, or return `Ok(String::new())` silently.
3. Keep logging unexpected capture failures (real tmux errors) distinctly.
4. Add a regression test around the final-capture race so the log severity does not regress.

This should reduce operator noise without changing the existing fallback behavior for callers that can continue using the persisted output files.

gabrielkoerich:
## Summary

All done. Here's a summary of what was changed:

**`src/tmux.rs`** — two fixes:

1. **`capture_pane()`** (line 165): Now detects `can't find pane` / `can't find session` in stderr and returns `Ok(String::new())` at `debug` level instead of propagating an error. This is the root fix — the race condition was in `wait_for_completion()` where `is_session_active()` returns false but the session gets cleaned up before `capture_pane` runs.

2. **`wait_for_completion()`** (line 427-437): Updated com

Closes #2083

---
*Task #2083 · Created by minimax[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -162,9 +162,13 @@ impl TmuxManager {
 
         if !output.status.success() {
             let stderr = String::from_utf8_lossy(&output.stderr);
-            // "can't find pane" / "can't find session" are expected teardown races —
-            // the session was cleaned up between the caller's activity check and this call.
-            if stderr.contains("can't find pane") || stderr.contains("can't find session") {
+            // "can't find pane" / "can't find session" / "no server running" are expected
+            // teardown races — the session was cleaned up between the caller's activity
+            // check and this call, or the tmux server exited entirely.
+            if stderr.contains("can't find pane")
+                || stderr.contains("can't find session")
+                || stderr.contains("no server running")
+            {
                 tracing::debug!(
                     session,
                     "capture_pane: session/pane already gone (teardown race)"


gabrielkoerich:

PR #2121, opened 2026-04-07
perf: handle_review_changes fetches task from store twice — second opt_store_get_task at line 1312 is redundant

### LINKED ISSUE #2118: perf: handle_review_changes fetches task from store twice — second opt_store_get_task at line 1312 is redundant
## Root Cause

`handle_review_changes` in `src/engine/auto_merge.rs` fetches the task record from the store twice:

1. **Line 1087**: `let task_store_record = opt_store_get_task(...)` — to read `review_cycles` and check against `max_cycles`
2. **Line 1312**: `let task_store_record = opt_store_get_task(...)` — shadow-rebind to read `agent` and `complexity` for model routing

The second fetch is always issued in the non-escalation path (after `review_cycles < max_cycles` passes). The `agent` and `complexity` fields needed at line 1312 are already present in the first `task_store_record` from line 1087 — no writes between lines 1087 and 1312 modify those fields.

## Impact

Two DB round-trips per review cycle where one would suffice. `handle_review_changes` is called for every `CHANGES_REQUESTED` review event, so this doubles the store reads on every review feedback loop.

## Fix

Remove the redundant fetch at line 1312 and reuse the `task_store_record` obtained at line 1087:

```rust
// Line 1312 — delete this, use the one from line 1087 instead
// let task_store_record = opt_store_get_task(&Some(Arc::clone(store)), repo, &task.id.0).await;
let previous_agent = task_store_record  // already bound at line 1087
    .as_ref()
    .and_then(|t| t.agent.as_ref())
    ...
```

gabrielkoerich:
## Summary

The fix is complete. I removed the redundant `opt_store_get_task` call at line 1312 and updated the comment to clarify that `task_store_record` comes from the fetch at line 1087. Clippy passes cleanly.

Closes #2118

---
*Task #2118 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/qwen3.6-plus-free`*

### DIFF
--- a/src/engine/auto_merge.rs
+++ b/src/engine/auto_merge.rs
@@ -1308,10 +1308,10 @@ pub(crate) async fn handle_review_changes(
     // model directly produced invalid combos after failover (e.g. opencode:opus
     // when "opus" was the review agent's model, not a valid opencode model).
     //
-    // We read the task's existing agent and complexity from the store, then
-    // resolve a fresh model from the agent's configured pool.  If all models
-    // for that agent are cooled, we fall back to Status::New for a full re-route.
-    let task_store_record = opt_store_get_task(&Some(Arc::clone(store)), repo, &task.id.0).await;
+    // We read the task's existing agent and complexity from the store record
+    // fetched at line 1087, then resolve a fresh model from the agent's
+    // configured pool.  If all models for that agent are cooled, we fall back
+    // to Status::New for a full re-route.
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started (agent: opencode, model: github-copilot/gpt-5-mini)

PR #2122, opened 2026-04-07
bug: mention cursor (last_success_ts) never advances for already-processed or skipped mentions — re-fetches all historical mentions every tick

### LINKED ISSUE #2115: bug: mention cursor (last_success_ts) never advances for already-processed or skipped mentions — re-fetches all historical mentions every tick
## Root Cause

`scan_for_mentions` in `src/engine/sync.rs` uses a `last_success_ts` cursor that advances **only** when a mention is handled via two specific success paths (command-executed at line 1111 and normal task created at line 1129). All other paths use `continue` without updating the cursor:

- **Line 927-929**: already-processed mentions (`existing_mentions.contains`) — cursor never advances past these
- **Lines 931-936**: non-@orch mentions — cursor never advances past these  
- **Lines 989-993, 1025-1030**: closed-issue / non-collaborator skips — cursor never advances past these

## Impact

When a mention that falls into one of the skip paths is the **last** mention in a fetched batch, the cursor stays before it. On the next tick:
1. The cursor points before the already-handled mention → it is re-fetched
2. `existing_mentions.contains()` hits → `continue` without updating cursor
3. Cursor stays behind → mention is re-fetched again next tick, forever

In practice: any time there is a batch ending in a skipped or already-processed mention and no new unprocessed @orch mentions arrive after it, **every subsequent tick re-fetches the entire window** of already-processed mentions. This is O(N) wasted GitHub API calls and DB reads per tick, where N = number of mentions since the last truly-new mention.

## Fix

Advance `last_success_ts` in **all** paths that fully handle a mention — not just the two "success" paths. The intent of the cursor (only retry if something actually failed) is already served by the `existing_mentions` HashSet for dedup. The cursor just needs to not get stuck.

**src/engine/sync.rs line 927-929** — already-processed path:
```rust
if existing_mentions.contains(&mention.id) {
    if last_success_ts.as_deref() < Some(mention.created_at.as_str()) {
        last_success_ts = Some(mention.created_at.clone());
    }
    continue;
}
```

**Lines 931-936** — non-@orch path: same pattern before `continue`.

**Lines 989-993** — closed-issue skip: same pattern before `continue`.

**Lines 1025-1030** — non-collaborator skip: same pattern before `continue`.

gabrielkoerich:
## Summary

Fixed mention-scan cursor (mentions_last_checked) getting stuck by advancing last_success_ts for all handled/skip paths so already-processed or skipped mentions no longer prevent the cursor from moving forward.

### What was done

- Located scan_mentions implementation in src/engine/sync.rs
- Updated scan logic to advance last_success_ts when mentions are handled or intentionally skipped (already-processed, not-addressed-to-bot, closed-issue skips, non-collaborator skips)
- Ensured cursor advancement uses the mention.created_at timestamp only when it's newer than the current cursor
- Ran git diff and a cargo check to validate changes compile

### Remaining

- Monitor on-service behavior to ensure no regressions in mention re-processing semantics (the change advances the cursor for skipped mentions which is intended)
- Optional: add unit/integration tests that simulate mention batches ending with skipped entries to prevent regressions in future

Closes #2115

---
*Task #2115 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -925,6 +925,13 @@ async fn scan_mentions(
     for mention in mentions {
         // Skip if already processed
         if existing_mentions.contains(&mention.id) {
+            // Advance the cursor past already-processed mentions so the scan
+            // doesn't get stuck re-fetching the same window when the latest
+            // mention is a duplicate. Only advance when the mention's
+            // timestamp is newer than our current cursor.
+            if last_success_ts.as_deref() < Some(mention.created_at.as_str()) {
+                last_success_ts = Some(mention.created_at.clone());
+            }
             continue;
         }
 
@@ -932,6 +939,13 @@ async fn scan_mentions(
             && !mention.body.contains("@orch")
             && !mention.body.contains("@orchestrator")
         {
+            // This mention batch may include comments that the backend
+            // returned but that are not actually addressed to the bot
+            // (e.g., legacy fetch semantics). Treat these as handled for
+            // cursor advancement so we don't repeatedly re-fetch them.
+            if last_success_ts.as_deref() < Some(mention.created_at.as_str()) {
+                last_success_ts = Some(mention.created_at.clone());
+            }
             continue;
         }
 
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
## Automated Review — Approve

Fix is correct and minimal. All skip paths (already-processed, non-@orch, closed-issue, non-collaborator) now advance last_success_ts before continuing, while error paths (failed issue state check, failed collaborator check) intentionally leave the cursor unchanged to allow retry. The comparison logic using Option ordering is sound. Branch was rebased onto main — post-rebase the PR commit only changes sync.rs (23 lines added). CI results are stale but the change is a pure addition so re-run should pass.

---

PR #2161, opened 2026-04-08
bug: cleaned_external_ids query fails to filter NULL external_id — defeats batch cleanup optimization

### LINKED ISSUE #2155: bug: cleaned_external_ids query fails to filter NULL external_id — defeats batch cleanup optimization
## Root Cause

`store/tasks.rs:1450` maps the `external_id` column to `Vec<(String,)>` but does not filter out NULL values:

```rust
// store/tasks.rs:1445-1455
pub async fn cleaned_external_ids(&self, repo: &str) -> anyhow::Result<HashSet<String>> {
    let rows: Vec<(String,)> =
        sqlx::query_as("SELECT external_id FROM tasks WHERE repo = ? AND worktree_cleaned = 1")
            .bind(repo)
            .fetch_all(&self.pool)
            .await?;
    Ok(rows.into_iter().map(|(id,)| id).collect())
}
```

Internal tasks (`origin = 'internal'`) have no `external_id` (NULL). When any internal task has `worktree_cleaned = 1`, SQLx tries to decode NULL into `String` and returns an error. The caller at `cleanup.rs:223` uses `unwrap_or_default()`, so the error is silently swallowed and `already_cleaned` becomes an empty `HashSet`. Every cleanup cycle then re-processes all already-cleaned tasks, running `cleanup_task_worktree_with_opts` unnecessarily on tasks that have nothing to clean.

Compare `existing_external_ids` at line 740 which correctly adds `AND external_id IS NOT NULL`.

## Fix

Add `AND external_id IS NOT NULL` to the query:

```rust
sqlx::query_as(
    "SELECT external_id FROM tasks WHERE repo = ? AND worktree_cleaned = 1 AND external_id IS NOT NULL",
)
```

Alternatively, filter by `origin != 'internal'` to match the intent (only external tasks have meaningful `external_id` strings usable for this check).

gabrielkoerich:
## Summary

Perfect! ✓ The fix is complete and committed.

## Summary

**Fixed:** Bug #2155 — `cleaned_external_ids` query failing to filter NULL external_id values

### The Issue
The `cleaned_external_ids()` function at `src/store/tasks.rs:1450` didn't filter NULL values, causing SQLx to fail when decoding NULL into `String` (internal tasks have no external_id). The error was silently swallowed by `unwrap_or_default()`, leaving `already_cleaned` empty and forcing re-processing of all cleaned tasks on e

Closes #2155

---
*Task #2155 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `haiku`*

### DIFF
--- a/src/store/tasks.rs
+++ b/src/store/tasks.rs
@@ -1447,7 +1447,7 @@ impl TaskStore {
         repo: &str,
     ) -> anyhow::Result<std::collections::HashSet<String>> {
         let rows: Vec<(String,)> =
-            sqlx::query_as("SELECT external_id FROM tasks WHERE repo = ? AND worktree_cleaned = 1")
+            sqlx::query_as("SELECT external_id FROM tasks WHERE repo = ? AND worktree_cleaned = 1 AND external_id IS NOT NULL")
                 .bind(repo)
                 .fetch_all(&self.pool)
                 .await?;


gabrielkoerich:

---

PR #2180, opened 2026-04-08
perf: tick_dispatch_tasks spawns tmux subprocess even when dispatchable queue is empty

### LINKED ISSUE #2175: perf: tick_dispatch_tasks spawns tmux subprocess even when dispatchable queue is empty
## Problem

`tick_dispatch_tasks` always calls `tmux.batch_session_active()` (spawns `tmux list-panes -a`) regardless of whether there are any tasks to dispatch. When the queue is empty the result is never used, making the subprocess call pure overhead.

```rust
// src/engine/tick.rs (main branch, after #2174)
if dispatchable.is_empty() {
    tracing::debug!(count = 0, "dispatchable tasks found");
} else {
    // warn/info ...
}

let sequential_delay = ...;

// ← tmux list-panes -a spawned here unconditionally, even if dispatchable is empty
let dispatch_session_map = tmux.batch_session_active().await;

for (idx, task) in dispatchable.into_iter().enumerate() { ... }
```

## Impact

`tick_dispatch_tasks` is called from two places:

1. **Tick loop phase 3b** — once per project per tick interval (~10s)
2. **Event-driven dispatch subscriber** — on every task Routed event

With 2-3 projects and frequent task transitions, this is ~20-30 subprocess invocations per minute that each spawn `tmux list-panes -a` and discard the result. Each invocation is fast (<10ms) but the subprocess creation overhead adds up.

## Fix

Move both the session map fetch and the dispatch loop inside the non-empty branch:

```rust
if dispatchable.is_empty() {
    tracing::debug!(count = 0, "dispatchable tasks found");
    return Ok(());
} else {
    tracing::info!(count = dispatchable.len(), "dispatchable tasks found");
    if is_degraded { tracing::warn!(...); }
}

let sequential_delay = ...;

// Only fetched when there are tasks to dispatch
let dispatch_session_map = tmux.batch_session_active().await;

for (idx, task) in dispatchable.into_iter().enumerate() { ... }
```

**File:** `src/engine/tick.rs` around lines 1036–1060 (main branch post-#2174)

gabrielkoerich:
## Summary

All 2741 tests pass. The fix is clean — added `return Ok(())` in the empty-queue branch so `batch_session_active()` (which spawns `tmux list-panes -a`) is only called when there are actually tasks to dispatch.

Closes #2175

---
*Task #2175 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -1035,15 +1035,16 @@ pub(crate) async fn tick_dispatch_tasks(
 
     if dispatchable.is_empty() {
         tracing::debug!(count = 0, "dispatchable tasks found");
-    } else {
-        tracing::info!(count = dispatchable.len(), "dispatchable tasks found");
-        if is_degraded {
-            tracing::warn!(
-                healthy_agents = healthy_count,
-                threshold = threshold,
-                "degraded mode: using sequential dispatch"
-            );
-        }
+        return Ok(());
+    }
+
+    tracing::info!(count = dispatchable.len(), "dispatchable tasks found");
+    if is_degraded {
+        tracing::warn!(
+            healthy_agents = healthy_count,
+            threshold = threshold,
+            "degraded mode: using sequential dispatch"
+        );
     }
 
     let sequential_delay = if is_degraded {


gabrielkoerich:
## Automated Review — Approve

Reviewed PR #2180 end-to-end. Worktree was clean, optional rebase onto origin/main was a no-op (up to date), with expected sandbox lockfile noise only (non-blocking). Required CI checks were verified with watch mode and passed (`secrets`, `test`). Full diff was read (`src/engine/tick.rs` only). Change is scoped and correct: early `return Ok(())` on empty `dispatchable` avoids unnecessary `tmux.batch_session_active()` subprocess work when no tasks are dispatchable, with no behavioral regression for non-empty queues. This aligns with documented architecture and does not touch restricted config/migration areas.

---

PR #2217, opened 2026-04-08
bug: scheduled job executes after pre-run state persistence failure — violates restart-safety invariant

### LINKED ISSUE #2214: bug: scheduled job executes after pre-run state persistence failure — violates restart-safety invariant
## Problem

`src/engine/jobs.rs` lines 363–376 have a comment that explicitly states the intent:

> "Persist last_run BEFORE execution to prevent catch-up on restart. On restart the runtime state is loaded from SQLite, so writing last_run here ensures a crash during execution won't cause the job to be considered due again immediately."

However, when `s.upsert_job_state(&state)` fails (line 369), the error is logged and execution **continues anyway** (line 379). This directly violates the stated invariant: if last_run isn't persisted and the service crashes mid-execution, the job will be considered due again on restart and execute a second time.

```rust
// src/engine/jobs.rs:363-379
state.last_run = Some(now.format("%Y-%m-%dT%H:%M:%SZ").to_string());
if let Some(s) = store {
    if let Err(e) = s.upsert_job_state(&state).await {
        tracing::error!(
            job_id = job.id,
            ?e,
            "failed to persist job state before execution"
        );
        // ← execution continues here, violating the restart-safety guarantee
    }
}
execute_job(job, &mut state, backend, store, repo, transport).await;
```

## Impact

Scheduled jobs that create GitHub issues (type `"task"`) will create **duplicate issues** on crash-restart cycles when the pre-run persistence fails. The probability is low (SQLite write failure is rare), but the consequence is silent duplicate work.

## Fix

Skip execution and `continue` to the next tick when the pre-run state write fails:

```rust
if let Some(s) = store {
    if let Err(e) = s.upsert_job_state(&state).await {
        tracing::error!(
            job_id = job.id,
            ?e,
            "failed to persist job state before execution — skipping this tick to preserve restart safety"
        );
        continue; // retry next tick instead of risking duplicate execution
    }
}
```

gabrielkoerich:
## Summary

All 2759 tests pass. The fix is a one-liner `continue` in `src/engine/jobs.rs:375` — when the pre-run state write fails, execution is now skipped for that tick rather than proceeding and risking duplicate job execution on crash-restart cycles.

Closes #2214

---
*Task #2214 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/jobs.rs
+++ b/src/engine/jobs.rs
@@ -370,8 +370,9 @@ pub async fn tick(
                 tracing::error!(
                     job_id = job.id,
                     ?e,
-                    "failed to persist job state before execution"
+                    "failed to persist job state before execution — skipping this tick to preserve restart safety"
                 );
+                continue; // retry next tick instead of risking duplicate execution
             }
         }
 


gabrielkoerich:
## Automated Review — Approve

CI passes (all required checks green), branch is rebased on main, worktree is clean. The fix is exactly as described: a single `continue` added after the error log when `upsert_job_state` fails before execution, plus an improved log message. Minimal, surgical, and correctly preserves the restart-safety invariant described in the comment at line 363.

---

PR #2433, opened 2026-04-11
bug: load_skills_catalog has double-checked locking race — concurrent callers both run spawn_blocking on cache miss

### LINKED ISSUE #2429: bug: load_skills_catalog has double-checked locking race — concurrent callers both run spawn_blocking on cache miss
## Bug

`load_skills_catalog` in `src/engine/router/llm.rs` (lines 674–725) uses a double-checked locking pattern without holding the lock across the expensive operation. Two concurrent routing calls can both see a cache miss and both spawn a blocking thread to read the skills catalog from disk.

## Code

```rust
async fn load_skills_catalog(&self) -> anyhow::Result<String> {
    // Check cache first (quick lock, drop immediately)
    {
        let cache = self.skills_catalog.lock().await;
        if let Some(ref catalog) = *cache {
            return Ok(catalog.clone());
        }
    }  // <-- lock dropped here

    // Both concurrent callers reach here simultaneously
    let catalog = tokio::task::spawn_blocking(move || { ... }).await?;

    let mut cache = self.skills_catalog.lock().await;
    *cache = Some(catalog.clone());  // second writer silently overwrites first
    Ok(catalog)
}
```

## Impact

1. **Redundant blocking I/O**: During burst routing periods (multiple tasks dispatched simultaneously), N concurrent routing calls all miss the empty cache and each spawn a `spawn_blocking` thread reading skills.yml and scanning the skills directory. The blocking thread pool is a shared resource — a burst of N concurrent callers means N unnecessary filesystem scans.

2. **Stale overwrite (minor)**: If caller A finishes `spawn_blocking` after caller B, A's catalog write overwrites B's. Since the catalog is read from static files this is usually harmless, but it is still an ordering inversion.

## Affected File

**`src/engine/router/llm.rs`, lines 674–725**

## Fix

Hold the lock for the entire operation to ensure only one caller loads the catalog:

```rust
async fn load_skills_catalog(&self) -> anyhow::Result<String> {
    let mut cache = self.skills_catalog.lock().await;
    if let Some(ref catalog) = *cache {
        return Ok(catalog.clone());
    }

    let skills_dir_opt = ...;
    // spawn_blocking while holding the Mutex is fine because the Mutex is
    // a tokio::sync::Mutex (async-aware) — it doesn't block a thread.
    let catalog = tokio::task::spawn_blocking(move || { ... }).await?;

    *cache = Some(catalog.clone());
    Ok(catalog)
}
```

Holding a `tokio::sync::Mutex` across `.await` is the intended use of async mutexes. This serializes concurrent callers so only the first one does the I/O work; subsequent callers wait and receive the cached result.

gabrielkoerich:
## Summary

Fix committed. Here's the summary:

- **Changed**: `load_skills_catalog` now holds `skills_catalog` lock across the entire operation (including `spawn_blocking`)
- **Effect**: Concurrent callers on a cold cache are serialized — only the first does the filesystem I/O; the rest wait and get the cached result
- **Why safe**: `tokio::sync::Mutex` doesn't block worker threads across `.await`, it only suspends the task — this is the idiomatic pattern for serializing async operations with shared st

### What was done

- **Changed**: `load_skills_catalog` now holds `skills_catalog` lock across the entire operation (including `spawn_blocking`)
- **Effect**: Concurrent callers on a cold cache are serialized — only the first does the filesystem I/O; the rest wait and get the cached result
- **Why safe**: `tokio::sync::Mutex` doesn't block worker threads across `.await`, it only suspends the task — this is the idiomatic pattern for serializing async operations with shared state

### Files changed

- `src/engine/router/llm.rs`

Closes #2429

---
*Task #2429 · Created by minimax[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/engine/router/llm.rs
+++ b/src/engine/router/llm.rs
@@ -672,16 +672,15 @@ impl LlmRouter {
     /// Load skills catalog from skills.yml or skills directory.
     /// Cached after first load to avoid blocking I/O in async context.
     async fn load_skills_catalog(&self) -> anyhow::Result<String> {
-        // Check cache first (quick lock, drop immediately)
-        {
-            let cache = self.skills_catalog.lock().await;
-            if let Some(ref catalog) = *cache {
-                return Ok(catalog.clone());
-            }
+        let mut cache = self.skills_catalog.lock().await;
+        if let Some(ref catalog) = *cache {
+            return Ok(catalog.clone());
         }
 
         // Offload uncached loading to blocking thread pool to avoid blocking the Tokio reactor.
         // Clone any data we need to avoid capturing &self across thread boundary.
+        // Holding the tokio::sync::Mutex across .await is safe — it doesn't block a worker thread,
+        // it just suspends the task. This ensures only one caller does the I/O work.
         let skills_dir_opt: Option<PathBuf> = match std::env::var("ORCH_HOME") {
             Ok(orch_home) => Some(PathBuf::from(orch_home).join("skills")),
             Err(_) => None,
@@ -718,7 +717,6 @@ impl LlmRouter {
         .await
         .map_err(|e| anyhow::anyhow!("spawn_blocking failed: {e}"))?;
 
-        let mut cache = self.skills_catalog.lock().await;
         *cache = Some(catalog.clone());
 
         Ok(catalog)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Correct fix for the double-checked locking race. The solution holds the tokio::sync::Mutex across the spawn_blocking await point, which is the idiomatic and safe pattern for async mutexes — they suspend the task rather than blocking a worker thread. All 2949 tests pass, cargo fmt and cargo clippy are clean. The branch rebased cleanly onto main. CI was still pending when reviewed but local checks are green.

---

gabrielkoerich:

---

PR #2491, opened 2026-04-12
bug: stuck-task recovery calls resolve_task_id twice per task — result not cached between uses

### LINKED ISSUE #2489: bug: stuck-task recovery calls resolve_task_id twice per task — result not cached between uses
## Bug

In two stuck-task recovery paths in `src/engine/tick.rs`, `store.resolve_task_id()` is called twice for the same task within the same loop iteration. The `store_id` obtained in the first call (for cooldown logic) is not reused for the second call (for routing state reset), resulting in a redundant DB round-trip per stuck task.

## Root cause

**External stuck-task path (`tick_recover_stuck_tasks`), lines 506 and 560:**

```rust
if timing.has_session {
    let store_task = match store.resolve_task_id(repo, &task.id.0).await {  // call 1
        Ok(Some(store_id)) => match store.get(store_id).await { ... }
        ...
    };
    // ... cooldown logic using store_task
}
// ... label cleanup ...
if let Ok(Some(store_id)) = store.resolve_task_id(repo, &task.id.0).await {  // call 2 — redundant
    store_set_by_id(..., store_id, ...).await;
}
```

**Internal stuck-task path (`tick_recover_internal_stuck_tasks`), lines 677 and 724:**
Same pattern — `resolve_task_id` called at line 677 inside `if timing.has_session` and again at line 724.

When `timing.has_session` is true, both calls execute sequentially for the same external task ID. The `store_id` from call 1 is accessible within that block but not reused outside it, forcing call 2 to redo the lookup.

## Impact

Two SQLite queries per stuck task per recovery tick instead of one. Mild under normal conditions (stuck tasks are rare), but wasteful and inconsistent — the second call can theoretically return a different result if the task was concurrently deleted between calls.

## Fix

Cache the `store_id` (or `Option<store_id>`) from the first resolve and reuse it for the routing state reset:

```rust
let mut cached_store_id: Option<i64> = None;

if timing.has_session {
    match store.resolve_task_id(repo, &task.id.0).await {
        Ok(Some(store_id)) => {
            cached_store_id = Some(store_id);
            match store.get(store_id).await { ... }  // cooldown logic
        }
        ...
    }
}

// Use cached value; fall back to a second resolve only if not yet set
let store_id = match cached_store_id {
    Some(id) => Some(id),
    None => store.resolve_task_id(repo, &task.id.0).await.ok().flatten(),
};
if let Some(store_id) = store_id {
    store_set_by_id(..., store_id, ...).await;
}
```

Apply the same pattern to the internal path at lines 677/724.

gabrielkoerich:
Resolves #2489

---
*Task #2489 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -502,15 +502,19 @@ pub(crate) async fn tick_recover_stuck_tasks(
         // router picks a different agent/model on the next attempt. Without this, the router
         // sees the agent as healthy and selects it again, causing the hang to repeat.
         // (Mirrors the pattern in tick_detect_silent_agents lines 219–284.)
+        let mut cached_store_id: Option<i64> = None;
         if timing.has_session {
             let store_task = match store.resolve_task_id(repo, &task.id.0).await {
-                Ok(Some(store_id)) => match store.get(store_id).await {
-                    Ok(t) => Some(t),
-                    Err(e) => {
-                        tracing::warn!(task_id = task.id.0, error = %e, "failed to fetch task from store for stuck-task cooldown");
-                        None
+                Ok(Some(store_id)) => {
+                    cached_store_id = Some(store_id);
+                    match store.get(store_id).await {
+                        Ok(t) => Some(t),
+                        Err(e) => {
+                            tracing::warn!(task_id = task.id.0, error = %e, "failed to fetch task from store for stuck-task cooldown");
+                            None
+                        }
                     }
-                },
+                }
                 Ok(None) => None,
                 Err(e) => {
                     tracing::warn!(task_id = task.id.0, error = %e, "failed to resolve task id for stuck-task cooldown");
@@ -557,7 +561,11 @@ pub(crate) async fn tick_recover_stuck_tasks(
                 }
             }
         }
-        if let Ok(Some(store_id)) = store.resolve_task_id(repo, &task.id.0).await {
+        let resolved_store_id = match cached_store_id {
+            Some(id) => Some(id),
+            None => store.resolve_task_id(repo, &task.id.0).await.ok().flatten(),
+        };
+        if let Some(store_id) = resolved_store_id {
             store_set_by_id(
                 &Some(Arc::clone(store)),
                 store_id,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

gabrielkoerich:

CI passed and the tick.rs changes correctly implement the intended caching of resolve_task_id to avoid a redundant DB round-trip. However, the patch introduced a correctness regression in src/engine/sync.rs: an unconditional expect() is used when looking up is_pr_map for CreateMentionTask, which can panic when the CreateMentionTask has issue_num == None (those indices are intentionally not present in is_pr_map). This will cause a runtime panic during mention processing. Fix: revert to the previous safe behavior (use .copied().unwrap_or(false)) or only call expect when you can guarantee the entry exists (i.e. only for CreateMentionTask entries where issue_num is Some). Apart from that, tick.rs changes look correct and follow the repo's patterns for logging and store updates.

**Issues Found:**
- `src/engine/sync.rs` line 1375: Unconditional expect() on is_pr_map lookup for CreateMentionTask. For CreateMentionTask entries with issue_num == None the index won't exist in is_pr_map, causing a panic. Restore safe behavior (e.g. use .copied().unwrap_or(false)) or only call expect when issue_num is Some. [error]


---

gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

CI passes (test, secrets). Code correctly caches store_id from the first resolve_task_id call and reuses it for the second call in both external and internal stuck-task paths. Pattern matches the issue description exactly. Note: branch needs rebase onto main (50 commits behind, 70+ conflicts) — orch will handle this before posting the review decision, then CI will re-run. Build, format, and clippy all clean.

---

gabrielkoerich:

---

PR #2521, opened 2026-04-12
bug: control.rs memory LIKE query uses unescaped session_id, leaking cross-session memories when name contains _ or %

### LINKED ISSUE #2518: bug: control.rs memory LIKE query uses unescaped session_id, leaking cross-session memories when name contains _ or %
## Problem

`assemble_context` in `src/control.rs:233-235` builds a LIKE pattern from `session_id` without escaping SQLite LIKE metacharacters:

```rust
let memory_prefix = format!("control:memory:{session_id}:%");
let memory_rows = sqlx::query("SELECT key, value FROM kv WHERE key LIKE ?")
    .bind(&memory_prefix)
```

SQLite LIKE treats `_` as "any single character" and `%` as "any string". If a session name contains `_` (e.g., `prod_db`, `ops_1`), the pattern matches keys from other sessions whose names differ by only one character at the `_` positions.

**Example:** session `prod_db` generates pattern `control:memory:prod_db:%`. This matches memories from session `prodXdb` (where X is any char) — unintended cross-session leakage.

## Fix

Replace the inline LIKE query with the already-correct `kv_list_prefix` method which properly escapes metacharacters:

```rust
// In src/store/kv.rs:60-63 — correct escaping already exists:
let escaped = prefix.replace('\\', "\\\\\\\\").replace('%', "\\\\%").replace('_', "\\_");
let pattern = format!("{escaped}%");
// ... LIKE ? ESCAPE '\\'
```

The fix in `src/control.rs` is to replace lines 232-254 with:

```rust
let memory_prefix = format!("control:memory:{session_id}:");
let raw_memories = store.kv_list_prefix(&memory_prefix).await.unwrap_or_else(|e| {
    tracing::warn!(session_id, error = %e, "failed to load control memories from store");
    Vec::new()
});
```

Then update the downstream mapping since `kv_list_prefix` returns `Vec<(String, String)>` instead of rows.

## References

- `src/control.rs:233-235` — unescaped LIKE query
- `src/store/kv.rs:57-71` — correct escaping pattern (already exists)
- Same issue was correctly handled in `kv_list_prefix` but the inline query in `assemble_context` was missed

gabrielkoerich:
Resolves #2518

### DIFF
--- a/src/control.rs
+++ b/src/control.rs
@@ -27,7 +27,6 @@ use std::sync::{Arc, LazyLock, Mutex};
 
 use anyhow::{Context, Result};
 use regex::Regex;
-use sqlx::Row;
 use tokio::time::{timeout, Duration};
 use uuid::Uuid;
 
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/minimax-m2.5-free`*

gabrielkoerich:
## Automated Review — Approve

CI passes (test, secrets). Code correctly replaces unescaped LIKE query with kv_list_prefix which properly escapes LIKE metacharacters (% and _) to prevent cross-session memory leakage. Local checks pass (fmt, clippy).

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/minimax-m2.5-free`*

gabrielkoerich:

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/minimax-m2.5-free`*

Issue #2318, opened 2026-04-09
bug: claude agents running 12-30 minutes then killed as "silent" — grace period insufficient for complex tasks


gabrielkoerich:
## What you found

5 claude agent runs in 48 hours were killed by silence detection after running 12-30 minutes. These are complex tasks (refactoring, deep fixes) that produce minimal tmux pane output but are actively working.

**Evidence — claude silence detections in last 48h:**
```
86856|claude|sonnet|1805s → set task to new (hard timeout was ~1s away)
88498|claude|sonnet|997s → set task to new
88364|claude|sonnet|830s → set task to new
93067|claude|sonnet|1267s → set task to new
95214|claude|sonnet|774s → set task to new
```

Compare with opencode: ALL opencode sessions die at 600-611s regardless of task. But claude tasks run longer (774-1805s) before being killed — they're actually doing work but producing no tmux pane output.

## Why it's happening

The silence detection in `src/channels/capture.rs:184` checks `has_output == false`. For claude agents:
- `has_output` is only set when `diff_and_update` captures non-empty content (line 328)
- If the agent writes mostly to files (Edit, Write tools) with minimal terminal output, the tmux pane may show almost nothing
- Long-running tasks (refactoring, multi-file changes) may run entirely without terminal output

The 1805s case is particularly telling: the agent almost hit the hard 1800s timeout and would have been caught by the timeout handler instead. The silence detection killed it 5 seconds before the hard timeout would have fired naturally.

## Why it matters

- **5 failures in 48h** — real work was in progress (agent was running code, not stuck)
- **Complex tasks killed prematurely** — these require deep analysis, multi-file changes, and naturally produce minimal terminal output
- **No fallback to timeout path** — silence detection bypasses the graceful timeout handling
- **Waste of prior agent work** — if the agent was 80% done with changes, they're lost when task is reset to `new`

## Suggested fix

1. **Distinguish \"never started\" from \"started but quiet\"**: Check not just `has_output` but also whether the session has been running long enough that it likely started. A session that's 774s old with no output is more likely to be stuck than one that's 120s old.

2. **Increase grace period for claude agents doing complex work**: 120s is fine for quick tasks, but complex refactoring can take 15+ minutes without terminal output. Consider:
   - Per-agent grace periods (claude: 1800s, opencode: 300s)
   - Or base on task complexity estimate from routing

3. **Add output threshold**: Instead of requiring ANY output, require a minimum output threshold. Agents doing real work will eventually output something.

[...]

PR #2582, opened 2026-04-13
fix(capture): silence detection no longer bypassed for dead seen-alive sessions

### LINKED ISSUE #2573: bug: silence detection bypassed when agent exits with 0 but tmux session was seen alive
## Summary

When an agent (typically opencode with free models) exits silently with code 0 and produces no output, silence detection is bypassed because `seen_alive == true`. The task then waits the full `stuck_timeout` (30 minutes) before recovery kicks in.

This is a regression introduced in `3deb3df1` (issue #2318), which added a `seen_alive` guard to prevent killing long-running Claude agents doing file-heavy work with minimal terminal output. The fix was correct for that case, but created a blind spot for sessions that start successfully and then exit immediately with no output.

## Root Cause

In `src/channels/capture.rs:199-201`, `get_silent_sessions_for_repo()` skips any session where `seen_alive == true`:

```rust
if buf.seen_alive {
    continue;
}
```

The flow for a silent exit-0:
1. Session spawned, `capture_pane` succeeds on first tick (even with empty pane) → `seen_alive = true` (line 289)
2. Agent exits silently with code 0, producing no output
3. Capture service detects dead session, sends `is_final: true`, unregisters (line 253-267) — but this only feeds the transport/streaming layer
4. Silence detection skips the session because `seen_alive == true`
5. Task sits idle for **30 minutes** until `tick_recover_stuck_tasks()` catches it via `stuck_timeout: 1800`

## What existed before

Before `3deb3df1`, silence detection was: `has_output == false` + past `grace_period` → silent. No `seen_alive` exemption. This caught silent exit-0 cases within the 120s grace period.

## The gap

The capture service already detects dead sessions via `tmux::is_session_dead()` (line 253) and sends `is_final` chunks — but this information never feeds back into the engine's task lifecycle. It only serves the CLI streaming layer.

Two distinct cases are conflated by the `seen_alive` flag:
- **Session alive and working** (tmux session exists, agent is running, just no terminal output) — should NOT be killed
- **Session was alive but exited** (tmux session is dead, `has_output == false`) — SHOULD trigger silence fallback

## Proposed Fix

In `get_silent_sessions_for_repo()`, when `seen_alive == true && has_output == false`, check if the tmux session is actually still alive. If the session is dead, it's a silent exit — include it in the silent list:

```rust
if buf.seen_alive {
    // Session was confirmed alive, but if it's now dead with no output,
    // that's a silent exit — don't wait for stuck_timeout.
    if !buf.has_output && tmux::is_session_dead(&buf.session).await {
        // fall through to grace period check
    } else {
        continue;
    }
}
```

This preserves the #2318 fix (don't kill long-running agents with sparse output) while catching silent exit-0 within the grace period instead of waiting 30 minutes.

Alternative: when the capture tick detects a dead session with `has_output == false`, proactively notify the engine (e.g., via a channel or flag) instead of only sending `is_final` to transport.

## Observed in

Issue #2567 comment — task waited ~30 minutes before "silent exit 0, retrying with free model opencode/minimax-m2.5-free" triggered via stuck task recovery instead of silence detection.

## References

- `3deb3df1` — commit that added `seen_alive` guard (fix for #2318)
- `src/channels/capture.rs:164-208` — `get_silent_sessions_for_repo()`
- `src/channels/capture.rs:253` — dead session detection (only feeds transport)
- `src/engine/tick.rs:202` — `tick_detect_silent_agents()`
- `src/engine/tick.rs:473` — `tick_recover_stuck_tasks()` (30-min fallback)

gabrielkoerich:
## Summary

- When an agent exits silently (code 0, no output), `get_silent_sessions_for_repo` was unconditionally skipping any session where `seen_alive == true`, causing a 30-minute wait for `stuck_timeout` instead of the 120s grace period
- Fix: when `seen_alive=true` and `has_output=false`, check `tmux::is_session_dead()` — if the session is dead it's a silent exit and falls through to the grace-period check; if still alive it's skipped (preserving the #2318 fix for long-running agents with sparse output)
- Added `set_buffer_state_for_test` helper on `CaptureService` and updated integration test mock (`is_session_dead` returns `true` for session names ending in `-dead`) to cover both the silent-exit and alive-but-quiet cases

## Test plan

- [x] `cargo nextest run` — 2976/2976 pass
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] New test `get_silent_sessions_seen_alive_dead_session_is_silent` in `capture.rs` asserts both never-seen and dead-seen-alive sessions are silenced
- [x] New test `seen_alive_with_live_session_not_silenced` in `tests/capture_diff.rs` asserts the #2318 regression (alive session is never silenced) still holds

Closes #2573

---
*Task #2573 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/channels/capture.rs
+++ b/src/channels/capture.rs
@@ -37,9 +37,11 @@ pub struct OutputBuffer {
     pub last_hash: Option<[u8; 32]>,
     /// When the last capture occurred
     pub last_capture: DateTime<Utc>,
-    /// Whether the session has been seen alive at least once.
-    /// Prevents firing "session ended" for sessions that were registered
-    /// before the tmux session was actually created.
+    /// Whether the session has been seen alive at least once (i.e., `capture_pane`
+    /// succeeded at least once). Prevents firing "session ended" for sessions that
+    /// were registered before the tmux session was actually created.
+    /// Also used by silence detection: a seen-alive session with no output is only
+    /// treated as silent if the tmux session is currently dead (silent exit-0).
     pub seen_alive: bool,
     /// When this session was registered for capture.
     pub registered_at: DateTime<Utc>,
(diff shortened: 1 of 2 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

CI passes (secrets + test). Branch up to date. The fix correctly resolves the conflict between #2318 (don't kill long-running agents with sparse output) and #2573 (silent exit-0 bypasses silence detection): when `seen_alive=true && has_output=false`, call `is_session_dead()` — only fall through to silence detection if the tmux session is actually dead. Read lock is dropped before the async call, avoiding lock-during-await. Test helper + rename improve clarity without scope creep.

---

gabrielkoerich:

---

PR #2580, opened 2026-04-13
bug: sync_estimate_to_project blocks tick loop inline during routing

gabrielkoerich:
## Summary

Moved `sync_estimate_to_project` call in `tick_route_tasks` (tick.rs:1082) from inline `await` to `tokio::spawn` fire-and-forget. The tick loop no longer blocks on GraphQL mutations during routing. `Arc::clone(backend)` is passed into the spawned task. The startup reconciliation was already backgrounded by #2577, and the store query already filters terminal tasks.

### Files changed

- `src/engine/tick.rs`

Closes #2579

---
*Task #2579 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

PR #2578, opened 2026-04-13
bug: engine tick loop stalls for 12+ minutes — all tokio workers parked

gabrielkoerich:
Resolves #2574

Auto-created by orch review gate (agent forgot to open PR)

---
*Task #2574 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

PR #2583, opened 2026-04-13
bug: issues created during engine downtime permanently skipped by incremental ingest

gabrielkoerich:
Resolves #2581

Auto-created by orch review gate (agent forgot to open PR)

PR #2589, opened 2026-04-13
fix(jobs): add timeout to bash job type to prevent tick loop blocking

### LINKED ISSUE #2584: bug: bash job type blocks engine tick loop indefinitely — no timeout
## Root Cause

`src/engine/jobs.rs:481-486` — The `bash` job type in `execute_job` runs `bash -c {cmd}` directly with no timeout:

```rust
let output = tokio::process::Command::new("bash")
    .arg("-c")
    .arg(cmd)
    .current_dir(dir)
    .output_with_context()
    .await;
```

This call chains inline through the entire tick loop:

```
engine main loop
 └── tick() inline          [tick.rs:1587]
      └── tick_job_scheduler() inline  [tick.rs:1627]
           └── jobs::tick() inline    [tick.rs:1574]
                └── execute_job() inline  [jobs.rs:383]
                     └── bash -c {cmd}   ← no timeout
```

A bash command that hangs — due to a network operation, a blocking subprocess, or any other reason — parks a Tokio worker for the entire duration. Because the tick loop runs inline (not spawned), all other phases (routing, dispatch, stuck-task recovery, PR sync) are blocked for every project until the command returns.

This is the same class of bug fixed in #2578 (sync_estimate_to_project) and #2580 (sync_tick inline).

## Impact

- All task routing, dispatch, and recovery for **all** projects stalls while the bash job runs
- No watchdog can help — the 6× tick-interval watchdog detects the stall but cannot interrupt it (#2578)
- Stuck tasks can pile up during the stall window
- If the job file (`.orch.yml`) is configured to run a bash command that performs network I/O, this will fire on every cron interval

## Fix

Wrap the `bash` command with `tokio::time::timeout` using a configurable duration (e.g., default 5 minutes, configurable via `jobs.bash_timeout_seconds`):

```rust
"bash" => {
    if let Some(ref cmd) = job.command {
        let dir = job.dir.as_deref().unwrap_or(".");
        let timeout = std::time::Duration::from_secs(300); // 5 min default
        match tokio::time::timeout(
            timeout,
            tokio::process::Command::new("bash")
                .arg("-c")
                .arg(cmd)
                .current_dir(dir)
                .output_with_context(),
        )
        .await
        {
            Ok(Ok(o)) if o.status.success() => { /* done */ }
            Ok(Ok(o)) => { /* command failed */ }
            Ok(Err(e)) => { /* spawn error */ }
            Err(_) => {
                tracing::error!(job_id = job.id, "bash job timed out after {}s", timeout.as_secs());
                state.last_task_status = Some("timeout".to_string());
            }
        }
    }
}
```

Alternatively, spawn the bash job as a background task and track its handle — but the simpler timeout approach is consistent with how skill sync handles `git pull/clone` (`src/engine/sync.rs:1486-1549`).

gabrielkoerich:
## Summary

- Wrap `bash` job type execution in `tokio::time::timeout` to prevent an unbounded block on the inline tick loop
- A bash command that hangs (network operation, blocking subprocess) would otherwise park a Tokio worker for its entire duration, blocking all other tick phases (routing, dispatch, recovery, PR sync) for every project
- This is the same class of bug fixed in #2578 (sync_estimate_to_project) and #2580 (sync_tick inline)
- Default timeout: 5 minutes, configurable via `jobs.bash_timeout_seconds` in config

## Root Cause

`src/engine/jobs.rs:481-486` ran `bash -c {cmd}` directly with no timeout. Because `jobs::tick()` is called inline from the engine main loop (`tick.rs:1627`), a hung bash job blocks the entire tick loop.

## Fix

```rust
let timeout_secs: u64 = crate::config::get("jobs.bash_timeout_seconds")
    .ok()
    .and_then(|s| s.parse().ok())
    .unwrap_or(300);
let timeout = Duration::from_secs(timeout_secs);

let output = tokio::time::timeout(
    timeout,
    tokio::process::Command::new("bash").arg("-c").arg(cmd)...
)
.await;
```

On timeout, log an error and set `last_task_status = "timeout"`.

## Test plan

- [x] `cargo fmt -- --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes  
- [x] `cargo nextest run` — 2976 tests pass

---
*Task #2584 · Created by minimax[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/engine/jobs.rs
+++ b/src/engine/jobs.rs
@@ -478,18 +478,30 @@ pub async fn execute_job(
                 let dir = job.dir.as_deref().unwrap_or(".");
                 tracing::info!(job_id = job.id, cmd, dir, "running bash command");
 
-                let output = tokio::process::Command::new("bash")
-                    .arg("-c")
-                    .arg(cmd)
-                    .current_dir(dir)
-                    .output_with_context()
-                    .await;
+                // Configurable timeout with 5-minute default. A bash job that hangs
+                // (network operation, blocking subprocess) would otherwise park a Tokio
+                // worker for its entire duration, blocking the inline tick loop.
+                let timeout_secs: u64 = crate::config::get("jobs.bash_timeout_seconds")
+                    .ok()
+                    .and_then(|s| s.parse().ok())
+                    .unwrap_or(300);
+                let timeout = std::time::Duration::from_secs(timeout_secs);
+
+                let output = tokio::time::timeout(
+                    timeout,
+                    tokio::process::Command::new("bash")
+                        .arg("-c")
+                        .arg(cmd)
+                        .current_dir(dir)
+                        .output_with_context(),
+                )
+                .await;
 
                 match output {
-                    Ok(o) if o.status.success() => {
+                    Ok(Ok(o)) if o.status.success() => {
                         state.last_task_status = Some("done".to_string());
                     }
-                    Ok(o) => {
+                    Ok(Ok(o)) => {
                         let stderr = String::from_utf8_lossy(&o.stderr);
                         tracing::warn!(
                             job_id = job.id,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Branch rebased onto main (new commit 58bbd5f1 from #2583 already applied the mod.rs/sync.rs removals, so the rebased PR commit is now jobs.rs-only). CI results are stale due to rebase; will re-run after push.

The fix is correct and minimal: wraps bash execution in tokio::time::timeout with a configurable duration (default 300s via jobs.bash_timeout_seconds), uses the same crate::config::get pattern as other numeric config reads (e.g. webhook.dedup_window_seconds), handles all four match arms properly, and sets last_task_status = 'timeout' to unblock the tick loop. No scope creep, no architecture violations.

---

gabrielkoerich:

---

PR #2678, opened 2026-04-15
bug: kv_increment .max(1) is unreachable dead code and contradicts function contract

### LINKED ISSUE #2640: bug: kv_increment .max(1) is unreachable dead code and contradicts function contract
## Summary

`TaskStore::kv_increment` in `src/store/kv.rs:41` applies `.max(1)` to the SQL `RETURNING` value. This guard is unreachable under any normal code path and its presence implies the SQL contract is weaker than it actually is.

## Location

`src/store/kv.rs`, line 41:
```rust
Ok(row.0.max(1) as u64)
```

## Why it's unreachable

The SQL is:
```sql
INSERT INTO kv (key, value, ...) VALUES (?1, '1', ...)
ON CONFLICT(key) DO UPDATE
  SET value = CAST(value AS INTEGER) + 1, ...
RETURNING CAST(value AS INTEGER)
```

- **Insert path**: value = '1' → RETURNING = 1. Always ≥ 1.
- **Update path**: value = CAST(prev AS INTEGER) + 1. The only path that sets a KV counter to a non-positive integer is `record_agent_success` which calls `kv_set(key, "0")`. After that reset, the next increment returns CAST('0' AS INTEGER) + 1 = 1. Always ≥ 1.

The guard would only fire if value stored in the DB were "-1" or lower (i.e., CAST + 1 ≤ 0). No code path does this, so the guard is dead code.

## Problem

1. **Misleading contract**: The guard signals to readers that the SQL might return 0, implying uncertainty about the invariant. Future callers may add their own `.max(1)` defensively, compounding the confusion.
2. **Masks corruption**: If a counter somehow reached "-1" (external manipulation, future bug), `.max(1)` silently returns 1, making the corruption look like a clean "first failure". The correct behaviour is to return the actual value (or return an error) so callers can detect the anomaly.
3. **Contract mismatch**: The doc comment says "Returns the new value after the increment" without caveating that values below 1 are clamped. Callers like `read_and_increment_failure_count_with_prefix` (cooldown.rs:226) use the return value directly for backoff computation; clamping to 1 would cause them to use the base backoff instead of the escalated one.

## Fix

Remove the `.max(1)` guard:
```rust
// Before
Ok(row.0.max(1) as u64)

// After
Ok(row.0 as u64)
```

If a defensive guard is desired, use an assertion that panics loudly in debug builds instead of silently clamping:
```rust
debug_assert!(row.0 >= 1, "kv_increment returned unexpected value {}", row.0);
Ok(row.0 as u64)
```

gabrielkoerich:
## Summary

Removed the unreachable .max(1) clamp from kv_increment and added a debug assertion to surface unexpected values.

### What was done

- Inspected src/store/kv.rs and confirmed kv_increment used row.0.max(1) which silently clamped unexpected values.
- Replaced the .max(1) clamp with a debug_assert!(row.0 >= 1, ...) and returned the actual value as u64.
- Ran library tests (cargo test --lib) and a targeted cooldown test; test suite ran and failures shown earlier were unrelated to this change. Committed the change with a concise commit message.

### Remaining

- No follow-up code changes required for this task. Consider monitoring for any runtime assertions in debug builds if a database corruption or unexpected negative values occur.

### Files changed

- `src/store/kv.rs`

Closes #2640

---
*Task #2640 · Created by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `github-copilot/gpt-5-mini`*

### DIFF
--- a/src/store/kv.rs
+++ b/src/store/kv.rs
@@ -43,7 +43,11 @@ impl TaskStore {
         // which silently masked database corruption or unexpected negative values.
         // Use a debug assertion to loudly detect unexpected values in debug builds
         // while returning the actual stored value in release builds.
-        debug_assert!(row.0 >= 1, "kv_increment returned unexpected value {}", row.0);
+        debug_assert!(
+            row.0 >= 1,
+            "kv_increment returned unexpected value {}",
+            row.0
+        );
         Ok(row.0 as u64)
     }
 
--- a/src/store/kv.rs
+++ b/src/store/kv.rs
@@ -38,7 +38,13 @@ impl TaskStore {
         .bind(key)
         .fetch_one(&self.pool)
         .await?;
-        Ok(row.0.max(1) as u64)
+        // The SQL always returns a value >= 1 (insert path returns 1, update adds 1 to
+        // the previous integer value). Previously this was clamped with `.max(1)`,
+        // which silently masked database corruption or unexpected negative values.
+        // Use a debug assertion to loudly detect unexpected values in debug builds
+        // while returning the actual stored value in release builds.
+        debug_assert!(row.0 >= 1, "kv_increment returned unexpected value {}", row.0);
+        Ok(row.0 as u64)
     }
 
     /// Delete a key from the KV store.


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

The fix is correct and minimal. Removed the unreachable `.max(1)` clamp and added a `debug_assert!` as recommended in the task description. Code follows existing patterns. A `cargo fmt` issue was auto-fixed (debug_assert multiline formatting) and committed. All 3048 tests pass, clippy clean.

---

gabrielkoerich:

---

PR #2721, opened 2026-04-16
router: holding skills_catalog Mutex across spawn_blocking await

### LINKED ISSUE #2714: router: holding skills_catalog Mutex across spawn_blocking await
src/engine/router/llm.rs lines ~680-736

Problem:
The router locks self.skills_catalog via self.skills_catalog.lock().await and then calls tokio::task::spawn_blocking(...).await while the lock is still held. This means the async Mutex is held across an await point that itself blocks on a blocking thread; other callers trying to read the skills_catalog will be serialized until the blocking operation completes. The file comment incorrectly states that holding the tokio::sync::Mutex across .await is safe — it is not safe from a concurrency/throughput perspective.

Why it's a root cause:
This serializes potentially long-running I/O (scanning skills directory or reading files) and can lead to contention and reduced parallelism. It also increases the risk of deadlocks if callers try to acquire other locks in different orders.

Suggested fix:
Limit the scope of the MutexGuard. Acquire the lock, check if cached value exists, clone it if present and return; if not present, drop the guard before spawning blocking I/O and then re-acquire to set the cache: { let mut cache_guard = self.skills_catalog.lock().await; if let Some(c) = &*cache_guard { return Ok(c.clone()); } drop(cache_guard); let catalog = tokio::task::spawn_blocking(...).await?; let mut cache_guard = self.skills_catalog.lock().await; *cache_guard = Some(catalog.clone()); Ok(catalog) }

Risk/priority: Medium-High — causes unnecessary serialization of router calls and can slow task routing when skills syncs or scans occur.

---
_Reopened by #2705 — original issue was prematurely closed by an agent bug (tasks marked done without completing work)._

gabrielkoerich:
## Summary

The fix is complete. Here's a summary:

## Fix Applied: Router Skills Catalog Mutex Scope Limitation

**Problem**: `load_skills_catalog()` was holding the `skills_catalog` Mutex across the `spawn_blocking().await` call, serializing all router calls when the catalog needed to be loaded from disk.

**Solution**: Limited the MutexGuard scope by:
1. Acquiring lock only briefly to check the cache (drops immediately)
2. If cache miss, performing blocking I/O WITHOUT holding any lock
3. Re-acquiring lo

### What was done

- Problem**: `load_skills_catalog()` was holding the `skills_catalog` Mutex across the `spawn_blocking().await` call, serializing all router calls when the catalog needed to be loaded from disk.
- Acquiring lock only briefly to check the cache (drops immediately)
- If cache miss, performing blocking I/O WITHOUT holding any lock
- Re-acquiring lock only to update the cache after I/O completes
- Changes**: `src/engine/router/llm.rs` - 11 insertions, 8 deletions
- ✅ `cargo fmt -- --check` passes
- ✅ `cargo clippy --all-targets -- -D warnings` passes
- ✅ All 112 llm-specific tests pass
- ✅ All 144 router tests pass
- Note**: The failing test `engine::tests::configured_agents_fallback_returns_default_agents` is a pre-existing issue (fails before my changes too) and unrelated to this fix. It should be addressed in a separate issue.

### Files changed

- `src/engine/router/llm.rs`

Closes #2714

---
*Task #2714 · Created by glm[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/engine/router/llm.rs
+++ b/src/engine/router/llm.rs
@@ -685,15 +685,16 @@ impl LlmRouter {
     /// Load skills catalog from skills.yml or skills directory.
     /// Cached after first load to avoid blocking I/O in async context.
     async fn load_skills_catalog(&self) -> anyhow::Result<String> {
-        let mut cache = self.skills_catalog.lock().await;
-        if let Some(ref catalog) = *cache {
-            return Ok(catalog.clone());
-        }
+        // First, check the cache with a short-lived lock
+        {
+            let cache = self.skills_catalog.lock().await;
+            if let Some(ref catalog) = *cache {
+                return Ok(catalog.clone());
+            }
+        } // Lock is dropped here
 
-        // Offload uncached loading to blocking thread pool to avoid blocking the Tokio reactor.
-        // Clone any data we need to avoid capturing &self across thread boundary.
-        // Holding the tokio::sync::Mutex across .await is safe — it doesn't block a worker thread,
-        // it just suspends the task. This ensures only one caller does the I/O work.
+        // Cache miss — offload loading to blocking thread pool.
+        // Capture only what we need to avoid holding any locks across the await.
         let skills_dir_opt: Option<PathBuf> = match std::env::var("ORCH_HOME") {
             Ok(orch_home) => Some(PathBuf::from(orch_home).join("skills")),
             Err(_) => None,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Clean, focused fix. Limits MutexGuard scope to: (1) brief cache-check lock, (2) blocking I/O without lock, (3) brief cache-write lock. Removes incorrect comment. cargo fmt ✓, clippy ✓, CI required checks pass ✓. One pre-existing test failure (configured_agents_fallback_returns_default_agents) also fails on origin/main due to local config — unrelated to this PR.

---

gabrielkoerich:

---
*Reviewed by glm[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

PR #2741, opened 2026-04-17
fix: propagate KV read errors in get_model/get_agent instead of silently defaulting

### LINKED ISSUE #2740: bug: control session silently resets to defaults when KV reads fail
## Summary
Control session model/agent selection silently falls back to defaults when KV reads fail.

## Location
- `src/control.rs:323-339`
  - `get_model()` uses `kv_get("control:model").await.ok().flatten().unwrap_or_else(...)`
  - `get_agent()` uses `kv_get("control:agent").await.ok().flatten().unwrap_or_else(...)`

## What is wrong
Any KV/store error is treated as if the key were missing, so control sessions quietly switch to `claude:sonnet` defaults during DB faults. That hides the underlying failure and can cause unexpected model/agent changes mid-operations.

## Expected behavior
Store read errors should be observable and distinct from an actually missing key.

## Proposed fix
- Change `get_model`/`get_agent` to return `anyhow::Result<String>`.
- Preserve default fallback only for the explicit `Ok(None)` path.
- On `Err`, either propagate to caller (preferred) or log at `warn/error` and keep previous in-memory value.

## Why this matters
Silent fallback changes execution behavior at exactly the time reliability is degraded, which makes incident diagnosis harder and can route control actions to the wrong agent/model.

gabrielkoerich:
Fixes #2740.

## Summary

- `get_model()` and `get_agent()` now return `anyhow::Result<String>` instead of `String`
- Store errors are propagated with context instead of being swallowed by `.ok()`
- Default fallback (`claude:sonnet`) is preserved only for the `Ok(None)` case (key genuinely absent)
- All callers updated to propagate the error with `?`
- Tests updated to `.unwrap()` the result

## Before

```rust
store.kv_get("control:model").await.ok().flatten().unwrap_or_else(|| "sonnet".to_string())
```

Any DB error was treated identically to a missing key, silently switching the control session to `claude:sonnet`.

## After

```rust
store.kv_get("control:model").await.context("reading control:model from store")?.unwrap_or_else(|| "sonnet".to_string())
```

Store errors surface to the caller immediately, making DB faults visible at the call site rather than masking them as a model switch.

## Test plan

- [x] `cargo fmt && cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo nextest run` — 3101 passed, 19 skipped

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

---
*Task #2740 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/cli/chat.rs
+++ b/src/cli/chat.rs
@@ -7,8 +7,8 @@ use std::io::{self, BufRead, Write};
 /// Interactive REPL mode — reads from stdin, sends to control session.
 pub async fn interactive(session_id: &str) -> anyhow::Result<()> {
     let store = crate::cli::init_store().await?;
-    let model = control::get_model(&store).await;
-    let agent = control::get_agent(&store).await;
+    let model = control::get_model(&store).await?;
+    let agent = control::get_agent(&store).await?;
 
     let session_label = if session_id == TaskStore::DEFAULT_SESSION {
         String::new()
@@ -161,8 +161,8 @@ pub async fn stats(session_id: &str) -> anyhow::Result<()> {
     let store = crate::cli::init_store().await?;
 
     let summary = store.get_session_cost_summary(session_id).await?;
-    let fallback_model = control::get_model(&store).await;
-    let fallback_agent = control::get_agent(&store).await;
+    let fallback_model = control::get_model(&store).await?;
+    let fallback_agent = control::get_agent(&store).await?;
     let current_model = summary.primary_model.clone().unwrap_or(fallback_model);
     let current_agent = summary.primary_agent.clone().unwrap_or(fallback_agent);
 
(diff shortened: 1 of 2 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Reviewed full diff for PR #2741 after Step 1-4 flow. Rebase attempt was a no-op (branch already up to date) with expected sandbox lockfile errors and clean git status afterward. Required CI checks are green (`test`, `secrets`). Local fallback checks from workflow also pass: `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, SELECT-* guard, and `cargo llvm-cov nextest --lcov --output-path lcov.info` (3101 passed, 19 skipped). Change is aligned with task intent and architecture: KV read errors are now propagated instead of silently defaulting, while `Ok(None)` still preserves defaults.

---

gabrielkoerich:

---

PR #2795, opened 2026-04-18
bug: try_get("task_id").unwrap_or(0) logs activity against task 0 on column decode failure

### LINKED ISSUE #2791: bug: try_get("task_id").unwrap_or(0) logs activity against task 0 on column decode failure
## Summary

In `src/store/tasks.rs:1689`, when `complete_run` reads back the run context row it uses `try_get("task_id").unwrap_or(0)` for the task ID. If the column fails to decode (e.g., after a migration that changes the column type, or a corrupt row), the fallback value is `0`. All subsequent `append_activity` calls at lines 1694–1722 then write audit trail entries against task ID `0`, silently corrupting the activity log.

The same row also defaults `run_type`, `agent`, and `model` to empty strings via `unwrap_or_default()` (lines 1690–1692), which would produce low-quality activity entries even if task_id happened to be right.

## Location

- `src/store/tasks.rs:1689–1692`

## What's wrong

```rust
let task_id: i64 = run_ctx.try_get("task_id").unwrap_or(0); // task 0 is wrong
let run_type: String = run_ctx.try_get("run_type").unwrap_or_default();
let agent: String = run_ctx.try_get("agent").unwrap_or_default();
let model: String = run_ctx.try_get("model").unwrap_or_default();
```

If `try_get("task_id")` fails (not just missing row — the `None` case is handled at line 1724 — but a decode error on the actual row), the function proceeds to log activity against `task_id = 0`.

## Fix

Propagate the decode error instead of defaulting:

```rust
let task_id: i64 = run_ctx.try_get("task_id")
    .map_err(|e| anyhow::anyhow!("failed to decode task_id from run_id {}: {e}", run.run_id))?;
let run_type: String = run_ctx.try_get("run_type")
    .map_err(|e| anyhow::anyhow!("failed to decode run_type from run_id {}: {e}", run.run_id))?;
let agent: String = run_ctx.try_get("agent")
    .map_err(|e| anyhow::anyhow!("failed to decode agent from run_id {}: {e}", run.run_id))?;
let model: String = run_ctx.try_get("model")
    .map_err(|e| anyhow::anyhow!("failed to decode model from run_id {}: {e}", run.run_id))?;
```

This makes decode failures visible (the caller already propagates errors via `?`) instead of silently producing garbage data in the activity log.

gabrielkoerich:
## Summary

Done. Fixed `src/store/tasks.rs:1689–1692` — replaced `unwrap_or(0)` and `unwrap_or_default()` with `map_err` propagation so decode failures surface as errors instead of silently logging activity against task ID 0. All 3183 tests pass, linter and formatter clean. Issue #2791 closed.

```json
{"success": true, "issue": "gh-issue-2791-bug-try-get-task-id-unwrap-or-0-logs-act", "commit": "fix(store): propagate decode errors instead of defaulting in complete_run"}
```

### Files changed

- `src/store/tasks.rs`

Closes #2791

---
*Task #2791 · Created by minimax[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/store/tasks.rs
+++ b/src/store/tasks.rs
@@ -1686,10 +1686,18 @@ impl TaskStore {
         .execute(&self.pool)
         .await?;
         if let Some(run_ctx) = run_ctx {
-            let task_id: i64 = run_ctx.try_get("task_id").unwrap_or(0);
-            let run_type: String = run_ctx.try_get("run_type").unwrap_or_default();
-            let agent: String = run_ctx.try_get("agent").unwrap_or_default();
-            let model: String = run_ctx.try_get("model").unwrap_or_default();
+            let task_id: i64 = run_ctx.try_get("task_id").map_err(|e| {
+                anyhow::anyhow!("failed to decode task_id from run_id {}: {e}", run.run_id)
+            })?;
+            let run_type: String = run_ctx.try_get("run_type").map_err(|e| {
+                anyhow::anyhow!("failed to decode run_type from run_id {}: {e}", run.run_id)
+            })?;
+            let agent: String = run_ctx.try_get("agent").map_err(|e| {
+                anyhow::anyhow!("failed to decode agent from run_id {}: {e}", run.run_id)
+            })?;
+            let model: String = run_ctx.try_get("model").map_err(|e| {
+                anyhow::anyhow!("failed to decode model from run_id {}: {e}", run.run_id)
+            })?;
             if run.outcome == "timeout" {
                 self.append_activity(
                     task_id,


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3-super-free`*

gabrielkoerich:

The PR introduces a potential issue: the error messages in the map_err calls reference `run.run_id` but the variable `run` may not be in scope at that point. Looking at the diff, the `run` variable is defined earlier in the function (outside the shown snippet) but we need to verify. However, the more critical issue is that the PR does not actually fix the problem described in the issue. The issue states that the problem occurs when `try_get` fails to decode (returns an Err), but the current code in the PR still uses `try_get` which returns a Result, and then uses `map_err` to convert the Err to an Err with a custom message, and then uses `?` to propagate. This is correct for handling decode errors. However, we must check if the variable `run` is accessible in the scope where these lines are placed. If not, the code will not compile. Additionally, we should ensure that the error handling is consistent with the rest of the function. Let's examine the surrounding code to confirm.

**Issues Found:**
- `src/store/tasks.rs` line 1689: The variable `run` may not be in scope at the point where the error messages are constructed. We need to check the surrounding code to ensure `run` is defined. If not, we should use a different identifier or adjust the scope. [error]


---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3-super-free`*

gabrielkoerich:
The concern about `run` not being in scope is unfounded. The `run` parameter is declared at the function signature:

```rust
pub async fn complete_run(&self, run: &CompleteRun<'_>) -> anyhow::Result<()>
```

It is a function parameter and is therefore in scope throughout the entire function body, including inside the `if let Some(run_ctx)` block at line 1688 where the `map_err` closures reference it. The code compiles cleanly and all 3183 tests pass (clippy is also clean). No changes are needed.

gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Required CI checks were verified with `gh pr checks 2795 --watch --fail-fast --required` and passed (`test`, `secrets`). Optional rebase hit expected worktree lockfile permission errors but reported the branch already up to date, with a clean `git status` before/after, so this was treated as non-blocking per instructions. Full diff review of `src/store/tasks.rs` confirms the fix correctly removes silent fallbacks (`unwrap_or(0)` / `unwrap_or_default()`) and now propagates decode errors with contextual `run_id`, preventing activity logging against task `0` and avoiding silent data-quality corruption. Change is in scope, aligns with documented architecture constraints, and introduces no migration/config violations.

---

gabrielkoerich:

---

PR #2820, opened 2026-04-19
bug: extract_json_field silently returns None for quoted string values — broken quoted-string parser in fallback.rs

### LINKED ISSUE #2817: bug: extract_json_field silently returns None for quoted string values — broken quoted-string parser in fallback.rs
## Problem

**`src/engine/runner/fallback.rs:46–70`** — the `extract_json_field` function has a broken quoted-string path:

```rust
// Handle quoted strings
if value_str.starts_with('"') {
    value_str
        .strip_prefix('"')
        .and_then(|s| s.strip_suffix('"'))  // BUG: strips LAST '"' from remaining JSON, not field closing '"'
        .and_then(|s| s.strip_suffix(','))
        .map(|s| s.to_string())
}
```

After `strip_prefix`, `value_str` is `rest_of_json_from_value_start` (e.g. `429","error":"rate_limit"}`). `strip_suffix('"')` would only succeed if the entire remaining JSON ends with `"` — it strips from the wrong end, returning `None` for any field that isn't the last one in the object.

This silently falls through to `unwrap_or_else(|| "unknown")` for fields like `"error"` or any future quoted `"error_status"`.

**Currently masked** because all three extracted fields (`error_status`, `attempt`, `retry_delay_ms`) happen to be numeric, so the numeric branch is taken. But the fix in bb590d6b that introduced this function has a latent correctness defect: any caller that tries to extract a quoted string field (e.g. adding `error_message` logging) will get `"unknown"` with no indication of failure.

## What the fix should be

Find the closing `"` within the value, not at the end of the string:

```rust
if value_str.starts_with('"') {
    let rest = &value_str[1..]; // after opening quote
    let end = rest.find('"')?; // find closing quote
    Some(rest[..end].to_string())
}
```

This correctly extracts quoted string values regardless of their position in the JSON object.

gabrielkoerich:
## Summary

Fixed extract_json_field to correctly extract quoted string values by finding the closing quote within the value rather than stripping from the wrong end. Added regression test.

### Files changed

- `src/engine/runner/fallback.rs:46-57`

Closes #2817

---
*Task #2817 · Created by minimax[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/engine/runner/fallback.rs
+++ b/src/engine/runner/fallback.rs
@@ -49,12 +49,9 @@ fn extract_json_field(json: &str, field: &str) -> Option<String> {
         let value_str = &json[value_start..];
 
         // Handle quoted strings
-        if value_str.starts_with('"') {
-            value_str
-                .strip_prefix('"')
-                .and_then(|s| s.strip_suffix('"'))
-                .and_then(|s| s.strip_suffix(','))
-                .map(|s| s.to_string())
+        if let Some(rest) = value_str.strip_prefix('"') {
+            let end = rest.find('"')?; // find closing quote
+            Some(rest[..end].to_string())
         }
         // Handle raw numbers (integers or floats)
         else {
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Branch is up to date with main. Both required CI checks pass. The fix is correct: `strip_prefix('"').and_then(|s| s.strip_suffix('"'))` was broken because after stripping the opening quote, `strip_suffix` operated on the rest of the JSON string (everything after the opening quote of the value), only succeeding when the entire remainder ended with `"` — i.e., the field was last in the object. The replacement `rest.find('"')?` correctly finds the closing quote of the value regardless of position. The test covers mid-object quoted strings, end-of-object quoted strings (no trailing comma), and numeric fields as a regression guard. No scope creep, no TODOs left behind.

---

gabrielkoerich:

---

PR #2842, opened 2026-04-19
bug: parser treats missing status as done and can mark malformed agent output as success

### LINKED ISSUE #2835: bug: parser treats missing status as done and can mark malformed agent output as success
## Summary
`map_generic_response()` defaults `status` to "done" when a JSON blob contains any supported field but omits `status`/`result`. This can classify malformed/partial agent output as a successful completion.

## Where
- `src/parser.rs:415-420`

## Why this is a bug
- The parser only requires *any* supported field (`summary`, `message`, `files`, etc.) to enter generic mapping.
- If `status` is missing, it currently does:
  - `unwrap_or("done")`
- That means payloads like `{"summary":"still working"}` normalize to `done`, which feeds success paths in runner outcome classification.
- Result: tasks can be marked complete even when the model did not provide an explicit terminal status.

## Expected behavior
- Missing `status`/`result` should not default to success.
- Parser should either:
1. return an error (preferred), or
2. map to a non-success state like `blocked` with explicit parse error context.

## Suggested fix
- In `map_generic_response()`, require `status` or `result` to be present and string-typed; otherwise return `Err`.
- Add regression tests for generic JSON payloads containing `summary`/`files` without status to ensure they do not parse as `done`.

gabrielkoerich:
Resolves #2835

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -416,7 +416,7 @@ fn map_generic_response(val: &serde_json::Value) -> anyhow::Result<AgentResponse
         .get("status")
         .or_else(|| obj.get("result"))
         .and_then(|v| v.as_str())
-        .unwrap_or("done")
+        .context("agent response is missing required 'status' or 'result' field")?
         .to_string();
 
     let summary = obj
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

CI passes. The fix is a clean one-liner: replace `unwrap_or("done")` with `.context(...)?` so missing status/result returns an error instead of defaulting to done. Three regression tests cover the key cases (summary only, files+summary, message only). Code follows existing patterns. No issues found.

---

gabrielkoerich:

---

PR #2943, opened 2026-04-22
fix(store): ensure every task_run gets a terminal outcome

### LINKED ISSUE #2937: bug(task_runs): runs can remain NULL outcome indefinitely after early-return and shutdown paths
## What I found

`task_runs` contains a large number of stale rows with `outcome IS NULL` and `completed_at IS NULL`, including tasks that are no longer running (many are `done`).

### Evidence
- Query:
  - `SELECT COUNT(*) FROM task_runs tr JOIN tasks t ON t.id=tr.task_id WHERE tr.outcome IS NULL AND tr.started_at < datetime('now','-30 minutes') AND t.status NOT IN ('in_progress');`
  - Result: **176** rows
- Recent examples (last 12h):
  - `task_runs.id=10088` -> task `2908` currently `blocked`, `started_at=2026-04-22T01:21:05Z`, still NULL outcome
  - `task_runs.id=10092` -> task `2917` currently `needs_review`, `started_at=2026-04-22T01:24:30Z`, still NULL outcome
- Historical examples include rows attached to `done` tasks (same pattern: no outcome/completed_at).

## Why it is happening (root cause)

1. `start_run` writes rows with `outcome = NULL` by design:
   - `src/store/tasks.rs:1722-1728`
2. `complete_run` is called only after `self.run(...)` succeeds in `run_with_context`:
   - `src/engine/runner/mod.rs:1011-1013` uses `?` on `self.run(...)`; early `Err` exits `run_with_context` before completion write
   - `complete_run` is later at `src/engine/runner/mod.rs:1202-1229`
3. Graceful shutdown resets task status and kills sessions, but does not finalize open `task_runs` rows:
   - `src/engine/mod.rs:1893-1971`

So certain failure/interrupt paths leave orphaned run records permanently incomplete.

## Why it matters

- Distorts failure analytics and cost/outcome reporting
- Makes incident debugging harder (looks like ghost in-progress runs)
- Requires special-case filtering downstream (`sync.rs` already has empty-run bypass logic), masking data-integrity problems instead of fixing source

## Suggested fix

- Ensure every `start_run` has a terminal write:
  - in `run_with_context`, wrap `self.run(...)` and call `complete_run` on error path too (e.g. `outcome='failed'` or `outcome='aborted'`, with error details)
- On graceful shutdown (`engine/mod.rs`), finalize active incomplete runs for tasks being reset (`in_progress`/`in_review`) with `outcome='aborted'`, `completed_at=now`, and a structured shutdown reason
- Add invariant tests:
  - no `outcome IS NULL` rows for non-`in_progress` tasks older than a short threshold
  - forced error/shutdown paths still produce terminal run records

## Related history

This looks related to old incomplete-run handling issues (e.g. #1270), but the data integrity problem is still reproducible on current HEAD.

gabrielkoerich:
## Summary

- Fix `run_with_context` to always call `complete_run` on early-return paths: error from `run()` now sets `outcome=failed`, guard-skip now sets `outcome=skipped`
- Fix graceful shutdown (`engine/mod.rs`) to abort incomplete runs for all tasks being reset (in_progress + in_review), setting `outcome=aborted`
- Add `abort_runs_for_tasks(task_ids, reason)` helper to `TaskStore` that chunks by 500 to stay within SQLite parameter limits

## Root cause

`start_run` writes rows with `outcome = NULL` by design, but three exit paths never called `complete_run`:
1. `run()` returning `Err` — `?` operator in `run_with_context` propagated immediately
2. Guard skipping — returned early without finalizing the run
3. Graceful shutdown — reset task statuses and killed sessions but left runs orphaned

## Test plan

- [x] `cargo fmt -- --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes  
- [x] `cargo nextest run` passes (3368 tests)

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -1895,9 +1895,14 @@ pub async fn serve() -> anyhow::Result<()> {
                         // tmux sessions will be killed when the process exits.
                         let mut reset_count = 0u32;
                         let mut review_reset_count = 0u32;
+                        let mut task_ids_to_abort: Vec<i64> = Vec::new();
+
                         for engine in &project_engines {
                             if let Ok(tasks) = engine.task_manager.list_external_by_status(Status::InProgress).await {
                                 for task in &tasks {
+                                    if let Ok(Some(store_id)) = engine.store.resolve_task_id(&engine.repo, &task.id.0).await {
+                                        task_ids_to_abort.push(store_id);
+                                    }
                                     if let Err(e) = engine.task_manager.update_task_status(&task.id, Status::Routed).await {
                                         tracing::warn!(task_id = task.id.0, ?e, "failed to reset task on shutdown");
                                     } else {
@@ -1908,6 +1913,7 @@ pub async fn serve() -> anyhow::Result<()> {
                             // Also reset internal in_progress tasks
                             if let Ok(tasks) = engine.store.list_internal_by_status(&engine.repo, crate::store::TaskStatus::InProgress).await {
                                 for task in &tasks {
+                                    task_ids_to_abort.push(task.id);
                                     let task_id = format!("internal:{}", task.id);
                                     if let Err(e) = engine.task_manager.update_task_status(
                                         &crate::backends::ExternalId(task_id.clone()),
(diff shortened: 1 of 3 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by glm[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

PR #3006, opened 2026-04-23
fix(control,tests): prevent orphaned memories on DB failure; fix flaky test paths

### LINKED ISSUE #2942: bug(control): persist_memories called before insert_control_message — memories can be orphaned on DB failure
## Problem

In `src/control.rs`, `persist_memories()` at line 774 is called **before** `store.insert_control_message()` at lines 792-807. If the `insert_control_message` call fails (DB error, constraint violation, etc.), the `?` operator propagates the error and the function returns early — but the memories are already durably written to the KV store.

```rust
// Line 772-807 in src/control.rs
let parsed = parse_response(&result.text);

persist_memories(store, session_id, &parsed.memories).await?;  // ← writes KV first

// ... cost calculation ...

store
    .insert_control_message(          // ← if this fails, memories are orphaned
        session_id,
        "assistant",
        ...
    )
    .await?;
```

## Impact

If `insert_control_message` fails, the session's KV store contains memory keys (e.g. `control:memory:{session}:{key}`) with no corresponding assistant message in the `control_messages` table. On the next session context assembly, those memories are injected into the prompt as if the exchange happened — but the conversation history shows no exchange at all. This creates a subtle context inconsistency that can confuse the agent.

## Fix

Swap the order: call `store.insert_control_message()` first (lines 792-807), then call `persist_memories()` at line 774's position. That way if the message insert fails, the `?` short-circuits before any KV state is written, keeping the two stores in sync.

**File:** `src/control.rs`, lines 772-807

### LINKED ISSUE #2967: bug(tests): integration test temp DB path collides under parallel execution
## Problem

In `tests/integration_engine.rs:81`, the `store_opens_and_migrates` test constructs a temporary database path using only the process ID:

```rust
let tmp = std::env::temp_dir().join(format!("orch-engine-open-{}.db", std::process::id()));
```

`cargo test` runs async tests concurrently within the **same process** on different threads. Since `std::process::id()` is process-scoped, all concurrent test threads share the same PID and therefore the same temp file path.

This causes a race:
- Thread A creates and opens `orch-engine-open-{pid}.db`
- Thread B tries to create and open the same file path simultaneously
- One thread may delete the file while the other is using it, or SQLite may encounter a busy/locked database

Other tests in the same file correctly use an atomic counter (the `temp_db()` helper at line 170), but this early test does not.

## Impact

- Flaky test failures under `cargo test` (especially with `cargo nextest run` or high parallelism).
- Can cause spurious CI failures that are hard to reproduce.
- The cleanup at lines 84-86 (`remove_file`) can delete a file another thread is actively using.

## Fix

Use the same `temp_db()` helper (or an atomic counter) that other tests in the file use:

```rust
// Existing helper at ~line 170:
fn temp_db() -> PathBuf {
    static COUNTER: AtomicU64 = AtomicU64::new(0);
    let n = COUNTER.fetch_add(1, Ordering::SeqCst);
    std::env::temp_dir().join(format!("orch-test-{}-{}.db", std::process::id(), n))
}
```

Then:
```rust
let tmp = temp_db();
```

## File/Line

- `tests/integration_engine.rs:81`

## Severity

Low — test-only issue, but causes real flaky CI failures.

gabrielkoerich:
## Summary

- **`src/control.rs`** (#2942): Swap `persist_memories` / `insert_control_message` order so if the DB write fails, KV memory keys are never written — preventing orphaned memories that would inject stale context into future sessions.
- **`tests/integration_engine.rs`** (#2967): Replace all 7 PID-only temp DB paths (`orch-engine-*-{pid}.db`) with the existing `temp_db()` helper which appends an atomic counter, eliminating the race condition when multiple test threads share the same path.

## Test plan

- [x] `cargo fmt -- --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo nextest run` — 3383/3383 passed

Closes #2942
Closes #2967

---
*Task internal-2963 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/control.rs
+++ b/src/control.rs
@@ -771,9 +771,10 @@ pub async fn send_message(
     // Parse response for summary and memory tags
     let parsed = parse_response(&result.text);
 
-    persist_memories(store, session_id, &parsed.memories).await?;
-
-    // Store assistant message with token usage
+    // Store assistant message with token usage.
+    // Must happen before persist_memories so that if the DB write fails the
+    // KV store is not left with orphaned memory keys that have no corresponding
+    // conversation history entry.
     let total_tokens =
         total_tokens_u64_to_i64_saturating(result.input_tokens, result.output_tokens);
     let input_tokens = opt_token_u64_to_i64_saturating(result.input_tokens);
@@ -805,6 +806,8 @@ pub async fn send_message(
             cost_usd,
         )
         .await?;
+
+    persist_memories(store, session_id, &parsed.memories).await?;
 
     // Release the per-session lock before eviction check.
     drop(_guard);
--- a/tests/integration_engine.rs
+++ b/tests/integration_engine.rs
@@ -78,7 +78,7 @@ impl ExternalBackend for MockBackend {
 /// the checksum, and the engine loops forever on startup.
 #[tokio::test]
 async fn store_opens_and_migrates() {
-    let tmp = std::env::temp_dir().join(format!("orch-engine-open-{}.db", std::process::id()));
+    let tmp = temp_db("open");
     let store = TaskStore::open(&tmp).await;
     assert!(store.is_ok(), "TaskStore::open failed: {:?}", store.err());
     let _ = std::fs::remove_file(&tmp);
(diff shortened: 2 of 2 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Required checks are green (`test`, `secrets`). I also ran local fallback checks from CI workflow commands (`cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, SELECT-* guard, `zola --root docs build`, `cargo nextest run`) and they passed. Diff was fully reviewed (2 files) and is aligned with architecture constraints in AGENTS.md; no blocking correctness/security/migration/config issues found.

---

PR #3266, opened 2026-06-05
Add /restart command to control plane (Telegram, Discord, chat)

### LINKED ISSUE #3263: Add /restart command to control plane (Telegram, Discord, chat)
## Summary

Add a `/restart` command to the control plane so the orch service can be restarted from any connected channel — Telegram, Discord, or `orch chat` — not just the local terminal.

## Motivation

Restarting currently requires `orch service restart` (or `brew services restart orch`) on the host machine. When away from the host, there's no way to recover from a hung state, pick up new config, or apply a fresh binary after `brew upgrade`. All existing channels already have an authenticated bidirectional connection — they should be able to drive this.

## Proposed behavior

- New control command `/restart` available in:
  - Telegram channel
  - Discord channel
  - `orch chat` REPL
- Restricted to authorized users (reuse each channel's existing auth/allowlist)
- Acknowledge the request, then trigger the existing graceful shutdown path so in-flight tasks reset to `routed` / `needs_review` cleanly
- After restart, post a confirmation in the originating channel (store a "restart pending: <channel>:<chat_id>" marker in the KV store, read it on startup, and emit a message)

## Implementation notes

- Reuse the graceful shutdown machinery (SIGTERM handler) — don't add a parallel teardown path
- The actual restart needs to be performed by something that outlives the orch process. Options:
  - Exit with a known code and let `brew services` / launchd respawn
  - Spawn a detached helper that waits for the parent to exit, then re-execs
- Whichever option is chosen, it must work under the standard `brew services` install

## Open questions

- Single `/restart` or a generic `/service <action>` (restart, status, version, logs tail)?
- Should `/restart` also be available over the localhost websocket for CLI tools?

## Related

- #3258 (Telegram-specific version — supersede or close as duplicate if this lands)

gabrielkoerich:
## Summary

Added /restart command to control plane (Telegram, Discord, orch chat). Stores restart_pending KV marker before sending SIGTERM to self, posts confirmation on next startup.

### Files changed

- `prompts/control_system.md`
- `src/cli/chat.rs`
- `src/control.rs`
- `src/engine/mod.rs`

Closes #3263

---
*Task #3263 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/prompts/control_system.md
+++ b/prompts/control_system.md
@@ -20,6 +20,7 @@ You can run CLI commands to manage tasks, check status, and take actions. Use ba
 - `orch service restart` — restart the service
 - `/model [agent:]<model>` — switch the sticky model (or show current agent:model)
 - `/agent <claude|codex|opencode>` — switch the sticky agent and its default model (or show current)
+- `/restart` — gracefully restart the orch service; sends confirmation when back online
 - `gh pr list` — list open pull requests
 - `gh run list` — list CI workflow runs
 - `gh issue list` — list open issues
--- a/src/cli/chat.rs
+++ b/src/cli/chat.rs
@@ -41,7 +41,7 @@ pub async fn interactive(session_id: &str) -> anyhow::Result<()> {
             break;
         }
 
-        match control::maybe_handle_control_command(&store, session_id, message).await? {
+        match control::maybe_handle_control_command(&store, session_id, None, message).await? {
             Some(response) => {
                 println!("{response}");
                 println!();
@@ -65,7 +65,7 @@ pub async fn interactive(session_id: &str) -> anyhow::Result<()> {
 pub async fn single_message(session_id: &str, message: &str) -> anyhow::Result<()> {
     let store = crate::cli::init_store().await?;
     if let Some(response) =
-        control::maybe_handle_control_command(&store, session_id, message).await?
+        control::maybe_handle_control_command(&store, session_id, None, message).await?
     {
         println!("{response}");
         return Ok(());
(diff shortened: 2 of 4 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

CI passes (check, secrets, test all green). Branch is already up to date with main.

The implementation correctly:
- Adds `/restart` to `parse_control_command` and the `ControlCommand` enum
- Passes `channel_thread` through the existing `maybe_handle_control_command` / `send_message` call chain — Telegram/Discord set it via `conversation_key()` in `channel_handler.rs`, CLI passes `None`
- Uses SIGTERM after a 2s delay (enough time for the acknowledgment reply to be delivered) — correctly reuses the existing graceful shutdown path as required
- Stores a KV marker `control:restart_pending` before shutdown so the post-restart startup code can send a confirmation back to the originating channel
- Startup confirmation code (in `engine/mod.rs`) runs after 5s (allowing channel connections to re-establish), looks up the marker, sends to the right channel, then deletes the key — including in the malformed-key case to avoid infinite retries
- Auth is implicitly handled by each channel's existing allowlist; `/restart` requires no new auth surface
- CLI path (`orch chat`) correctly gets no post-restart confirmation since `channel_thread` is `None` and the CLI session is gone after restart
- Tests updated: `parse_control_command` coverage and the `expected` set of known commands both updated

---

gabrielkoerich:

---

PR #3275, opened 2026-06-05
fix(parser): add missing normalize_status aliases + guard detect_rate_limit false positives

### LINKED ISSUE #3273: bug(parser): normalize_status missing 'noop', 'green', 'waiting', 'alerts_sent' aliases — 4 task failures Jun 5
## What I Found

4 task runs on Jun 5, 2026 failed with `unrecognized status` errors from new aliases not in the normalization map:

```
151823|internal:151823|claude|sonnet|failed|unrecognized status: green|2026-06-05T04:01:09Z
151870|internal:151870|claude|opus|failed|unrecognized status: noop|2026-06-05T07:38:39Z
151894|internal:151894|claude|sonnet|failed|unrecognized status: WAITING|2026-06-05T08:22:59Z
151928|internal:151928|claude|sonnet|failed|unrecognized status: alerts_sent|2026-06-05T14:00:41Z
```

Note: `alert` alias (tasks 151878, 151648, 151660) was fixed in HEAD (commit 05eb2166, #3254) but not yet deployed in v0.79.0. That one will resolve on next deploy.

## Why It's Happening

`normalize_status` in `src/parser.rs:154` converts non-canonical agent status values to canonical ones. The current `done` arm (HEAD) is:

```rust
"done" | "completed" | "complete" | "skipped" | "skip" | "ok" | "success"
| "no_changes_needed" | "no_trades_no_positions" | "changes_made" | "change_made"
| "changes_pushed" | "acknowledged" | "flat" | "no_action" | "no_action_needed"
| "missed" | "changes_addressed" | "review_addressed" | "already_finalized_in_attempt_1"
| "alert" => "done".to_string(),
```

Missing aliases (all semantically "task completed, nothing to do"):

| Alias | Agent usage | Evidence |
|-------|------------|---------|
| `green` | Trading/health tasks: "everything is green" | internal:151823 |
| `noop` | "no operation performed, nothing changed" | internal:151870 |
| `WAITING` (→ `waiting`) | "waiting for conditions" — agent treating no-action as done | internal:151894 |
| `alerts_sent` | Alert notification tasks: "alerts were sent successfully" | internal:151928 |

`normalize_status` lowercases before matching (`resp.status.to_ascii_lowercase()` at line 155), so `WAITING` → `waiting`. The issue is `waiting` is not in the map at all.

The unknown statuses are kept as-is (line 204: `_ => resp.status.clone()`), then the downstream code sees an invalid status and records `outcome=failed`.

## Why It Matters

- 4 task failures in 12 hours from new aliases (this pattern recurs with each new task type)
- Each failure wastes a complete agent run — the task was successful but the status string is unrecognized
- Tasks are re-dispatched from scratch despite having completed their work
- Pattern is accelerating: 3210→3221→3230→3254→3255→now (new aliases added each self-improvement cycle)

## Suggested Fix

**File: `src/parser.rs`**

Add the 4 new aliases to the `done` arm in `normalize_status`:

```rust
| "alert"
| "noop"          // "no operation performed"
| "green"         // "everything is green / healthy"
| "waiting"       // "waiting for conditions" (from WAITING via to_ascii_lowercase)
| "alerts_sent"   // "alerts dispatched successfully"
=> "done".to_string(),
```

Also add test cases for each new alias in the normalize_status test block at line 940.

**Longer-term fix**: The instruction prompt given to agents should explicitly enumerate the allowed status values (`done`, `in_progress`, `needs_review`, `blocked`) so new domain-specific terms don't leak as status values. Currently the prompt says to set `status` but doesn't constrain the vocabulary.

## Evidence

```sql
sqlite3 ~/.orch/orch.db "SELECT external_id, agent, model, error FROM task_runs WHERE error LIKE '%unrecognized status%' AND started_at > datetime('now', '-24 hours') ORDER BY started_at;"
internal:151823|claude|sonnet|unrecognized status: green
internal:151870|claude|opus|unrecognized status: noop
internal:151894|claude|sonnet|unrecognized status: WAITING
internal:151928|claude|sonnet|unrecognized status: alerts_sent
```

### LINKED ISSUE #3274: bug(runner): opencode false-positive rate_limit when agent runs cargo tests — 'rate_limit' keyword in nextest output triggers detect_rate_limit
## What I Found

On Jun 5, 2026 at 20:29:31Z, a task run for `opencode/nemotron-3-ultra-free` was recorded as `outcome=rate_limit` but the error content is **cargo nextest test output**, not an actual rate limit:

```
task:   151930 (3267)
agent:  opencode
model:  opencode/nemotron-3-ultra-free
outcome: rate_limit
error:  opencode rate limit: 01s] (3551/3639) orch::bin/orch store::tests::recent_memory_errors_on_corrupt_json
        PASS [   0.229s] (3552/3639)... (+330 more chars)
```

The "rate limit" detection triggered on text that is clearly `cargo nextest` output (test run progress: `(3551/3639) orch::...`, `PASS [...]`). This put the model into an unnecessary cooldown.

## Why It's Happening

`detect_rate_limit` in `src/engine/runner/agents/mod.rs:1069` searches for patterns like `"rate_limit"`, `"rate limit"` in the **last 3000 bytes** of combined stdout+stderr. When an opencode agent works on the orch codebase and runs `cargo nextest`, the test output includes test function names that contain these patterns:

```
src/store/tests.rs:4266:async fn record_rate_limit_returns_id() {...}
src/store/tests.rs:4276:async fn rate_limits_counted_in_metrics_summary() {...}
src/engine/runner/agents/mod.rs:1755:fn pattern_detect_rate_limit() {...}
```

The nextest output line would look like:
```
PASS [  0.01s] orch::store::tests::record_rate_limit_returns_id
```

This contains `"rate_limit"` → `detect_rate_limit` fires → `outcome=rate_limit`.

The false positive guard (using `terminal_reason:completed`) exists in `review.rs` for review runs (see test `terminal_reason_completed_guard_suppresses_false_rate_limit` at `mod.rs:3246`), but this guard is **not applied** in the opencode agent runner's task completion path.

## Why It Matters

- False cooldowns prevent the model from being reused for a period (based on the model failure backoff: 5min→...→4h)
- Currently only 1 confirmed occurrence, but every opencode run that executes `cargo nextest` on the orch repo is at risk
- The false rate_limit triggers `record_model_failure` for a perfectly healthy model
- Other test function names that could trigger false positives: `"capacity"` (`test_storage_capacity`), `"throttled"` (`test_throttled_response`), etc.

## Suggested Fix

**File: `src/engine/runner/agents/opencode.rs`** (or wherever opencode parse/classify logic lives)

Apply the same guard already used in `review.rs`: after parsing the opencode JSONL output, if the result indicates successful completion (e.g., the final JSONL line shows a completion status), suppress rate_limit detection:

```rust
// Only apply rate-limit detection when the output does NOT indicate
// successful completion. Test output from cargo nextest can contain
// function names like "rate_limit_returns_id" that trigger false positives.
let completed = output_indicates_successful_completion(&combined);
if !completed {
    if let Some(e) = detect_rate_limit(scan_tail) {
        return e;
    }
}
```

Alternatively, add a separate `safe_head` scan that only fires rate_limit for patterns appearing BEFORE the first JSONL line (i.e., as startup errors), not within the agent's work product.

A more targeted fix: scan for the `"rate_limit"` (underscored) pattern ONLY as a standalone word in an error context (surrounded by spaces/newlines), not as part of a test function name.

## Evidence

```sql
sqlite3 ~/.orch/orch.db "SELECT tr.error FROM task_runs tr WHERE tr.outcome = 'rate_limit' AND tr.agent = 'opencode' AND tr.started_at > datetime('now', '-24 hours');"

opencode rate limit: 01s] (3551/3639) orch::bin/orch store::tests::recent_memory_errors_on_corrupt_json
        PASS [   0.229s] (3552/3639)... (+330 more chars)
```

Related: `review.rs` `terminal_reason_completed_guard_suppresses_false_rate_limit` test (mod.rs:3246) documents this exact pattern for review runs.

Jah-yee:
## Summary

Two small fixes for the orch parser:

### Fix 1: normalize_status missing aliases (closes #3273)

`normalize_status` did not recognize `noop`, `green`, `alerts_sent` → `done` and `waiting` → `in_progress`. This caused agent statuses from some providers to not be normalized correctly.

### Fix 2: detect_rate_limit false-positive guard (closes #3274)

`detect_rate_limit` could trigger on bare `rate_limit` strings early in output (e.g. nextest progress lines or stored-error prefixes) without actual error context. Added a guard: when the match position is < 100 bytes from start, require an error indicator (`429`/`http`/`error`/`failed`) within a 150-byte window.

## Changes

- `src/parser.rs`: add `noop`, `green`, `alerts_sent` to `done`; add `waiting` to `in_progress`
- `src/engine/runner/agents/mod.rs`: add early-match error-context guard in `detect_rate_limit`

## Testing

Existing tests in `pattern_detect_rate_limit` and `detect_rate_limit_message_contains_context_not_tail` pass. The new guard is designed to not affect any existing true-positive cases.

### DIFF
--- a/src/engine/runner/agents/mod.rs
+++ b/src/engine/runner/agents/mod.rs
@@ -1111,6 +1111,25 @@ pub(crate) mod patterns {
                     .next()
                     .is_some_and(|c| c.is_ascii_digit())
             }));
+
+        // Guard against false positives: if "rate_limit" appears near the start of the
+        // text without an error-context indicator nearby, it may be from test output
+        // (e.g. nextest progress lines) or a stored-error prefix rather than a real
+        // rate limit error. Require an error indicator (429/http/error/failed) nearby.
+        if let Some(pos) = match_pos {
+            if pos < 100 {
+                let window_start = pos.saturating_sub(50);
+                let window_end = (pos + 100).min(lower.len());
+                let context = &lower[window_start..window_end];
+                let has_error_context =
+                    context.contains("429") || context.contains("http")
+                    || context.contains("error") || context.contains("failed");
+                if !has_error_context {
+                    return None;
+                }
+            }
+        }
+
         if match_pos.is_some() || has_429 || has_529 {
             let message = if let Some(pos) = match_pos {
                 extract_context_around(text, pos, 300)
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -165,6 +165,8 @@ fn normalize_status(mut resp: AgentResponse) -> AgentResponse {
         | "skip"
         | "ok"
         | "success"
+        | "noop"
+        | "green"
         | "no_changes_needed"
         | "no_trades_no_positions"
         | "changes_made"
@@ -178,11 +180,12 @@ fn normalize_status(mut resp: AgentResponse) -> AgentResponse {
         | "changes_addressed"
         | "review_addressed"
         | "already_finalized_in_attempt_1"
-        | "alert" => "done".to_string(),
+        | "alert"
+        | "alerts_sent" => "done".to_string(),
         // Canonical progress statuses.
         // `partial` is used by some models to indicate partial progress —
         // treat it as in_progress so the task remains open for follow-up.
-        "in_progress" | "running" | "partial" => "in_progress".to_string(),
+        "in_progress" | "running" | "partial" | "waiting" => "in_progress".to_string(),
         "in_review" | "reviewing" => "in_review".to_string(),
         "needs_review" | "pending_review" | "ready_for_review" => "needs_review".to_string(),
         // Canonical error statuses.


gabrielkoerich (changes requested):
Thanks for the PR and welcome! 👋 Splitting feedback by the two fixes since they're independent.

## Fix 1 — `normalize_status` aliases ✅

`noop` / `green` / `alerts_sent` → `done` and `waiting` → `in_progress` match what #3273 documents and are purely additive. Happy to land this once tests are added — issue #3273 explicitly asks for them, and the existing pattern at `src/parser.rs:940-1058` is one test per alias (e.g. `parse_normalizes_noop_alias_to_done`). Without them, nothing prevents a future refactor from silently dropping these arms.

## Fix 2 — `detect_rate_limit` guard ❌ — please rework

Two problems I'd like to see addressed before merging:

### 1. It breaks an existing test

The PR description says existing tests pass, but I checked out the branch and ran `cargo test --lib pattern_detect_rate_limit`:

```
---- engine::runner::agents::tests::pattern_detect_rate_limit ----
assertion failed: patterns::detect_rate_limit(\"You've hit your usage limit\").is_some()
```

Trace: the pattern `\"you've hit your usage limit\"` matches at `pos = 0`. Guard fires (`pos < 100`). Window `[0..27] = \"you've hit your usage limit\"` contains none of `429` / `http` / `error` / `failed` → returns `None`. The `\"You've hit your session limit · resets ...\"` test case has the same problem (`session limit` at pos ~16, no error context).

Those are real production Claude strings — suppressing them turns a 1-occurrence false positive into routine false negatives on real rate limits. Per `CLAUDE.md` the cooldown system errs on the side of caution: a missed rate limit causes credit burn + cascading model failures, which is worse than a brief unnecessary cooldown.

### 2. The guard doesn't actually address the #3274 false positive

The reported FP comes from `cargo nextest` output in the last 3000 bytes of agent stdout:

```
PASS [   0.229s] (3552/3639) orch::store::tests::record_rate_limit_returns_id
```

The `rate_limit` substring sits at whatever byte offset that line lands on inside the 3000-byte tail — typically nowhere near `pos < 100`. The threshold is arbitrary and orthogonal to the failure mode. And even when it does fire near the start, neighboring nextest lines very often contain `error` (test names like `recent_memory_errors_on_corrupt_json`, `errors_propagated`) — so the \"error context\" check passes and the FP still goes through.

### Root cause + suggested direction

The real bug is that `rate_limit` is matching as a substring of `record_rate_limit_returns_id`. Two approaches that fix this without weakening real-error detection:

- **Word-boundary check** on `rate_limit` / `rate limit` / `ratelimit` / `capacity` / `throttled` / `overloaded` — preceding and following char must not be `[A-Za-z0-9_]`. Kills the test-name FP cleanly.
- Or wire in the `terminal_reason:completed` guard already used in `review.rs` (see `mod.rs:3244-3268`) on the opencode task-completion path, as #3274 suggests.

A regression test using the actual nextest line from the issue would be great to lock this in.

## Process nits

- These two fixes are unrelated — splitting them into separate PRs would let Fix 1 land immediately and keep the changelog (auto-generated from conventional commits) clean.
- Before opening, please run `cargo fmt && cargo clippy --all-targets -- -D warnings && cargo nextest run` locally as called out in CLAUDE.md — that would have caught the regression above.

Happy to re-review once Fix 1 has tests and Fix 2 is reworked (or split off into its own PR). Thanks again!

gabrielkoerich:
One more process note, since this will apply to future PRs too: please open **one PR per issue**. This PR bundles fixes for #3273 and #3274, which makes review harder and muddies the auto-generated changelog (CI tags releases from conventional commits on `main`).

Concretely for this PR:

- Keep this branch focused on #3273 (the `normalize_status` aliases) — add the per-alias tests and it can land.
- Open a separate PR for #3274 with the reworked rate-limit guard (word-boundary check or `terminal_reason:completed` guard) and a regression test using the actual nextest line from the issue.

Going forward, one issue → one branch → one PR is the convention here. Thanks!

Jah-yee:
Thanks for the detailed review! I've addressed both fixes in a new PR: **#3276**

## Fix 1 — normalize_status tests ✅
Added all four requested tests (noop, green, alerts_sent → done; waiting → in_progress).

## Fix 2 — detect_rate_limit word-boundary check
Replaced the  guard with a proper  check:

- Removed the positional threshold entirely (that was the bug)
- Check: both chars before/after the pattern must not be alphanumeric or `_`
- Applied **only** to `rate_limit`/`ratelimit` (the patterns that appear in identifiers)
- Natural-language phrases (`usage limit`, `session limit`, `quota exceeded`, etc.) are unaffected

The new approach correctly:
- ✅ Detects "You've hit your usage limit" at position 0
- ✅ Detects "You've hit your session limit · resets 10:30am"
- ❌ Rejects `record_rate_limit_returns_id` (word boundary fails: preceding 'd', following '_')

## On splitting
Happy to close this PR and land #3276 if you prefer one PR per fix. Just say the word!

Jah-yee:
Superseded by #3276 — all requested changes (tests for Fix 1, word-boundary guard for Fix 2) incorporated in the new PR.

PR #3410, opened 2026-07-17
bug(router): opencode free-model discovery cache poisoned by transient CLI failure — 1h false 'all models cooled' degraded windows

### LINKED ISSUE #3409: bug(router): opencode free-model discovery cache poisoned by transient CLI failure — 1h false 'all models cooled' degraded windows
## What I found

The pre-emptive health check repeatedly (and falsely) marks the `opencode` agent as degraded with reason `"all models cooled"`, even though `SELECT key, value FROM kv WHERE key LIKE 'cooldown:opencode%'` shows no active cooldowns for most of opencode's models at the time. Log evidence (last ~30h):

```
2026-07-16T16:44:12Z WARN marking agent as degraded agent="opencode" reason="all models cooled" rate_limit_count=0
2026-07-16T17:45:01Z INFO agent recovered, clearing degraded flag agent="opencode" rate_limit_count=0    # 61 min later
2026-07-16T22:48:00Z WARN marking agent as degraded agent="opencode" reason="all models cooled" rate_limit_count=0
2026-07-16T23:48:50Z INFO agent recovered, clearing degraded flag agent="opencode" rate_limit_count=0    # 61 min later
2026-07-17T13:58:52Z WARN marking agent as degraded agent="opencode" reason="all models cooled" rate_limit_count=0
2026-07-17T14:59:42Z INFO agent recovered, clearing degraded flag agent="opencode" rate_limit_count=0    # 61 min later
2026-07-17T19:19:14Z WARN marking agent as degraded agent="opencode" reason="all models cooled" rate_limit_count=0
2026-07-17T20:20:03Z INFO agent recovered, clearing degraded flag agent="opencode" rate_limit_count=0    # 61 min later
```

4 occurrences in ~30h, each lasting almost exactly 60-61 minutes, with `rate_limit_count=0` every time (i.e. not triggered by actual rate limits). `is_agent_in_cooldown("opencode")` is also not the cause — `cooldown:opencode` in KV is a stale timestamp in the past (`1784103443` = 2026-07-15 08:17:23). During each window, `orch cooldown list` / the KV table shows most opencode free models with `cooldown=0` (never cooled), confirming the degraded flag does not correspond to real cooldown state.

## Why it's happening

Root cause is in `src/engine/runner/agents/opencode.rs`. `discover_free_opencode_models()` (line 712) maintains a 1-hour TTL cache (`FREE_MODELS_CACHE`) of opencode's free-tier models, refreshed by shelling out to `opencode models` via `run_opencode_models_discovery_async()` (line 843). That discovery function returns an **empty `Vec`** on *any* transient failure:

- CLI not in PATH (line 844-847)
- spawn error (line 856-859)
- non-zero exit (line 883-886)
- 30s timeout (line 887-893)

When the 1-hour cache expires, the background refresh task unconditionally overwrites the cache with whatever came back — including empty:

```rust
// opencode.rs:732-741
let discovered = run_opencode_models_discovery_async()
    .await
    .into_iter()
    .filter(|m| m.to_lowercase().contains("free"))
    .collect::<Vec<_>>();
let cache = FREE_MODELS_CACHE.get_or_init(|| Mutex::new((0, Vec::new())));
let now = chrono::Utc::now().timestamp();
if let Ok(mut guard) = cache.lock() {
    *guard = (now, discovered);   // <-- overwrites even when discovered is empty
}
```

There is no check for `discovered.is_empty()` before writing. Contrast with the sibling `ALL_MODELS_CACHE` path used in `expanded_model_pool` (`src/engine/router/config.rs:645-652`), which explicitly guards against acting on an empty discovery result:

```rust
let discovered = crate::engine::runner::agents::opencode::discover_opencode_models();
if !discovered.is_empty() {
    let discovered_set: std::collections::HashSet<String> = discovered.into_iter().collect();
    expanded_pool.retain(|m| discovered_set.contains(m));
}
```

`discover_free_opencode_models()` has no equivalent guard — a single transient `opencode models` failure poisons the cache with `[]` for the full 1-hour TTL. `~/.orch/config.yml` configures opencode's `simple`, `complex`, and `review` complexity tiers with **only** `opencode:free` as the pool entry (github-copilot alternatives are commented out), so `expanded_model_pool()` (`src/engine/router/config.rs:614-654`) expands to an empty pool for those tiers whenever the free-models cache is poisoned. `has_available_model_for_complexity()` then returns `false` for every complexity tier → `model_checker` in `refresh_health()` (`src/engine/router/mod.rs:429-437`) returns `false` → `refresh_degraded_agents()` (`src/engine/cooldown.rs:1023,1034`) marks the agent degraded with reason `"all models cooled"`, even though nothing is actually in cooldown — it's a cache/discovery artifact, not agent health.

## Why it matters

- Confirmed CLI works fine when tested manually (`opencode models` returns instantly, 5 free models, exit 0) — so the failure is transient/intermittent, exactly the class of bug an unconditional-overwrite cache amplifies into an hour-long outage.
- During each ~60-min window, `agent_is_routable("opencode", ...)` filters opencode out of routing entirely for `simple`/`complex`/`review` complexity tasks (its only non-degraded tier, `medium`, uses `kimi: sonnet` per config, i.e. doesn't even route to opencode) — so effectively **all** opencode routing is unavailable for up to an hour per occurrence.
- Observed 4x in 30 hours (~3-4x/day cadence), i.e. opencode is falsely marked fully degraded roughly 3-4 hours per day, forcing tasks that would otherwise go to opencode's free models onto paid agents (claude/codex/kimi/minimax) instead — real cost/capacity impact, not just a log nuisance.
- The log message name (`"all models cooled"`) actively misleads triage — operators/self-improvement runs checking `cooldown:opencode%` KV keys find nothing wrong, because nothing is actually cooled.

## Suggested fix

In `discover_free_opencode_models()` (`src/engine/runner/agents/opencode.rs:728-741`), don't let a failed/empty discovery overwrite a previously-populated cache. Options:
1. Only overwrite the cache when `discovered` is non-empty; on empty, keep the stale (but still valid) prior list and/or set a much shorter retry TTL (e.g. 30-60s) so a real failure is retried quickly instead of waiting the full hour.
2. Distinguish "discovery genuinely ran and found zero free models" from "discovery failed" (e.g. have `run_opencode_models_discovery_async()` return `Option<Vec<String>>`, `None` on any failure path) and only cache `Some(_)` results.

Apply the same guard to `discover_opencode_models()` (`ALL_MODELS_CACHE`, lines 767-800) for consistency, even though it's currently shielded downstream by the `if !discovered.is_empty()` check in `expanded_model_pool` — that guard is incidental protection, not a documented invariant, and a future caller of `discover_opencode_models()` could easily reintroduce the same bug.

## Evidence

```sql
sqlite3 ~/.orch/orch.db "SELECT key, value, datetime(value, 'unixepoch') FROM kv WHERE key LIKE 'cooldown:opencode%';"
cooldown:opencode|1784103443|2026-07-15 08:17:23   -- stale, in the past
cooldown:opencode:opencode/mimo-v2-omni-free|0|1970-01-01 00:00:00   -- never cooled
cooldown:opencode:opencode/qwen3.6-plus-free|0|1970-01-01 00:00:00   -- never cooled
cooldown:opencode:opencode/nemotron-3-super-free|0|1970-01-01 00:00:00   -- never cooled
```

Manual CLI check (current, healthy):
```
$ opencode models
opencode/big-pickle
opencode/deepseek-v4-flash-free
opencode/hy3-free
opencode/mimo-v2.5-free
opencode/nemotron-3-ultra-free
exit code: 0
```

gabrielkoerich:
## Summary

Fixed opencode model discovery so transient empty CLI refreshes no longer poison the cached model pool for an hour and falsely degrade the agent as "all models cooled".

### What was done

- Added a shared cache update helper in `src/engine/runner/agents/opencode.rs` that preserves existing discovered model lists when a refresh returns empty after prior successful discovery.
- Applied that guarded update path to both the free-model cache and the all-models cache so transient `opencode models` failures no longer collapse opencode routability.
- Added regression tests covering empty refresh behavior for both caches plus the cold-cache empty case.
- Ran `cargo fmt -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test opencode::tests:: --lib`, and `cargo nextest run` successfully.
- Committed the fix as `4c2177b3 fix(router): preserve opencode discovery cache on empty refresh`.

### Files changed

Closes #3409

---
*Task #3409 · Created by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.4`*

### DIFF
--- a/src/engine/runner/agents/opencode.rs
+++ b/src/engine/runner/agents/opencode.rs
@@ -55,6 +55,26 @@ pub(crate) fn reset_model_caches_for_test() {
     ALL_MODELS_REFRESH_IN_PROGRESS.store(false, Ordering::Release);
 }
 
+fn update_discovered_models_cache(
+    cache: &Mutex<(i64, Vec<String>)>,
+    discovered: Vec<String>,
+    cache_name: &'static str,
+) {
+    let now = chrono::Utc::now().timestamp();
+    let mut guard = cache.lock().unwrap_or_else(|e| e.into_inner());
+
+    if discovered.is_empty() && !guard.1.is_empty() {
+        tracing::warn!(
+            cache = cache_name,
+            cached_models = guard.1.len(),
+            "opencode model discovery returned empty; preserving previous cache contents"
+        );
+        return;
+    }
+
+    *guard = (now, discovered);
+}
+
 /// Runner for OpenCode agent.
 pub struct OpenCodeRunner;
 
@@ -735,10 +755,7 @@ pub fn discover_free_opencode_models() -> Vec<String> {
                 .filter(|m| m.to_lowercase().contains("free"))
                 .collect::<Vec<_>>();
             let cache = FREE_MODELS_CACHE.get_or_init(|| Mutex::new((0, Vec::new())));
-            let now = chrono::Utc::now().timestamp();
-            if let Ok(mut guard) = cache.lock() {
-                *guard = (now, discovered);
-            }
+            update_discovered_models_cache(cache, discovered, "free");
         };
         if let Ok(handle) = tokio::runtime::Handle::try_current() {
             // Use the current tokio runtime's handle to spawn the async discovery task,
@@ -783,10 +800,7 @@ pub fn discover_opencode_models() -> Vec<String> {
             });
             let discovered = run_opencode_models_discovery_async().await;
             let cache = ALL_MODELS_CACHE.get_or_init(|| Mutex::new((0, Vec::new())));
-            let now = chrono::Utc::now().timestamp();
-            if let Ok(mut guard) = cache.lock() {
-                *guard = (now, discovered);
-            }
+            update_discovered_models_cache(cache, discovered, "all");
         };
         if let Ok(handle) = tokio::runtime::Handle::try_current() {
             handle.spawn(refresh);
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by kimi[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

gabrielkoerich:
## Automated Review — Approve

Required CI checks pass on the latest run (check/test/secrets), rebase onto origin/main was a no-op, and the fix directly addresses the reported bug. The new update_discovered_models_cache helper prevents empty discovery results from overwriting a previously-populated cache in both discover_free_opencode_models and discover_opencode_models, eliminating the 1-hour false-degraded windows. Tests cover preservation of existing free/all model cache contents and the empty-cache-first-population case. One related cache writer, prime_free_model_cache, still overwrites unconditionally, but it is only called at Router::new() when the cache is typically empty and is not the observed failure path.

---

gabrielkoerich:

---

PR #3418, opened 2026-07-20
bug(runner): resolve_project_dir() falls through to service CWD for inactive repos with no bare-clone — worktree creation fails, newly exposed by dispatch_routed_tasks_global

### LINKED ISSUE #3416: bug(runner): resolve_project_dir() falls through to service CWD for inactive repos with no bare-clone — worktree creation fails, newly exposed by dispatch_routed_tasks_global
## What I found

Two tasks (internal DB ids 140644, 140647) that had been stranded in `routed` status since April for a project since removed from the active project list, got dispatched today for the first time by the newly-merged `dispatch_routed_tasks_global` sweep (#3415, commit c03e2312, merged ~2026-07-19 23:17). Both failed identically at worktree creation:

```
failed to create worktree at ~/.orch/worktrees/<project>/gh-issue-493-... for task 493 (stdout: , stderr: fatal: not a git repository (or any of the parent directories): .git, retry stdout: , retry stderr: fatal: not a git repository (or any of the parent directories): .git, retry error: )
```

Same error for the sibling task (490) at effectively the same timestamp. Both tasks have open PRs (review_cycles=1, an existing `pr_number` set) waiting on a fix-up round — they were re-routed back to `routed` after a review requested changes, then got stuck when the project was later removed from the active project list (the exact scenario #3413/#3415 was built to un-stick). Now that they dispatch, they fail one layer deeper.

## Why it's happening

`resolve_project_dir()` (`src/engine/runner/mod.rs:1560-1613`) resolves the source repo directory in this order:
1. `PROJECT_DIR` env var
2. `config::get_project_paths()` — walks the *active* project list in config, matching by `.orch.yml` repo
3. legacy `config.project_dir`
4. bare clone at `~/.orch/projects/<owner>/<repo>.git`
5. **fallback: `std::env::current_dir()`** (line 1611-1612)

For a project that was originally **user-managed** (a local clone path the operator added directly, never an orch-managed bare clone — see the "Directory layout" doc distinction) and has since been commented out / removed from the active project list, steps 1-4 all miss: it's not in the active config anymore, no legacy `project_dir` is set for it, and no bare clone was ever created for it (user-managed projects skip that path entirely by design). Step 5 then returns the **orch service process's own CWD**, which is not this task's repo at all — hence `git worktree add` failing with "not a git repository."

This is old code (unchanged since 2026-06-22, commit b7f960be), but the codepath was never reachable for this class of task before: tasks stranded in `routed` for an inactive repo previously never dispatched at all (that was the whole bug #3413/#3415 fixed). The new global dispatch sweep now reaches this exact combination — inactive AND originally-user-managed AND already has an in-flight PR — for the first time, and it falls through the entire resolution chain to a meaningless fallback.

## Why it matters

The two affected tasks have long-open PRs that can never receive their requested follow-up fix — every dispatch attempt will fail identically forever, since nothing about the failure changes the routing/project state. More importantly, any other task in the same situation (user-managed project, later removed from config, with an in-flight PR needing a fix-up round) will silently hit the same dead end now that #3415 makes these tasks reachable. This is a regression surface directly created by today's own fix, not a pre-existing latent bug that would resolve itself.

## Suggested fix

`resolve_project_dir()`'s final fallback (`src/engine/runner/mod.rs:1611-1612`) should not silently return the service's CWD as if it were a valid project directory. Options:
- Return an explicit error (e.g. "no resolvable project directory for repo X — not in active config, no bare clone, no legacy project_dir") instead of falling through to `current_dir()`. This surfaces a clear, actionable failure instead of a confusing "not a git repository" error from deep inside `git worktree add`.
- Separately, consider whether removing a project from the active config should also decide what happens to its *user-managed* clone path — e.g. persist the last-known project path per repo (not just the bare-clone convention) so tasks from removed-but-once-user-managed projects can still resolve correctly. That's a bigger design question; the immediate fix is just making the failure mode legible instead of misleading.

## Evidence

```
sqlite3 ~/.orch/orch.db "SELECT id, external_id, status, review_cycles, route_attempts, pr_number FROM tasks WHERE id IN (140644,140647);"
140644|493|needs_review|1|0|499
140647|490|needs_review|1|0|498

sqlite3 ~/.orch/orch.db "SELECT task_id, attempt, run_type, agent, model, outcome, started_at, error FROM task_runs WHERE task_id IN (140644,140647) ORDER BY task_id, attempt DESC LIMIT 2;"
140644|6|agent|opencode|opencode/hy3-free|failed|2026-07-20T11:32:03Z|failed to create worktree at <path>/gh-issue-493-... (stderr: fatal: not a git repository ...)
140647|6|agent|claude|sonnet|failed|2026-07-20T11:31:58Z|failed to create worktree at <path>/gh-issue-490-... (stderr: fatal: not a git repository ...)
```

Both first-ever dispatch attempts for these tasks since being stranded in April, both immediately following the merge of #3415 hours earlier, both failing identically.

gabrielkoerich:
## Summary

Fixed resolve_project_dir() in src/engine/runner/mod.rs to return an explicit error instead of silently falling back to the orch service's own CWD when a repo has no active config entry, no legacy project_dir, and no bare clone.

### What was done

- Replaced the final fallback in TaskRunner::resolve_project_dir() (was `Ok(std::env::current_dir()?)`) with an anyhow::bail! that names the unresolvable repo and where it looked (active config, legacy project_dir, bare-clone path), matching the pattern already used in cleanup.rs::resolve_project_dir_for_repo
- Updated the existing test resolve_project_dir_empty_project_dir_env_falls_through to assert the new error behavior (previously asserted the old CWD-fallback succeeded)
- Verified cargo fmt, cargo clippy --all-targets -- -D warnings, and cargo nextest run all pass (3795 tests passed, 19 skipped)
- Committed the change with a conventional commit message

### Files changed

- `src/engine/runner/mod.rs`

Closes #3416

---
*Task #3416 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/mod.rs
+++ b/src/engine/runner/mod.rs
@@ -1608,8 +1608,22 @@ impl TaskRunner {
             }
         }
 
-        // Fall back to current directory
-        Ok(std::env::current_dir()?)
+        // No resolvable project directory: not in active config, no legacy
+        // project_dir, no bare clone. Falling back to the service's own CWD
+        // here would silently point worktree creation at an unrelated
+        // directory (see issue #3416) — surface a clear error instead.
+        anyhow::bail!(
+            "no resolvable project directory for repo {} — not in active config, no bare clone at {}, no legacy project_dir set",
+            self.repo,
+            self.orch_home
+                .join("projects")
+                .join(self.repo.split('/').next().unwrap_or_default())
+                .join(format!(
+                    "{}.git",
+                    self.repo.split('/').nth(1).unwrap_or_default()
+                ))
+                .display()
+        )
     }
 }
 
@@ -2185,8 +2199,18 @@ mod tests {
         let result = runner.resolve_project_dir().await;
         std::env::remove_var("PROJECT_DIR");
 
-        // Should succeed (falls back to current dir) without panicking.
-        assert!(result.is_ok());
+        // No active config entry, no legacy project_dir, no bare clone for this
+        // repo: should return an explicit error rather than falling back to the
+        // service's own CWD (see issue #3416).
+        assert!(
+            result.is_err(),
+            "resolve_project_dir should error when no project dir can be resolved"
+        );
+        let err = result.unwrap_err().to_string();
+        assert!(
+            err.contains("owner/testrepo-nonexistent"),
+            "error should name the unresolvable repo: {err}"
+        );
     }
 
     // ── mock backend for process_delegations ─────────────────────────────────


gabrielkoerich:
🔍 Automated review started

---
*Review started by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.4`*

gabrielkoerich:
## Automated Review — Approve

Required CI checks on the current head passed after the rebase no-op. I read the full diff in `src/engine/runner/mod.rs`, checked the relevant architecture/docs, and found the change aligned with the documented distinction between user-managed project directories and bare clones. Replacing the `current_dir()` fallback with an explicit error fixes the misleading worktree failure without adding scope, and the updated test covers the newly intended behavior. Older failed check entries shown by `gh pr checks` were from the superseded cancelled run and are not blocking.

---
*Reviewed by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.4`*

gabrielkoerich:

---
*Reviewed by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.4`*

PR #3423, opened 2026-07-21
bug(router): timed-out LLM pool entries are never cooled down, unlike other failure types — retried at full cost every cycle

### LINKED ISSUE #3422: bug(router): timed-out LLM pool entries are never cooled down, unlike other failure types — retried at full cost every cycle
## What I found

The service log shows the same router pool entry repeatedly timing out with zero backoff between attempts:

```
grep "router LLM pool entry timed out" /opt/homebrew/var/log/orch.log
# 13 occurrences over ~42h (2026-07-20T03:16 .. 2026-07-21T21:01), all agent="minimax" model="haiku"
```

Every one of these 13 events is immediately followed (within tens of ms, elapsed time matching) by a `"slow tick"` WARN, e.g. elapsed_ms 50073, 50168, 82715, 77365, 80432 — i.e. essentially the full `router.timeout_seconds` window (currently 60s) plus overhead. 13 of the ~16 total "slow tick" warnings observed in this window are directly attributable to this one router pool entry.

## Why it's happening

`route_with_llm` (`src/engine/router/mod.rs`, ~lines 1191-1290) round-robins through the configured `router.pool` entries. On a **non-timeout** failure it calls `record_model_failure(agent, model)` (~line 1283), which applies the standard exponential-backoff cooldown, so `is_model_in_cooldown` (checked at the top of the loop, ~line 1222) skips that entry on future ticks until the cooldown expires.

On a **timeout** (the `is_timeout` branch, ~lines 1263-1275), the code only advances the pool index and returns the error — it never calls `record_model_failure` or records any cooldown at all:

```rust
if is_timeout {
    // Advance pool index... Continuing to try further pool entries
    // in the same tick would accumulate multiple timeout windows and
    // stall the tick loop for minutes (issue #3187).
    self.advance_pool_index_after_attempt(idx, n);
    tracing::warn!(agent, model = model_str, next_pool_idx = self.pool_index,
        "router LLM pool entry timed out — advancing pool index, will retry next tick");
    return Err(e);
} else {
    tracing::warn!(agent, model = model_str, error = %e, "pool entry failed, recording cooldown and trying next");
    crate::engine::runner::response::record_model_failure(agent, model_str).await;
}
```

So a pool entry that times out has no memory across ticks: the very next time round-robin reaches it, it's tried again in full, incurring the entire timeout window again. A chronically-erroring entry gets backed off; a chronically-*timing-out* entry (arguably the worse case, since it burns the full timeout duration rather than failing fast) does not.

## Why it matters

Every recurrence blocks that tick's routing decision for the full `router.timeout_seconds` (60s). This is currently the single largest contributor to "slow tick" warnings (13 of 16, ~81%, in the observed window) — a wasted ~60s of routing latency recurring every few hours purely because the router has no way to remember "this entry just timed out" between ticks, even though the exact same information is already tracked and acted on for non-timeout failures.

Note: this is a different bug from #3187 (which this code comment references) — #3187 was about not burning multiple timeout windows *within the same routing call*, which is preserved here (the function still returns immediately on the first timeout). This issue is about the complete absence of *any* backoff *across calls/ticks* for the timeout case specifically.

## Suggested fix

Apply the same generic per-`(agent, model)` cooldown mechanism in the timeout branch that's already used for the non-timeout branch — i.e. call `record_model_failure(agent, model_str)` (or a shorter-duration variant, since a timeout may be more transient than a hard error) before returning in the `is_timeout` branch. This lets `is_model_in_cooldown` skip a known-slow entry on the next round-robin pass, the same way it already skips a known-erroring one, without reintroducing any same-tick retry cascade.

gabrielkoerich:
## Summary

Applied the fix in src/engine/router/mod.rs: the is_timeout branch in route_with_llm now calls record_model_failure(agent, model_str) before returning, mirroring the non-timeout branch's cooldown behavior. cargo fmt --check passed. cargo clippy is still running in the background; will run cargo nextest run afterward and report final verification results once complete.

### Files changed

- `src/engine/router/mod.rs`

Closes #3422

---
*Task #3422 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/router/mod.rs
+++ b/src/engine/router/mod.rs
@@ -1265,13 +1265,20 @@ impl Router {
                         // then return immediately. Continuing to try further pool entries
                         // in the same tick would accumulate multiple timeout windows and
                         // stall the tick loop for minutes (issue #3187).
+                        //
+                        // Also record a standard cooldown, same as the non-timeout branch,
+                        // so a chronically-timing-out entry is skipped by is_model_in_cooldown
+                        // on future ticks instead of burning the full timeout window again
+                        // every time round-robin reaches it (issue #3422).
                         self.advance_pool_index_after_attempt(idx, n);
                         tracing::warn!(
                             agent,
                             model = model_str,
                             next_pool_idx = self.pool_index,
-                            "router LLM pool entry timed out — advancing pool index, will retry next tick"
+                            "router LLM pool entry timed out — recording cooldown, will retry next tick"
                         );
+                        crate::engine::runner::response::record_model_failure(agent, model_str)
+                            .await;
                         return Err(e);
                     } else {
                         tracing::warn!(


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

CI passes (check, test, secrets all green on latest run). The change adds record_model_failure to the is_timeout branch in route_with_llm, applying the same generic cooldown mechanism already used for non-timeout failures. This correctly prevents chronically-timing-out pool entries from burning the full timeout window every tick. The fix is minimal, uses existing infrastructure, and aligns with settled architecture (cooldown system in src/engine/cooldown.rs). No issues found.

---

gabrielkoerich:

---

Issue #3453, opened 2026-07-29
bug(review-prompt): pending CI status prose still causes review parse errors


gabrielkoerich:
## What I found

A recent review run ended with `outcome=parse_error` because the review agent stopped with a prose status update while required CI was still running instead of producing the required review JSON.

Evidence from the audit trail:

```sql
SELECT id, task_id, agent, model, outcome, started_at, completed_at, error
FROM task_runs
WHERE id = 20083;
```

Result:

```text
20083|155460|claude|sonnet|parse_error|2026-07-28T23:04:06Z|2026-07-28T23:10:14Z|parse error: failed to parse review response
```

The tail of the stored stdout shows the agent ended its turn with prose, not JSON:

```text
CI is still running on the active check run. I'll hold off on any further action until the monitor notifies me that the required jobs have completed — at that point I'll assess the results and issue the final review decision.
```

No review JSON follows that message, so `parse_review_response()` correctly rejected it.

## Why it's happening

This is adjacent to, but not identical to, the fixed background-CI issue in #3438. HEAD now says the CI command must run in the foreground and that a turn must not end while a background task is running. That closes the async-task variant.

The remaining instruction gap is in `prompts/review_task.md`:

- Lines 25-29 require `timeout 300 gh pr checks ... --watch --fail-fast --required || true` and foreground execution.
- Lines 55-58 say `NOT RUN or PENDING` should fall back to local checks.
- Lines 111-113 require JSON and forbid prose summaries, but they do not explicitly say: if CI is still pending after waiting, do not emit a status update and stop; continue to fallback local checks and then emit JSON.

A review agent can still interpret pending CI as a reason to pause the review and report status, producing a parse error even though the command was not backgrounded.

## Why it matters

- This burned one full review run: task_run `20083` ran from `2026-07-28T23:04:06Z` to `2026-07-28T23:10:14Z` before producing no parseable decision.
- It delays otherwise healthy PRs by forcing review reroutes.
- It is hard to distinguish from generic malformed-output parse errors unless the stored stdout is inspected.

## Suggested fix

Tighten `prompts/review_task.md` around Step 2 and Output Format:

1. After the `PENDING` decision branch at lines 55-58, explicitly state that pending CI is not a terminal review state: the agent must run fallback local checks, finish the review, and emit JSON in the same turn.
[...]

PR #3460, opened 2026-08-02
bug(sync): open issue #3453 has no corresponding orch task after 4 days

### LINKED ISSUE #3458: bug(sync): open issue #3453 has no corresponding orch task after 4 days
## What I found

Open issue #3453 in this repo has existed since 2026-07-29 but never spawned a task in the orchestrator's task store. Every other recent open issue has a corresponding task record.

Evidence:

sql: Error: No DBURL given

Usage:
sql [-hnr] [--table-size] [--db-size] [-p pass-through] [-s string] dburl [sqlcommand]
sql [-hnr] [--table-size] [--db-size] [-p pass-through] [-s string] dburl < sql_command_file

See 'man sql' for the options

This returns tasks for issues #3450, #3447, #3444, #3442, #3438, #3437, etc., all with  and recent  timestamps. Querying specifically for  returns nothing:

sql: Error: No DBURL given

Usage:
sql [-hnr] [--table-size] [--db-size] [-p pass-through] [-s string] dburl [sqlcommand]
sql [-hnr] [--table-size] [--db-size] [-p pass-through] [-s string] dburl < sql_command_file

See 'man sql' for the options

title:	bug(review-prompt): pending CI status prose still causes review parse errors
state:	OPEN
author:	gabrielkoerich (Gabriel Koerich)
labels:	bug
comments:	0
assignees:	
projects:	Autonomous Agents Board (Backlog)
milestone:	
issue-type:	
parent:	
sub-issues:	
sub-issues-completed:	
blocked-by:	
blocking:	
number:	3453
--
## What I found

A recent review run ended with `outcome=parse_error` because the review agent stopped with a prose status update while required CI was still running instead of producing the required review JSON.

Evidence from the audit trail:

```sql
SELECT id, task_id, agent, model, outcome, started_at, completed_at, error
FROM task_runs
WHERE id = 20083;
```

Result:

```text
20083|155460|claude|sonnet|parse_error|2026-07-28T23:04:06Z|2026-07-28T23:10:14Z|parse error: failed to parse review response
```

The tail of the stored stdout shows the agent ended its turn with prose, not JSON:

```text
CI is still running on the active check run. I'll hold off on any further action until the monitor notifies me that the required jobs have completed — at that point I'll assess the results and issue the final review decision.
```

No review JSON follows that message, so `parse_review_response()` correctly rejected it.

## Why it's happening

This is adjacent to, but not identical to, the fixed background-CI issue in #3438. HEAD now says the CI command must run in the foreground and that a turn must not end while a background task is running. That closes the async-task variant.

The remaining instruction gap is in `prompts/review_task.md`:

- Lines 25-29 require `timeout 300 gh pr checks ... --watch --fail-fast --required || true` and foreground execution.
- Lines 55-58 say `NOT RUN or PENDING` should fall back to local checks.
- Lines 111-113 require JSON and forbid prose summaries, but they do not explicitly say: if CI is still pending after waiting, do not emit a status update and stop; continue to fallback local checks and then emit JSON.

A review agent can still interpret pending CI as a reason to pause the review and report status, producing a parse error even though the command was not backgrounded.

## Why it matters

- This burned one full review run: task_run `20083` ran from `2026-07-28T23:04:06Z` to `2026-07-28T23:10:14Z` before producing no parseable decision.
- It delays otherwise healthy PRs by forcing review reroutes.
- It is hard to distinguish from generic malformed-output parse errors unless the stored stdout is inspected.

## Suggested fix

Tighten `prompts/review_task.md` around Step 2 and Output Format:

1. After the `PENDING` decision branch at lines 55-58, explicitly state that pending CI is not a terminal review state: the agent must run fallback local checks, finish the review, and emit JSON in the same turn.
2. In Output Format around lines 111-113, add a sentence such as: `Never end with a CI-waiting status update. If CI is still pending after the required wait command, run fallback local checks and output the JSON decision using those results.`
3. Add a regression test or fixture if review prompt tests exist; otherwise a prompt snapshot assertion is enough.

This is a prompt-instruction fix, not a parser change: the parser is correct to reject non-JSON review status prose. confirms the issue is open and labeled .

## Why it matters

- A bug that exists only as a GitHub issue with no orch task will never be routed, dispatched, or fixed by the autonomous pipeline.
- It creates silent backlog rot: the issue looks tracked on GitHub but is invisible to the engine.
- The gap went unnoticed for 4 days because routine monitoring looks at ID              TYPE         STATUS               PROJECT                  AGENT      AGE    TRIES TITLE
-----------------------------------------------------------------------------------------------------------------------------
internal:155550 internal     in_progress          gabrielkoerich/orch      kimi       3m     1     Daily review (last 24h) and , not at the GitHub issue list.

## Possible causes

- The issue was created via the GitHub API during an agent session, and the subsequent sync tick failed to ingest it.
- A transient sync failure or deduplication bug dropped the issue from the ingest cursor.
- The issue matched an implicit filter in / that excludes certain issue origins or states.

## Suggested fix

1. Add a reconciliation query that compares open GitHub issues in this repo against the  table and warns/logs when an open issue has no task record.
2. Investigate why #3453 specifically was skipped; fix the ingest gap so it cannot recur.
3. Until the root cause is fixed, manually create a task for #3453 so the review-prompt bug it describes can be worked.

## Scope note

This is about the orchestrator's own issue ingestion for this repo, not a downstream project.

gabrielkoerich:
## Summary

Fixed the root cause of GitHub issues silently disappearing from orch's ingest pipeline: the issues_last_ingested cursor advanced to chrono::Utc::now() instead of the max updated_at actually observed in the fetch, creating a watermark race where an issue whose updated_at lags real time (API replication lag or request latency) gets excluded from the current fetch while the cursor still jumps past it, permanently hiding it from every future since query.

### What was done

- Traced ingest_external_tasks in src/engine/sync.rs and confirmed it advanced the issues_last_ingested KV cursor to wall-clock now() unconditionally after every fetch
- Identified the watermark race: an issue whose GitHub updated_at lags the cursor value (replication lag / request latency) is silently excluded from every subsequent since-filtered fetch forever, matching the reported symptom of issue #3453 never getting a task
- Fixed ingest_external_tasks to derive the cursor from the max updated_at actually observed in the fetched tasks instead of now(), matching the safer per-item cursor pattern already used by the mentions ingestion path (MentionCursor::advance)
- Added a regression test (ingest_advances_cursor_to_max_observed_updated_at_not_wall_clock) asserting the cursor equals the fetched task's updated_at, not wall-clock time
- Ran cargo fmt, cargo clippy --all-targets -D warnings, and the full test suite (cargo nextest run) — all pass except one pre-existing, environment-dependent tmux test failure (session_blocks_dispatch_cleans_dead_session) verified to fail identically on unmodified origin/main, unrelated to this change
- Committed the fix to a single commit on this branch

### Remaining

- Issue #3453 itself is still not ingested: its updated_at already predates the corrupted cursor value from the original bug, and the engine's 24h startup fallback window won't reach back far enough to recover it since it has had no activity in 4 days. Recovering that specific issue requires an operator action (e.g. bump its updated_at with a comment/label, or a manual DB nudge) — this is outside the scope of a code fix and is not something this agent should do by mutating the live production task store

### Files changed

- `src/engine/sync.rs`

Closes #3458

---
*Task #3458 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -2638,9 +2638,20 @@ pub(crate) async fn ingest_external_tasks(
     }
 
     // Record last ingest time for next incremental fetch.
-    let now = chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string();
-    if let Err(e) = store.kv_set(&kv_key, &now).await {
-        tracing::error!(key = kv_key, err = %e, "ingest: failed to record last ingest time");
+    //
+    // Advance the cursor to the max `updated_at` actually observed in this fetch, not to
+    // wall-clock now(). Using now() creates a watermark race: an issue created or updated
+    // between the moment GitHub computed the `since` query results and the moment this line
+    // runs (API replication lag, or simply request latency) can be excluded from *this*
+    // fetch, and then the cursor jumps past its updated_at anyway. On every subsequent tick
+    // `since` is newer than that issue's updated_at, so it is silently skipped forever unless
+    // it receives a new comment/label to bump updated_at, or the engine restarts (which clears
+    // the cursor, see clear_issues_last_ingested). Deriving the cursor from observed data only
+    // is the same safe pattern used by the mentions cursor (see MentionCursor::advance above).
+    if let Some(latest) = all_tasks.iter().map(|(t, _)| t.updated_at.as_str()).max() {
+        if let Err(e) = store.kv_set(&kv_key, latest).await {
+            tracing::error!(key = kv_key, err = %e, "ingest: failed to record last ingest time");
+        }
     }
 
     Ok(())
@@ -2849,6 +2860,36 @@ mod tests {
             .unwrap();
         assert_eq!(task2.title, "Second issue");
         assert_eq!(task2.status, crate::store::TaskStatus::InProgress);
+    }
+
+    #[serial(cooldown_state)]
+    #[tokio::test]
+    async fn ingest_advances_cursor_to_max_observed_updated_at_not_wall_clock() {
+        // Regression test for the watermark race: the `issues_last_ingested` cursor must be
+        // derived from the max `updated_at` actually seen in the fetch, not from
+        // chrono::Utc::now(). Otherwise an issue whose updated_at lags real time (e.g. GitHub
+        // API replication lag between the `since` query and this line running) gets skipped by
+        // every subsequent `since` fetch forever, even though it was never actually ingested.
+        crate::engine::cooldown::reset_global_state().await;
+        let mut task = make_ext_task("1", "First issue");
+        task.updated_at = "2026-01-01T00:00:00Z".to_string();
+        let backend: Arc<dyn ExternalBackend> =
+            IngestMockBackend::with_tasks(vec![(Status::New, task)]);
+        let store = Arc::new(TaskStore::open_memory().await.unwrap());
+
+        ingest_external_tasks(&backend, "owner/repo", &store)
+            .await
+            .unwrap();
+
+        let cursor = store
+            .kv_get("issues_last_ingested:owner/repo")
+            .await
+            .unwrap();
+        assert_eq!(
+            cursor,
+            Some("2026-01-01T00:00:00Z".to_string()),
+            "cursor must match the max updated_at observed in the fetch, not wall-clock now()"
+        );
     }
 
     #[serial(cooldown_state)]


gabrielkoerich:
🔍 Automated review started

---
*Review started by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.4`*

gabrielkoerich:
## Automated Review — Approve

Required GitHub CI for PR #3460 is green on the latest pull_request run (`check`, `secrets`, and `test` all passed on August 2, 2026). The older failed `push` checks shown by `gh pr checks` are stale from the pre-rebase head and are not a blocker. I read the full diff and the change is aligned with the sync architecture: advancing `issues_last_ingested` to the max observed `updated_at` fixes the watermark race without introducing scope creep, and the added regression test covers the intended behavior. Local fallback validation also passed for formatting, clippy, docs build, and the new targeted regression test; an ad hoc full `cargo nextest run` invocation exited non-zero locally without surfacing a failing test name, but GitHub CI is authoritative here and passed.

---
*Reviewed by codex[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `gpt-5.4`*

gabrielkoerich:

---

PR #3494, opened 2026-08-09
bug(github): transport-level HTTP failures open the github:5xx circuit breaker — local connectivity blips halt the routing phase for 5 minutes

### LINKED ISSUE #3492: bug(github): transport-level HTTP failures open the github:5xx circuit breaker — local connectivity blips halt the routing phase for 5 minutes
## Summary

`send_with_retries()` in `src/github/http.rs` has two failure arms that both end by calling `engine_cooldown::set_agent_cooldown("github:5xx", circuit_cooldown_secs)`:

- The **5xx response** arm (~line 484) — a genuine GitHub server error. Correct.
- The **transport error** arm (~line 509) — `Err(e)` from `req.send().await`, i.e. DNS failure, connection refused, no route to host, TLS failure. This is a *local* condition; GitHub was never reached and its health is unknown.

Both write the same key with the same 300s duration, so a local connectivity blip is recorded and treated as a GitHub outage.

That key is not inert. `tick.rs` (~line 1435) checks it and logs `GitHub 5xx circuit breaker open — skipping routing phase`, and `cooldown.rs` throttles non-critical work on it. So a few seconds of local network unavailability suspends routing for five minutes.

## Evidence (last 24h)

Two breaker openings, both from the transport arm, both while the host's network interface was still coming back after a resume:

```
WARN orch::github::http: HTTP send failed, will retry if attempts remain err=error sending request for url (...) attempt=0
WARN orch::github::http: HTTP send failed, will retry if attempts remain err=error sending request for url (...) attempt=1
WARN orch::github::http: HTTP send failed, will retry if attempts remain err=error sending request for url (...) attempt=2
WARN orch::github::http: HTTP send failed after 3 attempts — setting circuit-breaker
WARN orch::engine::sync: comment fetch failed err=GitHub API transient 5xx circuit-breaker active for 300s
WARN orch::engine::cooldown: GitHub 5xx circuit breaker OPEN (discovered via generic cooldown) — throttling non-critical work remaining_secs=300
```

Note the error text: `error sending request for url` with no status code. Nothing was received from GitHub. Concurrent `telegram getUpdates failed` warnings for the same wall-clock windows confirm the failure was local to the host, not specific to GitHub.

Downstream cost, same day: **126** `GitHub 5xx circuit breaker open — skipping routing phase` log lines. All three retry attempts complete within ~5 seconds, then routing is suspended for 300 seconds.

## Root cause

The transport arm treats "I could not send the request" as equivalent to "the server returned 5xx". They warrant different responses:

- 5xx: the server is unhealthy — backing off protects both sides. A shared, long, cross-cutting breaker is right.
- Transport error: the *local* network is unavailable — nothing about GitHub is known. The retries already absorb a brief blip; a 5-minute global breaker on top over-corrects, and mislabels the condition in the log and in the `github:5xx` key that other subsystems key off.

## Suggested direction

Separate the two conditions:

- Keep the 5xx arm exactly as it is.
- For the transport arm, either do not set the shared `github:5xx` key at all (let the caller's error handling and the next tick's natural retry cover it), or set a distinct, much shorter key so the routing-phase skip is not triggered by a local blip.
- Either way the log message should distinguish "GitHub returned 5xx" from "could not reach GitHub" — the current wording sends the operator looking at GitHub status when the problem is on the host.

Explicitly **not** proposed: any config change, removing the circuit breaker, or disabling retries. The retry loop is correct; only the transport arm's escalation to the shared breaker is wrong.

## Impact

- Routing suspended for 5 minutes per local network blip, with no GitHub-side justification.
- Log lines misattribute a local condition to GitHub, sending triage in the wrong direction.
- The `github:5xx` cooldown key — read by several subsystems as a signal about GitHub's health — carries a value that says nothing about GitHub.

gabrielkoerich:
## Summary

Now let's check tick.rs where the breaker is checked for skipping the routing phase.
Now let's check existing tests around this function to see if any assert on the old behavior/log text.
That test doesn't touch this path (make_req itself fails, not req.send()). Good, no conflicts. Let's run the relevant test module to be safe.
All passing. Now let's add a regression test verifying that a transport error does not trip the `github:5xx` circuit breaker.
Good, in-memory map, no store required. Now 

### What was done

- Transport errors still retry with the same exponential backoff as before — no behavior change to the retry loop itself.
- Log messages now say "transport error, GitHub not reached" instead of conflating with 5xx, so triage points at the host's network instead of GitHub's status.
- Added a regression test (`send_with_retries_transport_error_does_not_trip_5xx_breaker`) that forces a real connection-refused error and asserts `github:5xx` cooldown stays unset.

### Files changed

- `src/github/http.rs`

Closes #3492

---
*Task #3492 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/github/http.rs
+++ b/src/github/http.rs
@@ -494,9 +494,17 @@ impl GhHttp {
                     return Ok(resp);
                 }
                 Err(e) => {
-                    // Network/transport error — treat as transient like 5xx
-                    tracing::warn!(err = %e, attempt, "HTTP send failed, will retry if attempts remain");
-                    last_err = Some(anyhow::anyhow!("HTTP send failed: {}", e));
+                    // Transport error — the request never reached GitHub (DNS failure,
+                    // connection refused, no route to host, TLS failure, etc). This is a
+                    // local/network condition, not evidence that GitHub itself is unhealthy,
+                    // so unlike the 5xx arm above it must NOT trip the shared "github:5xx"
+                    // circuit breaker: other subsystems (routing-phase skip, cooldown
+                    // throttling) key off that breaker as a signal about GitHub's health.
+                    tracing::warn!(err = %e, attempt, "HTTP request failed to send (transport error, GitHub not reached), will retry if attempts remain");
+                    last_err = Some(anyhow::anyhow!(
+                        "HTTP request failed to send (transport error): {}",
+                        e
+                    ));
                     if attempt + 1 < attempts {
                         // Cap exponent at 63 (max valid shift for u64) to prevent overflow
                         let exp = 1u64.checked_shl(attempt.min(63)).unwrap_or(u64::MAX);
@@ -506,12 +514,13 @@ impl GhHttp {
                         tokio::time::sleep(sleep_dur).await;
                         continue;
                     }
-                    // Exhausted attempts — set circuit-breaker and break so last_err is read
+                    // Exhausted attempts — break so last_err is read. Deliberately does not
+                    // call set_agent_cooldown("github:5xx", ...): GitHub was never reached,
+                    // so its health is unknown and the shared breaker must stay untouched.
                     tracing::warn!(
-                        "HTTP send failed after {} attempts — setting circuit-breaker",
+                        "HTTP request failed to send after {} attempts (transport error, GitHub not reached) — not tripping the github:5xx circuit-breaker",
                         attempts
                     );
-                    engine_cooldown::set_agent_cooldown("github:5xx", circuit_cooldown_secs).await;
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

CI passes (check, test, secrets all green). The diff correctly separates transport-level HTTP failures from genuine GitHub 5xx responses by removing the `set_agent_cooldown("github:5xx", ...)` call from the transport error arm in `send_with_retries()`. The 5xx arm remains unchanged. This prevents local connectivity blips from triggering the routing-phase skip at tick.rs:1431. The new test validates that transport errors (connection refused on a bound-then-dropped listener) do not trip the `github:5xx` circuit breaker. Clean, well-scoped fix that aligns with the settled architecture for the `github:5xx` cooldown key.

---

gabrielkoerich:

---

PR #3508, opened 2026-08-12
bug(runner): opencode model-discovery cache timestamp isn't updated on failed refresh — no backoff, list frozen at startup snapshot

### LINKED ISSUE #3506: bug(runner): opencode model-discovery cache timestamp isn't updated on failed refresh — no backoff, list frozen at startup snapshot
## What I found

`orch.log` shows the opencode model-discovery background refresh logging "returned empty; preserving previous cache contents" repeatedly across a 71-hour window (2026-08-09 21:22 through 2026-08-12 21:02), with **zero** corresponding successful-refresh log entries and the exact same `cached_models` count (7 for the free-model cache, 54 for the all-models cache) on every occurrence:

```
2026-08-09T23:16:19Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
2026-08-10T00:16:39Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-10T01:17:29Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
2026-08-10T02:18:19Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-10T05:19:59Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-10T06:20:49Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
2026-08-10T07:21:39Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
2026-08-10T09:22:40Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-10T10:23:09Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
2026-08-12T15:52:40Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-12T16:53:11Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
2026-08-12T17:54:00Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-12T18:54:51Z WARN opencode model discovery returned empty; preserving previous cache contents cache="all" cached_models=54
2026-08-12T20:56:30Z WARN opencode model discovery returned empty; preserving previous cache contents cache="free" cached_models=7
```

Running `opencode models` manually on the same host succeeds in ~1.8s with output, and does not hit the 30s subprocess timeout, so the CLI itself is not broken.

## Why it's happening

`update_discovered_models_cache()` (`src/engine/runner/agents/opencode.rs:58-76`), added by the #3409 fix to stop empty discoveries from *wiping* the cache, has a side effect nobody accounted for: it returns **without bumping the cache timestamp** when preserving stale contents.

```rust
fn update_discovered_models_cache(cache, discovered, cache_name) {
    let now = chrono::Utc::now().timestamp();
    let mut guard = cache.lock().unwrap_or_else(|e| e.into_inner());

    if discovered.is_empty() && !guard.1.is_empty() {
        tracing::warn!(cache = cache_name, cached_models = guard.1.len(),
            "opencode model discovery returned empty; preserving previous cache contents");
        return;   // <-- guard.0 (timestamp) is never touched here
    }

    *guard = (now, discovered);   // only the success path updates the timestamp
}
```

The 1-hour TTL check in `discover_free_opencode_models()` / `discover_opencode_models()` (same file, ~lines 767-774 and 819-825) is `now.saturating_sub(*ts) < 3600`. Because a failed refresh never advances `ts`, the cache looks perpetually "expired" the moment it first goes stale — every subsequent call that lands after the original 1-hour window immediately re-triggers a new background subprocess spawn (gated only by the `..._REFRESH_IN_PROGRESS` flag against *concurrent* runs, not against *repeated sequential* ones). There is no backoff: a persistently-failing discovery call is retried on essentially every call instead of once per hour as the TTL design intends.

Separately (harder to fully confirm without adding a success-path log line, which currently doesn't exist): the observed 100% failure rate across 14 refresh attempts over 71h, combined with the model counts never changing even once, suggests the background-refresh code path (`discover_free_opencode_models`/`discover_opencode_models`, which spawn via `tokio::runtime::Handle::try_current().spawn(...)` onto the shared engine runtime) may behave differently from the startup-priming path (`prime_free_model_cache`, `src/engine/runner/agents/opencode.rs:858-884`, which runs the same underlying `run_opencode_models_discovery_async()` in a dedicated `std::thread::spawn` + fresh `tokio::runtime::Runtime`) — the dedicated-thread path is presumably what populated the cache with its current 7/54 entries in the first place, since manual invocation of the CLI works fine.

## Why it matters

- The free/all opencode model pools have effectively been frozen at their last-successful-refresh snapshot (likely last service restart) for the observed 71-hour window: newly available free models are never picked up, and models that disappear from the provider are only ever routed around one-by-one via the existing per-model cooldown mechanism, never removed from the pool at the source.
- This is consistent with the long tail of "opencode dispatches dead/unavailable model" issues already closed in this tracker (#3475, #3169, #3151, #3141, #3118, #3109, #3017, #3010) — a working discovery refresh is the mechanism that's supposed to prevent the pool from accumulating dead models in the first place.
- Even setting aside pool staleness, the missing backoff means a persistently-failing discovery keeps spawning a real subprocess (`opencode models`, ~1-2s CPU each) far more often than the intended once-per-hour cadence, for a call that is failing every time.

## Suggested fix

1. In `update_discovered_models_cache()`, always update the timestamp (`guard.0 = now`) even on the preserve-stale-contents path, so a failing discovery genuinely backs off for the full TTL window instead of being retried on next call. Only `guard.1` (the model list) should be conditionally preserved.
2. Add a `tracing::debug!`/`info!` on the success path of the same function (there currently isn't one), so it's possible to tell from logs whether background refreshes are ever succeeding at all versus this being a persistent 100% failure — right now that distinction is invisible.
3. Once (2) is in place and a real failure is confirmed on the background-refresh path specifically, compare it against `prime_free_model_cache`'s dedicated-thread approach to find what differs (e.g. capture stderr instead of `Stdio::null()` — `src/engine/runner/agents/opencode.rs` around line 897 — the discovery subprocess's stderr is currently discarded, which is likely why the debug-level failure branches at lines ~902, ~929, ~942 have never surfaced a reason in the log even in debug builds).

## Evidence

Log lines above are verbatim from `/opt/homebrew/var/log/orch.log`, 2026-08-09 through 2026-08-12 (14/14 discovery-refresh log lines in that window are the empty-preserve warning; zero success-path log lines exist because none are emitted). Manual `time opencode models` on the same host: completed in 1.84s with normal output, confirming the CLI itself is healthy.

gabrielkoerich:
## Summary

Fixed the opencode model-discovery cache bug: update_discovered_models_cache() now advances the cache timestamp even when preserving stale contents on a failed refresh, so a persistently-failing discovery genuinely backs off for the full 1h TTL instead of retrying on nearly every call. Also added a debug log line on the successful-refresh path so future log analysis can distinguish real successes from the empty-preserve warning.

### What was done

- Updated update_discovered_models_cache() in src/engine/runner/agents/opencode.rs to set guard.0 = now on the preserve-stale-contents path (previously only guard.1 was implicitly kept and guard.0 was never touched)
- Added a tracing::debug! log line on the success path (cache name + discovered model count) so background-refresh success vs. 100%-failure is now distinguishable in logs
- Updated the two existing unit tests (empty_free_discovery_preserves_existing_cache_contents, empty_all_models_discovery_preserves_existing_cache_contents) to assert the timestamp advances past the seeded value while cache contents are still preserved
- Ran cargo fmt -- --check, cargo clippy --all-targets -- -D warnings, and cargo nextest run (full suite, 3868 tests) — all clean

### Remaining

- Item 3 from the issue (comparing the background-refresh subprocess path against prime_free_model_cache's dedicated-thread approach, e.g. capturing stderr instead of Stdio::null()) was intentionally left for a follow-up once the new debug log line confirms whether background refreshes are failing or simply never running

### Files changed

Closes #3506

---
*Task #3506 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/agents/opencode.rs
+++ b/src/engine/runner/agents/opencode.rs
@@ -69,9 +69,18 @@ fn update_discovered_models_cache(
             cached_models = guard.1.len(),
             "opencode model discovery returned empty; preserving previous cache contents"
         );
+        // Still advance the timestamp so a persistently-failing discovery
+        // backs off for the full TTL window instead of being retried on
+        // every subsequent call.
+        guard.0 = now;
         return;
     }
 
+    tracing::debug!(
+        cache = cache_name,
+        discovered_models = discovered.len(),
+        "opencode model discovery refreshed cache"
+    );
     *guard = (now, discovered);
 }
 
@@ -1535,7 +1544,9 @@ mod tests {
         update_discovered_models_cache(cache, Vec::new(), "free");
 
         let guard = cache.lock().unwrap_or_else(|e| e.into_inner());
-        assert_eq!(guard.0, 123);
+        // Timestamp must advance on a failed refresh so the TTL provides a
+        // real backoff window instead of retrying on every subsequent call.
+        assert!(guard.0 > 123);
         assert_eq!(guard.1, vec!["opencode/model-free".to_string()]);
     }
 
@@ -1557,7 +1568,7 @@ mod tests {
         update_discovered_models_cache(cache, Vec::new(), "all");
 
         let guard = cache.lock().unwrap_or_else(|e| e.into_inner());
-        assert_eq!(guard.0, 456);
+        assert!(guard.0 > 456);
         assert_eq!(
             guard.1,
             vec![


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/laguna-s-2.1-free`*

gabrielkoerich:
## Automated Review — Approve

CI passes on the latest run (31641626118: check, test, secrets all green; review-gate green). The older run's checks were cancelled by the concurrency group (cancelled ≠ failed). The fix correctly advances guard.0 in the 'preserving stale contents' path, ensuring the 1h TTL provides real backoff instead of retrying on every call. Debug log added on the success path for observability. Tests updated from assert_eq! to assert!(guard.0 > ...) to verify timestamp advancement. No clippy warnings, formatting is clean. No architecture conflicts. Minimal, targeted change.

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/laguna-s-2.1-free`*

gabrielkoerich:

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/laguna-s-2.1-free`*

PR #3511, opened 2026-08-12
bug(router): route_with_llm candidate filter is agent-level only — agents whose only complexity-tier model is cooled still get offered to the classifier

### LINKED ISSUE #3509: bug(router): route_with_llm candidate filter is agent-level only — agents whose only complexity-tier model is cooled still get offered to the classifier
## Summary

`route_with_llm()` in `src/engine/router/mod.rs` (~line 1199) builds its LLM candidate list by filtering only agent-level state:

```rust
let uncooled_agents: Vec<String> = self
    .available_agents
    .iter()
    .filter(|a| !is_agent_in_cooldown(a) && !is_agent_degraded(a))
    .cloned()
    .collect();
```

Every other routing path (`agent_is_routable`, used by round-robin and weighted round-robin) additionally checks `config.has_available_model_for_complexity(agent, complexity)`, which looks at whether the *specific model* that would be resolved for that complexity tier is itself in cooldown (`is_model_in_cooldown(agent, model)`). `route_with_llm` doesn't do this — presumably because complexity isn't known until the LLM responds, so there's no single complexity to check against up front.

The result: an agent whose configured model is cooled for every complexity tier it could plausibly be assigned still gets offered to the classifier on every routing call, forever, until the model's cooldown itself expires. There's a post-hoc "routing sanity" check that catches the mismatch and reroutes to a working agent, so there's no functional failure — but every one of these is a fully wasted LLM classification call.

## Evidence

In the last 24h, `orch.log` shows 17 occurrences of:

```
WARN ... orch::engine::router: LLM selected cooled agent/model; rerouting to available agent agent=minimax fallback=claude complexity=medium|complex
```

All 17 are `agent=minimax`. Its `medium`/`complex`/`review` tiers all resolve to `minimax:opus` per `model_map` — that model has been continuously cooled for well over a day (`failure_count:minimax:opus` = 19, multi-hour persisted cooldown throughout the window). `minimax` itself has no agent-level cooldown active (`cooldown:minimax` is a stale/expired past timestamp), so it always passes the `uncooled_agents` filter and gets handed to the classifier — which then picks it roughly 2 out of every 3 times it's offered for a medium/complex task, every time requiring the reroute-to-claude fallback.

## Suggested direction

Extend the `route_with_llm` candidate filter to also exclude an agent when none of its configured models (across all complexity tiers, since the eventual complexity isn't known yet) are currently uncooled — i.e. reuse the same per-model cooldown check `has_available_model_for_complexity` already performs, but aggregated across tiers instead of gated on one. This is a generic classifier-scope fix, not a per-agent or per-model special case — it applies to any agent whose only resolvable models happen to be cooled, not just the one in this evidence.

Not proposing any change to cooldown durations or the backoff mechanism itself — those are working as designed. This is specifically about the candidate list the LLM router hands to the classifier before a model is ever chosen.

gabrielkoerich:
## Summary

Fixed route_with_llm's LLM candidate filter to exclude agents whose every configured model (across all complexity tiers) is cooled, not just agent-level cooldown/degradation, so classifier calls are no longer wasted on agents like minimax whose only resolvable model is stuck in cooldown.

### What was done

- Added Router::agent_has_any_available_model() helper that checks has_available_model_for_complexity across simple/medium/complex/review tiers (mirrors the aggregate check the pre-emptive health check already used)
- Updated route_with_llm's uncooled_agents filter to also require agent_has_any_available_model, so agents with no agent-level cooldown but a fully-cooled model no longer get offered to the LLM classifier
- Refactored refresh_health's inline model_checker closure to reuse the new helper instead of duplicating the tier-loop logic
- Added two regression tests: agent_has_any_available_model_false_when_only_model_cooled and llm_candidate_excludes_agent_with_all_models_cooled
- Verified cargo fmt --check, cargo clippy --all-targets -D warnings, and cargo nextest run (3872 tests) all pass

### Files changed

- `src/engine/router/mod.rs`

Closes #3509

---
*Task #3509 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/router/mod.rs
+++ b/src/engine/router/mod.rs
@@ -424,17 +424,8 @@ impl Router {
         // `orch cooldown list` simultaneously reporting "No active cooldowns".
         sync_from_kv(store).await;
 
-        let config_ref = &self.config;
         let agents = self.available_agents.clone();
-        let model_checker = |agent: &str| -> bool {
-            // Agent has at least one available model across any complexity tier
-            for comp in &["simple", "medium", "complex", "review"] {
-                if config_ref.has_available_model_for_complexity(agent, comp) {
-                    return true;
-                }
-            }
-            false
-        };
+        let model_checker = |agent: &str| -> bool { self.agent_has_any_available_model(agent) };
         refresh_degraded_agents(
             store,
             &agents,
@@ -448,6 +439,15 @@ impl Router {
     /// Check if an agent is available.
     pub fn is_agent_available(&self, agent: &str) -> bool {
         self.available_agents.contains(&agent.to_string())
+    }
+
+    /// True if the agent has at least one uncooled model across any complexity
+    /// tier. Used where the eventual complexity isn't known yet (e.g. before
+    /// LLM classification), so we can't check a single tier.
+    fn agent_has_any_available_model(&self, agent: &str) -> bool {
+        ["simple", "medium", "complex", "review"]
+            .iter()
+            .any(|comp| self.config.has_available_model_for_complexity(agent, comp))
     }
 
     fn agent_is_routable(&self, agent: &str, complexity: &str) -> bool {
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/laguna-s-2.1-free`*

gabrielkoerich:
## Automated Review — Approve

CI latest run (check+test+secrets) all PASS; older run was cancelled (stale). cargo fmt, clippy (-- -D warnings), and nextest (167 router tests, including 2 new regression tests) all pass locally. PR adds agent_has_any_available_model(&self, agent) method that aggregates the existing per-model cooldown check (via config.has_available_model_for_complexity) across all 4 complexity tiers, and applies it in route_with_llm's candidate filter so the LLM classifier is no longer offered agents whose only resolvable models are cooled. Same method also refactored into refresh_health (functionally identical to the pre-existing inline closure). No hardcoded model names, no migrations, no config changes, no special-casing — uses only the generic cooldown system. Tests are #[serial(cooldown_state)] #[tokio::test] with proper reset_global_state. The sync.rs model_checker (line 1940) is a separate code path that clones config directly and is out of scope — same logic, different impl.

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/laguna-s-2.1-free`*

gabrielkoerich:

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/laguna-s-2.1-free`*

PR #3474, opened 2026-08-06
bug(runner): opencode step_finish reason="length" truncation not distinguished from malformed-output parse_error

gabrielkoerich:
## Summary

Fixed opencode step_finish reason="length" truncation being misclassified as a generic parse_error and hit with the 4h-7d persistent-model cooldown. Added AgentResult.truncated_by_length, set by find_opencode_result when the trailing step_finish event reports reason=="length". review.rs now checks this flag before the generic parse_error branch, recording a distinguishable "truncated" outcome and applying the standard exponential backoff (record_model_failure) instead of record_persistent_model_failure. Other agent extractors (claude, codex) set the new field to false. Added unit tests for both the truncated and non-truncated opencode NDJSON cases.

### Files changed

- `src/engine/review.rs`
- `src/engine/runner/agents/mod.rs`
- `src/engine/runner/agents/opencode.rs`
- `src/engine/runner/agents/claude.rs`
- `src/engine/runner/agents/codex.rs`

Closes #3473

---
*Task #3473 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

PR #3513, opened 2026-08-13
bug(runner): opencode zero-output truncation still misclassified as parse_error — early-return bypasses truncated_by_length check

### LINKED ISSUE #3512: bug(runner): opencode zero-output truncation still misclassified as parse_error — early-return bypasses truncated_by_length check
## What I found

PR #3474 (commit c2b39771, merged Aug 7) taught `find_opencode_result` to flag `truncated_by_length` when a review model's `step_finish` reason is `"length"`, so the review pipeline applies standard exponential backoff instead of the harsh 4h→7d persistent-model cooldown reserved for genuinely format-incapable models.

That fix does not fire for the most common real-world trigger of length-truncation: a reasoning model that spends its **entire** output/reasoning token budget on hidden reasoning and emits **zero** text tokens. Live evidence from the last 24h, all `run_type='review'`, `agent='opencode'`:

- Task 156428, `opencode/nemotron-3-ultra-free`, outcome `parse_error`. Final `step_finish` in the NDJSON: `{"reason":"length","tokens":{"total":71482,"input":39482,"output":0,"reasoning":40000}}` — 40k reasoning tokens, 0 output tokens.
- Task 156347, `opencode/nemotron-3-ultra-free`, outcome `parse_error`. Final `step_finish`: `{"reason":"unknown","tokens":{"input":0,"output":0,"reasoning":0,...}}`.
- Task 156415, `opencode/nemotron-3-ultra-free`, outcome `parse_error`, run cut off after a `step_start` with no following `step_finish`.

All three should have been classified as `truncated` (standard backoff), not `parse_error` (persistent backoff). Instead the model is currently sitting in a 7h+ cooldown:

```
$ orch cooldown list
KEY                                     REMAINING   REASON
opencode:opencode/nemotron-3-ultra-free 7h16m       persisted
```

Over the same 48h window this model also **succeeded** twice and produced a clean `truncated`-eligible timeout once — it is not a format-incapable model, it's a free-tier reasoning model that occasionally blows its token budget, exactly the case #3474 was written to protect.

## Why it's happening

Root cause is ordering in `find_opencode_result()` (`src/engine/runner/agents/opencode.rs`):

```rust
// line ~1080
let result_text = if is_error { ... } else { extract_ndjson_text(&events).unwrap_or_default() };

if result_text.is_empty() && !is_error {   // line 1087-1089
    return None;                            // <-- bails out here
}

...

let truncated_by_length = events            // line 1111-1116
    .iter()
    .rev()
    .find(|event| event.get(\"type\")... == Some(\"step_finish\"))
    .and_then(|event| event.get(\"part\")?.get(\"reason\")?.as_str())
    .is_some_and(|reason| reason == \"length\");

Some(super::AgentResult { ..., truncated_by_length })
```

When the model produces **zero** text (the common case for a reasoning model that exhausts its budget before writing any answer), `result_text` is empty and `is_error` is false (the opencode `detect_error()` in the same file only treats `step_finish.reason == \"error\"|\"failed\"` as an error — `\"length\"` and `\"unknown\"` are not caught). The function returns `None` at line 1088, **before** `truncated_by_length` is ever computed.

Back in `src/engine/review.rs` (~line 700, ~888), `agent_result` is then `None`, so:

```rust
if agent_result.as_ref().map(|r| r.truncated_by_length).unwrap_or(false) { ... }
```

evaluates to `false` and falls straight into the generic `parse_error` branch (review.rs:1151-1185), which calls `cooldown::record_persistent_model_failure()` — the 4h→7d backoff.

A secondary gap: even when `find_opencode_result` does return `Some`, `truncated_by_length` only recognizes `reason == \"length\"`. Task 156347's final event reports `reason: \"unknown\"` (an opencode-emitted value, not a default from missing-field handling) with all-zero tokens — the same \"model produced nothing\" failure mode, but it wouldn't be caught by the current string match even if reached.

The existing regression test (`find_opencode_result_length_truncation`, opencode.rs ~1730) doesn't catch this because its fixture includes partial text (`\"{\\\"decision\\\":\"`) before the `length` cutoff, so `result_text` is non-empty and the early return never triggers — it tests a different, less common shape of the same failure.

## Why it matters

Free-tier reasoning review models (nemotron-3-ultra-free, and likely other `-free` opencode models with a hidden reasoning budget) get hit with the 4h-7d persistent cooldown for a token-budget fluke instead of the intended short backoff, removing an otherwise-healthy model from the review pool for up to 7 days on a single bad roll. This directly undoes the stated goal of PR #3474/#3473. Over the last 48h this hit 3 of 6 runs for this one model (50% of its runs got the wrong, harsher cooldown).

## Suggested fix

In `find_opencode_result()`, compute `truncated_by_length` (and check for `reason == \"unknown\"` with zero output tokens as an equivalent signal) **before** the empty-`result_text` early return, and when either is true, return `Some(AgentResult { truncated_by_length: true, result_text: \"\".into(), ... })` instead of `None`, so `review.rs`'s existing `truncated_by_length` branch can classify it as `truncated` rather than falling through to `parse_error`. Add a regression test using the exact zero-output/`reason:\"length\"` and zero-output/`reason:\"unknown\"` shapes seen in tasks 156428 and 156347 (both fixtures must have `output:0` and no `text` events, unlike the existing test which includes partial text).

## Evidence

```
sqlite3 ~/.orch/orch.db "SELECT id, task_id, outcome, error FROM task_runs WHERE task_id IN (156347,156415,156428) AND outcome='parse_error';"
20667|156347|parse_error|parse error: failed to parse review response
...|156415|parse_error|parse error: failed to parse review response
...|156428|parse_error|parse error: failed to parse review response

sqlite3 ~/.orch/orch.db "SELECT model, outcome, COUNT(*) FROM task_runs WHERE agent='opencode' AND run_type='review' AND started_at > datetime('now','-48 hours') GROUP BY model, outcome;"
opencode/nemotron-3-ultra-free|parse_error|3
opencode/nemotron-3-ultra-free|success|2
opencode/nemotron-3-ultra-free|timeout|1

$ orch cooldown list
opencode:opencode/nemotron-3-ultra-free 7h16m REMAINING (persisted)
```

gabrielkoerich:
## Summary

Fixed find_opencode_result() in src/engine/runner/agents/opencode.rs to compute truncated_by_length before the empty-result_text early return, so zero-output reasoning-model truncation (step_finish reason="length" or reason="unknown" with zero output tokens) is classified as truncated instead of falling through to the generic parse_error path that triggers the harsh 4h-7d persistent-model cooldown.

### What was done

- Moved truncated_by_length computation (and the token extraction it depends on) above the `result_text.is_empty() && !is_error` early return in find_opencode_result()
- Extended truncated_by_length to also flag step_finish reason="unknown" when output_tokens is exactly 0 (the same 'model wrote nothing' signal seen in task 156347)
- Changed the early return guard to also allow through when truncated_by_length is true, so Some(AgentResult{ truncated_by_length: true, result_text: "" }) is returned instead of None
- Added 3 regression tests: zero-output reason="length", zero-output reason="unknown", and a negative case confirming reason="unknown" with nonzero output tokens is NOT flagged (avoids over-broadening the signal)
- Ran cargo fmt, cargo clippy --all-targets -- -D warnings, and the full cargo nextest run (3878 tests) — all clean/passing

### Files changed

Closes #3512

---
*Task #3512 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/agents/opencode.rs
+++ b/src/engine/runner/agents/opencode.rs
@@ -1084,12 +1084,37 @@ pub fn find_opencode_result(ndjson: &str) -> Option<super::AgentResult> {
         extract_ndjson_text(&events).unwrap_or_default()
     };
 
-    if result_text.is_empty() && !is_error {
-        return None;
-    }
-
     // Extract tokens from step_finish
     let (input_tokens, output_tokens) = runner.extract_tokens(&events);
+
+    // The model was cut off for exceeding its output/reasoning token budget,
+    // not because it produced malformed prose. Distinguish this from a
+    // generic parse failure so callers don't apply the long
+    // format-incapable-model cooldown for a token-budget fluke. A reasoning
+    // model that spends its entire budget on hidden reasoning emits zero
+    // text and a final step_finish of reason="length" (or opencode's
+    // "unknown" with all-zero tokens) — same failure mode, must be computed
+    // before the empty-text early return below or it never gets checked.
+    let last_step_finish_reason = events
+        .iter()
+        .rev()
+        .find(|event| event.get("type").and_then(|v| v.as_str()) == Some("step_finish"))
+        .and_then(|event| {
+            event
+                .get("part")?
+                .get("reason")?
+                .as_str()
+                .map(str::to_string)
+        });
+    let truncated_by_length = match last_step_finish_reason.as_deref() {
+        Some("length") => true,
+        Some("unknown") => output_tokens == Some(0),
+        _ => false,
+    };
+
+    if result_text.is_empty() && !is_error && !truncated_by_length {
+        return None;
+    }
 
     // Extract cost from step_finish if available
     let cost_usd = events.iter().rev().find_map(|event| {
@@ -1103,17 +1128,6 @@ pub fn find_opencode_result(ndjson: &str) -> Option<super::AgentResult> {
             None
         }
     });
-
-    // The model was cut off for exceeding its output/reasoning token budget,
-    // not because it produced malformed prose. Distinguish this from a
-    // generic parse failure so callers don't apply the long
-    // format-incapable-model cooldown for a token-budget fluke.
-    let truncated_by_length = events
-        .iter()
-        .rev()
-        .find(|event| event.get("type").and_then(|v| v.as_str()) == Some("step_finish"))
-        .and_then(|event| event.get("part")?.get("reason")?.as_str())
-        .is_some_and(|reason| reason == "length");
 
     Some(super::AgentResult {
         is_error,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/hy3-free`*

gabrielkoerich:
## Automated Review — Approve

CI green (test/check/secrets pass on latest run). The fix moves truncated_by_length computation ahead of the empty-result_text early return and additionally recognizes reason="unknown" with zero output tokens as the same token-budget-exhaustion signal. Logic verified against extract_tokens (returns Some(0) for output:0) and detect_error (only 'error'/'failed' set is_error). Three new regression tests cover the zero-output length, zero-output unknown, and nonzero-output unknown (negative) cases. No migrations or config changes; aligns with PR #3474 intent.

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/hy3-free`*

gabrielkoerich:

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/hy3-free`*

Issue #3518, opened 2026-08-15
bug(engine): completed agent runs discarded by stuck-task reclaim racing session-exit detection


gabrielkoerich:
## What I found

Three times in the last three days, a `claude`/`sonnet` agent run on a long-running recurring scheduled task **completed successfully** (exit 0, valid JSON result, work committed) but was discarded by `tick_recover_stuck_tasks` reclaiming it as "stuck" ~100-400ms after the tmux session naturally ended — before the runner's own post-session code got a chance to record the result.

Task IDs and timestamps (all `agent=claude`, `model=sonnet`, attempt 1):

| task | dispatched | tmux session died | reclaimed as stuck | attempt-1 result |
|---|---|---|---|---|
| internal:156414 | 08-13 08:01:37Z | 08:19:29.493Z | 08:19:29.629Z (age=17m, threshold=10m) | discarded, but succeeded |
| internal:156454 | 08-14 00:01:xxZ | 00:11:19.346Z | 00:11:19.475Z (age=10m, threshold=10m) | discarded, but succeeded |
| internal:156489 | 08-15 08:01:12Z | 08:15:10.285Z | 08:15:10.411Z (age=13m, threshold=10m) | discarded, but succeeded |

For internal:156489 I inspected the attempt-1 artifacts directly:

- `.../attempts/1/exit.txt` = `0`
- `.../attempts/1/stderr.txt` = empty
- `.../attempts/1/output.json` = 852KB, 773 stream-json lines, final message: `{"status":"success","summary":"Daily bean-close run complete...","total_cost_usd":7.01,"num_turns":132,"duration_api_ms":677028}`

So the agent genuinely finished the task (11+ minutes of real API work, $7.01 spent, 132 turns) and produced a valid completion — but `task_runs` records attempt 1 as `outcome=failed, error="silence detection set task to new"`, and the runner log shows:

```
INFO orch::engine::runner::session: agent session completed task_id="internal:156489"
INFO orch::engine::runner: task already reset by silence detection — skipping fallback processing task_id="internal:156489" status="new"
```

i.e. by the time the runner's own completion path ran, the tick had already reset the task to `new` and cleared routing, so the entire successful result was thrown away. A second attempt was then dispatched from scratch (full re-route, cooldown reshuffle — router picked `minimax`, found it cooled, fell back to `claude` again) and happened to finish in 39s only because it reused the same worktree and found attempt 1's local commits already present.

## Why it's happening

[...]

PR #3528, opened 2026-08-17
bug(engine): stuck-task reclaim race recurred post-#3525 — internal:156673, 16h15m after the fix merged

### LINKED ISSUE #3526: bug(engine): stuck-task reclaim race recurred post-#3525 — internal:156673, 16h15m after the fix merged
## Summary

#3519 (merged 2026-08-15T21:29:30Z) added a dispatch-guard check to `tick_recover_stuck_tasks`'s no-session reclaim paths: skip the reclaim while the task's dispatch guard (the in-memory `dispatching` DashMap, keyed `"{repo}/{task_id}"`, held for the full duration of `runner.run_with_context`) is still present. #3523 (closed 2026-08-16T23:59:37Z via #3525) investigated a reproduction and fixed a real but separate bug (`route_new_tasks_global`'s cross-repo scope leak) — the PR author explicitly noted they could not reproduce the exact dispatch-guard race and asked for a fresh issue with fresh evidence if the symptom recurred.

It recurred, on current `HEAD` (`cc53189f`, which includes #3525), 16h15m after that fix merged.

## Evidence: internal:156673 ("Paper trading" cron task, gabrielkoerich/bean)

Reconstructed from `orch.log`:

- `16:01:01.537Z` — dispatched (`kimi`/`opus`), dispatch guard inserted.
- `16:01:10.535Z` — tmux session `orch-bean-internal-156673` created.
- `16:14:51.270Z` — capture service observes the tmux session has already exited (`session ended, sending final chunk`).
- `16:14:51.551Z` — `tick_recover_stuck_tasks`'s internal-task loop fires: `recovering stuck task: no session found — reclaiming early → new [age_mins=13 threshold_mins=10]`. Task reset to `new`, `task_runs` records `outcome=failed error="silence detection set task to new"`.
- `16:14:53.691Z` — the runner's own completion poll (inside the still-executing `run_with_context`) *only now* notices the session exited (`agent session completed`).
- `16:14:53.695Z` — runner finds the task already reset out from under it: `task already reset by silence detection — skipping fallback processing`.
- `16:15:08Z` — task auto-re-dispatched (to `claude`/`sonnet` after a cooled-agent reroute), completes successfully ~90s later, self-recovering — same "incidental safety net" behavior as every prior occurrence (#3518, #3519, #3523).

No functional harm this time (self-recovered, one duplicate ~90s agent invocation, final result correct — PR #2959 merged cleanly). But the shape is identical to the occurrences that motivated #3518/#3519/#3523.

## Why this still looks like a guard-coverage gap, not "the fix hasn't fully rolled out"

`run_with_context` for this task did not return until *after* `16:14:53.691Z` (the runner's own exit-detection log happens inside that call). That means the dispatch guard entry for `"gabrielkoerich/bean/internal:156673"` should still have been present in the `dispatching` map when `tick_recover_stuck_tasks` ran its `dispatching.contains_key(&dispatch_key)` check at `16:14:51.551Z`. The check should have skipped the reclaim. It didn't — same as the #3523 occurrence.

Two differences from the #3523 repro worth noting in case they narrow the search:
- Agent this time was `kimi`/`opus` (not `claude`/`sonnet`), so this isn't agent-specific.
- `age_mins=13` (not exactly at the `age_mins=10` boundary like #3523's `internal:156563`), i.e. the reclaim fired 3 minutes past threshold rather than right at it — the guard-check tick was itself running late (this window had ongoing GitHub 5xx instability and several slow-tick warnings, though not at this exact minute).

## Suggested direction

#3525's author already ruled out (for the #3523 repro) a static dispatch_key-format mismatch, external_id mutation mid-run, double-dispatch via `dispatch_routed_tasks_global`, and multi-process dispatching-map separation. Their remaining unproven theories were:
- (a) a concurrent repro exercising the real `dispatch_tasks_for_repo` spawn path racing a real `tick_recover_stuck_tasks` call (blocked by `TaskRunner` not being mockable/injectable), or
- (b) a process-restart / multi-instance scenario outside what's visible in the source tree.

This occurrence is a second independent data point (different task, different repo-internal job, different agent) with the same "guard should have been held but the check let the reclaim through" signature — worth prioritizing (a) or (b) now that it's reproduced twice against the same fix.

## Repro data

- Task: `internal:156673`, "Paper trading: scan setups + manage simulated positions" (gabrielkoerich/bean, internal cron job)
- `no_session_stuck_timeout`: 600s (10min)
- Fix commits already merged and in effect at time of this occurrence: `9c70ee00` (#3519), `cc53189f`/#3525 (#3523)
- Reclaim event: `2026-08-17T16:14:51Z`, i.e. `2026-08-16T23:59:35Z` (fix merge) + ~16h15m

gabrielkoerich:
## Summary

Could not conclusively root-cause the dispatch-guard reclaim race (still unreproduced after deep static analysis), but ruled out a DashMap cross-thread visibility bug via a genuine multi-threaded stress test, and promoted the debug! logs that would prove/disprove guard-hold timing to info! so the next occurrence carries direct evidence instead of requiring timestamp correlation.

### What was done

- Traced the full dispatch-guard lifecycle end-to-end: claim in dispatch_tasks_for_repo (tick.rs), RAII release in DispatchGuard::drop, and all three reclaim-check sites in tick_recover_stuck_tasks (same-repo external, cross-repo, internal) — confirmed key format, shared Arc, and claim-before-spawn ordering are all consistent and correct.
- Ruled out several concrete theories: dispatch_key format mismatch between internal-task listing and dispatch claiming (both use the same store_task_to_external conversion), a second/duplicate dispatch entry point (run_with_context has exactly one caller), unexpected dispatching.clear() calls outside tests, and a separate dispatch path for cron-scheduled internal tasks (jobs.rs creates the task then it flows through the same guarded dispatch_tasks_for_repo as everything else).
- Discovered that the debug! logs which would directly prove whether the guard was held at reclaim time ('dispatch guard claimed'/'released' in dispatch_guard.rs and dispatch_tasks_for_repo, and 'skipping ... reclaim: runner still actively processing' in all three tick_recover_stuck_tasks loops) are silently dropped in production because main.rs's default tracing filter is 'orch=info' — every prior occurrence (#3518/#3519/#3523/#3525) had to be diagnosed from indirect timestamp correlation instead of direct evidence. Promoted all five log lines to info! and added dispatch_key to the two that were missing it.
- Wrote a new #[tokio::test(flavor = "multi_thread", worker_threads = 8)] regression test that claims the dispatch guard on the main task (mirroring dispatch_tasks_for_repo's exact claim-before-spawn ordering), holds it on a spawned task pinned to a different worker thread until explicitly released via a oneshot channel, and hammers the real tick_recover_stuck_tasks 300 times from a concurrently-scheduled task while the guard is deterministically still held. Iterated on the test itself: an early sleep-based version was genuinely flaky (~50% failure) but the failure was a TOCTOU in the test harness itself (the loop asserted after the guard had legitimately already released), not a production bug — the corrected deterministic version passed 12,000+ concurrent checks across 40 consecutive runs with zero failures, empirically ruling out a DashMap cross-thread visibility bug as the root cause.
- Ran full quality gates: cargo fmt --check (clean), cargo clippy --all-targets -D warnings (clean), cargo nextest run (3888/3888 passed).

### Remaining

- The underlying race is still unproven — this is now the third occurrence (#3523, #3526, and whatever comes next) where the guard should have been held per static reasoning but the reclaim fired anyway. With the DashMap-visibility theory now empirically ruled out, the two live theories from #3525's own writeup remain: (a) something specific to the real TaskRunner's timing/scheduling under production tick cadence (including tick 'burst' catch-up after slow ticks, e.g. the GitHub 5xx instability noted in this occurrence) that a synthetic guard-hold test can't reproduce, or (b) a process-restart / multi-instance scenario outside the visible source tree.
- The info!-level logging promoted in this PR is the mechanism for gathering that missing evidence: when this recurs, orch.log will now show 'dispatch guard claimed' at dispatch time and 'dispatch guard released' at actual release time for the exact dispatch_key, which can be directly compared against the reclaim's timestamp — settling definitively whether the guard was truly absent (a real bug) or was released moments earlier by something unaccounted for (e.g. an unexpected panic unwind, or the process restarting).

### Files changed

- `src/engine/dispatch_guard.rs`
- `src/engine/tick.rs`

Closes #3526

---
*Task #3526 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/dispatch_guard.rs
+++ b/src/engine/dispatch_guard.rs
@@ -99,7 +99,14 @@ impl Drop for DispatchGuard {
         // signal that lets a future stuck-task-reclaim race be diagnosed from logs
         // by comparing this timestamp against the reclaim check's own log line for
         // the same key, instead of relying on static reasoning about ordering.
-        tracing::debug!(dispatch_key = %self.key, "dispatch guard released");
+        //
+        // info! (not debug!, issue #3526): the default log filter is "orch=info"
+        // (main.rs), so at debug! this line never reached orch.log in production —
+        // every prior occurrence of the reclaim race (#3518, #3519, #3523, #3525)
+        // had no way to show whether the guard was actually still held at reclaim
+        // time. Promoted alongside the matching "dispatch guard claimed" line and
+        // the stuck-task-reclaim skip logs in tick.rs.
+        tracing::info!(dispatch_key = %self.key, "dispatch guard released");
     }
 }
 
--- a/src/engine/tick.rs
+++ b/src/engine/tick.rs
@@ -681,8 +681,16 @@ pub(crate) async fn tick_recover_stuck_tasks(
         if !timing.has_session {
             let dispatch_key = format!("{repo}/{}", task.id.0);
             if dispatching.contains_key(&dispatch_key) {
-                tracing::debug!(
+                // Promoted from debug! to info!: the default log filter is "orch=info"
+                // (see main.rs), so this line — the only direct evidence that the guard
+                // check actually found and honored a held dispatch guard — was silently
+                // dropped in production. Every prior occurrence of the reclaim race
+                // (#3518/#3519/#3523/#3525/#3526) had to be diagnosed from timestamp
+                // correlation alone because this and the "dispatch guard
+                // claimed"/"released" lines never appeared in orch.log.
+                tracing::info!(
                     task_id = task.id.0,
+                    dispatch_key,
                     age_mins = timing.age.num_minutes(),
                     "skipping stuck-task reclaim: runner still actively processing this task"
                 );
@@ -982,9 +990,13 @@ pub(crate) async fn tick_recover_stuck_tasks(
         if !timing.has_session {
             let dispatch_key = format!("{}/{external_id}", task.repo);
             if dispatching.contains_key(&dispatch_key) {
-                tracing::debug!(
+                // See the same promotion note on the internal-task loop below: this was
+                // debug! (silently dropped under the default "orch=info" filter), leaving
+                // every prior reclaim-race occurrence undiagnosable from logs alone.
+                tracing::info!(
                     task_id = task.id,
                     repo = %task.repo,
+                    dispatch_key,
                     age_mins = timing.age.num_minutes(),
                     "skipping cross-repo stuck-task reclaim: runner still actively processing this task"
                 );
@@ -1113,8 +1125,18 @@ pub(crate) async fn tick_recover_stuck_tasks(
         if !timing.has_session {
             let dispatch_key = format!("{repo}/{task_id}");
             if dispatching.contains_key(&dispatch_key) {
-                tracing::debug!(
+                // Promoted from debug! to info! (issue #3526): the default log filter is
+                // "orch=info" (main.rs), so this line — along with the DispatchGuard
+                // claim/release lines in dispatch_guard.rs and dispatch_tasks_for_repo —
+                // never reached orch.log in production. Every occurrence of the reclaim
+                // race (#3518, #3519, #3523, #3525, #3526) had to be reasoned about from
+                // timestamp correlation across unrelated log lines instead of directly
+                // observing whether the guard was held, because the one log line that
+                // would prove it was silently filtered out. Promoting these closes that
+                // observability gap so the next occurrence carries conclusive evidence.
+                tracing::info!(
                     task_id,
+                    dispatch_key,
                     age_mins = timing.age.num_minutes(),
                     "skipping stuck-task reclaim: runner still actively processing this task"
                 );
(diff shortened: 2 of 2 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3.5-lightning-free`*

gabrielkoerich:
## Automated Review — Approve

CI passes (check, secrets, test all green), all 3888 tests pass locally, clippy and fmt check pass. The PR promotes debug! to info! in 5 locations to make dispatch-guard race diagnostics observable in production (since default log filter is "orch=info"), and adds a multi-threaded concurrency regression test that exercises the real tick_recover_stuck_tasks path under cross-thread DashMap visibility. No bugs, security issues, or config violations.

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3.5-lightning-free`*

gabrielkoerich:

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3.5-lightning-free`*

PR #3547, opened 2026-08-23
bug(sync): startup issue-ingest rescan is capped at a 24h lookback, permanently stranding issues skipped before the #3460 cursor fix

### LINKED ISSUE #3545: bug(sync): startup issue-ingest rescan is capped at a 24h lookback, permanently stranding issues skipped before the #3460 cursor fix
## What I found

GitHub issue #3453 (\"bug(review-prompt): pending CI status prose still causes review parse errors\") was created on `2026-07-29T21:03:53Z`, has 0 comments, and its `updated_at` has never changed since creation. As of today (`2026-08-23`, 25 days later) it still has **zero** corresponding row in `tasks`:

```sql
sqlite3 ~/.orch/orch.db "SELECT COUNT(*) FROM tasks WHERE external_id = '3453';"
-- 0
```

For comparison, every other recent bug issue (#3541, #3535, #3530, #3529, #3526, ...) has a matching task row. This exact symptom was already filed and \"fixed\" once, as #3458 (\"open issue #3453 has no corresponding orch task after 4 days\", closed 2026-08-02), whose root cause — the ingest cursor advancing to `chrono::Utc::now()` instead of the max observed `updated_at`, permanently excluding any issue whose `updated_at` lagged the watermark — was fixed by #3460 (commit `adbbddd1`) on `2026-08-02`. That fix has been live for 3 weeks, yet #3453 is still untracked.

## Why it's happening

#3460's fix only prevents the cursor from advancing *past* issues in future fetches — it does nothing to recover issues that the cursor had *already* jumped past before the fix landed. #3453 was already stranded before `adbbddd1` merged (its `updated_at` of `2026-07-29` predates the fix), so the only way it can still be recovered is via the periodic full-rescan-on-restart mechanism (`clear_issues_last_ingested`, called on every engine startup per `src/engine/mod.rs:1120`).

But that \"full rescan\" isn't actually full. Look at `src/engine/sync.rs:2808-2824`:

```rust
// Use 24h fallback on first run or if the key is missing.
// Note: on startup, the engine clears this key for all repos (see clear_issues_last_ingested)
// so the first ingest after a restart always uses the 24h window. ...
let kv_key = format!(\"issues_last_ingested:{repo}\");
let fallback = (chrono::Utc::now() - chrono::Duration::hours(24))
    .format(\"%Y-%m-%dT%H:%M:%SZ\")
    .to_string();
let since = store
    .kv_get(&kv_key)
    .await
    .ok()
    .flatten()
    .unwrap_or(fallback);
```

When `clear_issues_last_ingested` deletes the KV key on startup, the very next `kv_get` returns `None`, and `since` falls back to **`now() - 24h`**, not an unbounded/full scan. `list_active_open_issues(Some(&since))` then asks GitHub for issues updated since that point — which still excludes #3453, because its `updated_at` (`2026-07-29`) is far outside any 24h window computed after `2026-07-30`. Since nobody has commented on or labeled #3453 (it has 0 comments), its `updated_at` never moves into that window on any subsequent restart either. The service has certainly restarted many times since `2026-08-02` (multiple releases have shipped), and every one of those \"full rescans\" was actually just another 24h-bounded fetch that re-missed it.

The 24h fallback itself is intentional and correct for its original purpose — it was added to fix #2581 (\"issues created during engine downtime permanently skipped\"), where the only failure mode was *downtime gaps*, which are naturally bounded (the engine isn't down for more than a day). It was never designed to also serve as the recovery path for a *cursor-watermark* bug whose skip window can be arbitrarily large (here, at least 4 days before #3460 shipped, and now compounding indefinitely since #3453's `updated_at` never advances).

## Why it matters

- #3453 is a live, unaddressed bug report (review-prompt parse errors on pending-CI prose) that has been invisible to the entire task pipeline for 25 days, and will remain invisible forever under the current logic, since nothing will ever bump its `updated_at` unless a human notices and comments on it.
- The same failure mode could silently strand any other old, quiet, pre-#3460 issue — there is no way from the current mechanism to tell whether such issues exist without cross-checking `gh issue list` against `tasks.external_id` by hand (which is how this was found).
- The \"full rescan on startup\" comment (`src/engine/sync.rs:2810`, `3045`) implies a real safety net exists for exactly this scenario, but it silently doesn't cover it.

## Suggested fix

Distinguish the two purposes that currently share the same fallback constant:
- Regular incremental ticks: keep the 24h fallback (or whatever is cheapest) when the KV cursor is merely *missing* for routine reasons.
- The startup rescan specifically (right after `clear_issues_last_ingested` runs): fetch with `since = None` (i.e. all currently-open issues, unbounded by time) rather than `now() - 24h`. This is a one-time, once-per-restart cost bounded by the number of currently-open issues (not a history replay), so it doesn't reintroduce the per-tick cost problem #2493 fixed (that issue was about fetching all open issues *every tick*, not once at startup).
- Separately, one-time backfill: manually reconcile #3453 (and any other issue matching this pattern) once the fix lands, since the code fix alone won't retroactively pick it up without another cursor-clearing restart using the corrected logic.

Files: `src/engine/sync.rs:2808-2824` (fallback/cursor read), `src/engine/sync.rs:3056-3070` (`clear_issues_last_ingested`), `src/engine/mod.rs:1120` (startup call site).

## Evidence

```
$ gh issue view 3453 --json state,comments,createdAt,updatedAt
{"state":"OPEN","comments":0,"createdAt":"2026-07-29T21:03:53Z","updatedAt":"2026-07-29T21:03:53Z"}

$ sqlite3 ~/.orch/orch.db "SELECT COUNT(*) FROM tasks WHERE external_id = '3453';"
0

$ sqlite3 ~/.orch/orch.db "SELECT id, external_id, status, title FROM tasks WHERE external_id IN ('3541','3535','3530','3529','3526');"
156678|3526|done|bug(engine): stuck-task reclaim race recurred post-#3525 ...
156742|3529|done|bug(runner): kimi billing-cycle exhaustion is recorded as outcome=rate_limit ...
156854|3535|blocked|bug(runner): opencode \"not available in your country\" misclassified ...
157140|3541|done|bug(runner): claude/kimi mid-stream disconnect messages misclassified ...
```

gabrielkoerich:
## Summary

Verified and validated the already-committed fix (2dbc0af8) for the startup issue-ingest rescan being capped at a 24h lookback. The previous attempt's diff was already committed to this branch; this attempt re-confirmed it builds, lints, and passes tests, and pushed it to the remote.

### What was done

- Confirmed commit 2dbc0af8 replaces the delete of `issues_last_ingested:{repo}` in clear_issues_last_ingested() with a STARTUP_RESCAN_SENTINEL write, so the next ingest after a restart distinguishes 'never ingested' (24h fallback) from 'startup rescan requested' (since=None, unbounded fetch of all currently-open issues)
- Traced the fix through to the GitHub backend: since_for_fetch=None flows through list_active_open_issues -> list_all_open_issues -> the GitHub REST call, which correctly omits the `since` query param entirely when None, producing a truly unbounded fetch
- Ran cargo fmt -- --check (clean), cargo clippy --all-targets -- -D warnings (clean), and the full cargo nextest run (3932 tests passed, including the two new regression tests ingest_after_startup_rescan_marker_fetches_unbounded and ingest_with_missing_cursor_uses_24h_fallback)
- Pushed the branch to origin (gh-issue-3545-bug-sync-startup-issue-ingest-rescan-is)

### Remaining

- Per the issue's suggested fix, a one-time manual backfill/reconciliation for #3453 (and any other pre-#3460 stranded issues) is still needed once this fix is live — the code fix does not retroactively recover already-stranded issues without another startup rescan running the corrected logic. This is an operational follow-up, not code work.

### Files changed

- `src/engine/sync.rs`

Closes #3545

---
*Task #3545 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/sync.rs
+++ b/src/engine/sync.rs
@@ -2806,30 +2806,32 @@ pub(crate) async fn ingest_external_tasks(
     }
 
     // Track last ingest time for incremental fetches (same pattern as mentions_last_checked).
-    // Use 24h fallback on first run or if the key is missing.
-    // Note: on startup, the engine clears this key for all repos (see clear_issues_last_ingested)
-    // so the first ingest after a restart always uses the 24h window. This prevents issues
-    // created during engine downtime from being permanently skipped (GitHub's `since` filters
-    // by updated_at, not created_at, so issues created while the engine was down would otherwise
-    // be invisible).
+    // Use 24h fallback on first run (key genuinely missing) or if the key holds a normal
+    // timestamp cursor. On startup, the engine marks this key with `STARTUP_RESCAN_SENTINEL`
+    // for all repos (see clear_issues_last_ingested) instead of deleting it, so the first
+    // ingest after a restart can tell "never ingested before" (24h fallback is fine — any
+    // downtime gap is naturally bounded) apart from "cursor deliberately reset for a full
+    // rescan" (fetch with `since = None`, i.e. unbounded by time). A 24h fallback here would
+    // silently re-skip any issue whose `updated_at` already lags more than 24h — which is
+    // exactly the class of issue a startup rescan exists to recover.
     let kv_key = format!("issues_last_ingested:{repo}");
     let fallback = (chrono::Utc::now() - chrono::Duration::hours(24))
         .format("%Y-%m-%dT%H:%M:%SZ")
         .to_string();
-    let since = store
-        .kv_get(&kv_key)
-        .await
-        .ok()
-        .flatten()
-        .unwrap_or(fallback);
-    let since_for_fetch = since.clone();
+    let stored = store.kv_get(&kv_key).await.ok().flatten();
+    let since_for_fetch: Option<String> = match stored.as_deref() {
+        Some(STARTUP_RESCAN_SENTINEL) => None,
+        Some(ts) => Some(ts.to_string()),
+        None => Some(fallback),
+    };
 
     // Fetch all open issues in a single backend call and partition by status label locally.
     // The GitHub backend overrides list_active_open_issues() to use one list_all_open_issues()
     // request instead of a routable call + N per-status calls.
-    // Pass `since` to filter issues updated since last ingest for efficiency.
+    // Pass `since` to filter issues updated since last ingest for efficiency; `None` fetches
+    // all currently-open issues, unbounded by time (used for the once-per-restart rescan).
     let all_tasks: Vec<(crate::backends::ExternalTask, Option<Status>)> = match backend
-        .list_active_open_issues(Some(&since_for_fetch))
+        .list_active_open_issues(since_for_fetch.as_deref())
         .await
     {
         Ok(tasks) => tasks,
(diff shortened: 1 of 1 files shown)


gabrielkoerich:
🔍 Automated review started

---
*Review started by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3.5-lightning-free`*

gabrielkoerich:
## Automated Review — Approve

The PR correctly fixes the bug where startup issue-ingest rescan is capped at 24h lookback. Key changes:

1. **`clear_issues_last_ingested`**: Instead of deleting the KV key, it now sets it to `STARTUP_RESCAN_SENTINEL` (`__startup_full_rescan__`), signaling the next ingest to fetch unbounded (`since = None`).

2. **`ingest_external_tasks`**: Checks for the sentinel value - if present, fetches all open issues unbounded; if a regular timestamp, uses that; if missing (first run ever), uses the 24h fallback.

3. **Tests**: Two new tests verify the sentinel produces unbounded fetch and missing cursor uses 24h fallback.

4. **CI**: All required checks pass; the prior failures were cancellation artifacts, not code issues.

5. **Local testing**: All 3932 tests pass, clippy has zero warnings.

---
*Reviewed by opencode[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opencode/nemotron-3.5-lightning-free`*

gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Branch is already up to date with origin/main (rebase was a no-op). Required CI checks pass: gh pr checks initially showed 3 fail/3 pass, but investigation via `gh api .../check-runs` showed this is a duplicate-trigger artifact — this repo's CI workflow triggers on both `push:` and `pull_request:` with `concurrency: cancel-in-progress: true`. The push-triggered run was cancelled in favor of the pull_request-triggered run, which succeeded on all required jobs (check, secrets, test) for the PR head SHA. No actual failure occurred; this is a pre-existing repo CI configuration behavior unrelated to this PR's diff.

Code review: the fix directly addresses the issue's root cause — clear_issues_last_ingested() now writes a STARTUP_RESCAN_SENTINEL instead of deleting the KV key, letting ingest_external_tasks distinguish 'genuinely missing cursor' (24h fallback, cheap) from 'startup rescan requested' (since=None, unbounded fetch of currently-open issues). After the unbounded fetch, the cursor is correctly advanced to the max observed updated_at (not wall-clock now()), consistent with the watermark-race fix from #3460, so subsequent ticks resume incremental fetching. Verified locally: cargo fmt --check clean, cargo clippy --all-targets -D warnings clean, and cargo nextest run for engine::sync — 84/84 passed, including the two new regression tests (ingest_after_startup_rescan_marker_fetches_unbounded, ingest_with_missing_cursor_uses_24h_fallback). Scope is tight and matches the issue's suggested fix; the one-time manual backfill for #3453 mentioned in the issue is correctly left as an operational follow-up rather than code. No DO NOT TOUCH violations (doesn't touch cooldown/router/config/migrations). Minor non-blocking edge case: if a repo has zero currently-open issues, the sentinel won't be overwritten (no tasks to derive max updated_at from) and the next tick will repeat an unbounded fetch — self-correcting once any issue exists, not worth blocking on.

---

gabrielkoerich:

---

PR #3574, opened 2026-08-31
bug(runner): rate-limit detector doesn't extract relative usage windows, causing retries before quota resets

### LINKED ISSUE #3572: bug(runner): rate-limit detector doesn't extract relative usage windows, causing retries before quota resets
## What I found

A provider-specific usage-limit error is being classified as a transient rate limit and cooled down with the default 5-minute-base exponential backoff, even though the provider's own message states the quota resets on a 5-hour window. This causes the router to retry the same model before the quota has actually refreshed.

Recent task runs in the last 24 hours:

| started_at (UTC) | agent | run_type | outcome | error excerpt |
|------------------|-------|----------|---------|---------------|
| 2026-08-30T23:01:10Z | kimi | agent | rate_limit | `Failed to authenticate. API Error: 403 You've reached your 5-hour usage limit. Your quota will reset when the current 5-hour ...` |
| 2026-08-31T00:11:11Z | kimi | review | rate_limit | same pattern |
| 2026-08-31T09:01:04Z | kimi | agent | rate_limit | same pattern |
| 2026-08-31T10:01:04Z | kimi | agent | rate_limit | same pattern |

All four are stored with `outcome = 'rate_limit'` and no parsed retry timestamp.

## Why it's happening

`src/engine/runner/agents/mod.rs::detect_rate_limit()` (~line 1136) matches generic patterns such as `usage limit` and produces `AgentError::RateLimit`. The cooldown path in `src/engine/cooldown.rs` then tries `parse_retry_at()` (~line 1242) to extract a vendor-specified retry time. The kimi message does not contain an absolute `try again at` or `reset at` timestamp; it only says the quota will reset when the current 5-hour window ends. Because `parse_retry_at()` returns `None`, the cooldown falls back to the standard rate-limit exponential backoff (5-minute base, 4-hour cap), which is far shorter than the 5-hour provider window.

The message is not a classic "billing cycle" exhaustion (there is no monthly/cycle language), so `detect_credit_exhaustion()` in `src/engine/cooldown.rs:482-549` does not classify it either. It sits in a gap: a time-windowed quota that the provider tells us about, but with a relative window instead of an absolute date.

## Why it matters

- Each retry before the 5-hour window refreshes is guaranteed to fail, wasting model calls and review/agent cycles.
- Because the failure is short-cooled, the same model can be re-selected multiple times within the same 5-hour window (observed: two failures one hour apart, then another two one hour apart the next day).
- The generic rate-limit path is designed for requests-per-minute style limits; it does not fit a rolling hourly quota that the provider explicitly advertises.

## Suggested fix

Extend the rate-limit/cooldown pipeline to handle relative usage windows when no absolute retry date is present:

1. In `src/engine/runner/agents/mod.rs::detect_rate_limit()`, preserve the original message so downstream code can inspect it for window hints.
2. In `src/engine/cooldown.rs::parse_retry_at()` (or a new helper called from the rate-limit cooldown path), detect relative windows such as `N-hour usage limit`, `hourly usage limit`, `daily usage limit`, `N-minute usage limit`, etc.
3. When such a window is found and no absolute timestamp is present, set the cooldown to the remaining window length (or the full window length if the start of the window is unknown). This should apply to the specific model, not the whole agent, because the quota is model-scoped.
4. Add regression tests covering messages like `You've reached your 5-hour usage limit. Your quota will reset when the current 5-hour window expires.`

This keeps the fix generic: any provider that emits a relative usage window gets a sensible cooldown instead of the default short backoff.

## Evidence

```sql
SELECT agent, started_at, run_type, outcome, error
FROM task_runs
WHERE agent = 'kimi'
  AND outcome = 'rate_limit'
  AND started_at > datetime('now', '-24 hours')
ORDER BY started_at;
```

Returns the four rows above, all with the `5-hour usage limit` phrasing and `outcome = 'rate_limit'`.

gabrielkoerich:
## Summary

Detect relative usage windows in rate-limit errors and apply model-specific cooldowns instead of default exponential backoff

### Files changed

- `src/engine/cooldown.rs`
- `src/engine/runner/mod.rs`

Closes #3572

---
*Task #3572 · Created by kimi[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `opus`*

### DIFF
--- a/src/engine/cooldown.rs
+++ b/src/engine/cooldown.rs
@@ -12,8 +12,12 @@
 //!
 //! When a rate limit includes a "try again at {date}" message, the cooldown
 //! is set to that specific timestamp instead of the default duration.
+//! Relative usage windows (e.g. "5-hour usage limit") are also detected and
+//! used to set a model-specific cooldown matching the advertised window.
 
 use chrono::TimeZone;
+use once_cell::sync::Lazy;
+use regex::Regex;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex, OnceLock};
 
@@ -616,6 +620,35 @@ pub async fn record_model_failure(agent_name: &str, model: &str) {
     let cooldown_until =
         chrono::Utc::now().timestamp() + compute_backoff(count, base, BACKOFF_MAX_SECS, true);
     set_cooldown_async(&key, cooldown_until, "model_error").await;
+}
+
+/// Record a rate-limit event for an agent+model.
+///
+/// If the error message advertises a relative usage window (e.g.
+/// "5-hour usage limit"), the cooldown is applied to the specific model for
+/// that window length. The quota is model-scoped, so the whole agent is not
+/// cooled. If no window is detected, fall back to the generic agent+model
+/// exponential backoff path.
+pub async fn record_rate_limit(agent_name: &str, model: Option<&str>, error_message: &str) {
+    if let Some(window_secs) = parse_relative_usage_window(error_message) {
+        if let Some(m) = model {
+            let cooldown_until = chrono::Utc::now().timestamp() + window_secs as i64;
+            set_model_cooldown(agent_name, m, window_secs).await;
+            tracing::info!(
+                agent = agent_name,
+                model = m,
+                window_secs,
+                cooldown_until,
+                "relative usage window detected: applying model-specific cooldown"
+            );
+            return;
+        }
+    }
+
+    record_agent_failure_with_message(agent_name, error_message).await;
+    if let Some(m) = model {
+        record_model_failure(agent_name, m).await;
+    }
 }
 
 /// Record a persistent model-specific failure and apply longer exponential backoff.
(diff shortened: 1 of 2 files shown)


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Branch was already up to date with main (rebase was a no-op). Required CI checks (secrets, check, check-release, test) all show SUCCESS on the latest run (33441808948) for commit 156ae880 — an earlier duplicate CI run (33441805219) for the same commit was cancelled/superseded, which is why `gh pr checks` transiently listed 'fail' entries; `gh pr view --json statusCheckRollup` confirms the authoritative latest-per-context result is all green, unrelated to this PR's code. Locally verified cargo fmt --check, cargo clippy --all-targets -D warnings, and `cargo test --lib engine::cooldown::` all pass (66/66).

Code review: the fix is well-scoped and matches the task's suggested approach exactly — `parse_relative_usage_window()` detects relative usage-window phrases (N-hour/minute/day, hourly/daily/weekly/monthly) gated behind usage/quota/window/limit context words to avoid false positives, and `record_rate_limit()` applies a model-specific cooldown for the advertised window instead of the default short exponential backoff, falling back to the previous agent+model backoff behavior when no window is detected (verified fallback path is byte-for-byte equivalent to the old inline code at the single call site in runner/mod.rs:1288). Reuses the existing `set_model_cooldown` helper (same one used by silence detection) rather than inventing a new mechanism. Does not hardcode any model/agent names, consistent with the project's 'no special-casing' cooldown architecture. Absolute vendor dates (parse_retry_at) are unaffected since the window regex doesn't match the Kimi message's 'reset when...' phrasing (confirmed by reading parse_retry_at — no date format matches). New unit tests cover the window extraction, context-gating (no false positive on bare '429 Too Many Requests'/'rate limit exceeded'), and both the windowed and fallback cooldown paths via `record_rate_limit`.

Minor non-blocking observation: `has_context` includes a bare 'limit' keyword, so a message like 'rate limit exceeded, retry in 1 minute' would now take the model-only-cooldown path (60s) instead of the old agent-wide exponential backoff. This is a narrow edge case, produces a short/harmless cooldown, and is not the scenario this PR targets — not worth blocking on.

---

gabrielkoerich:

---

PR #3579, opened 2026-09-01
bug(runner): truncated/malformed claude NDJSON result line loses structured error classification, falls back to generic Unknown

### LINKED ISSUE #3577: bug(runner): truncated/malformed claude NDJSON result line loses structured error classification, falls back to generic Unknown
## What I found

Task 162287, attempt 1 (2026-09-01T08:20:30Z), claude/sonnet: the stored `task_runs.error` is a garbled fragment, not a clean auth error:

```
claude agent error: unknown error (exit 1): i_error","fast_mode_state":"off","fast_mode_disabled_reason":"sdk_opt_in_required","subtype":"success","api_error_status":null,"result":"Failed to authenticate: OAuth session expired and could not be refreshed","type":"result","duration_ms":133,"uuid":"179b27a6-fd8c-41cd-886b-c66cf8b20242"}
^[[?25h
```

Hex dump confirms this is exactly what's stored (not a display artifact of my query tooling):

```
00000000: 636c 6175 6465 2061 6765 6e74 2065 7272  claude agent err
00000020: 7220 2865 7869 7420 3129 3a20 695f 6572  r (exit 1): i_er
00000030: 726f 7222 2c22 6661 7374 5f6d 6f64 655f  ror","fast_mode_
...
00000140: 2d63 3636 6366 3862 3230 3234 3222 7d0d  -c66cf8b20242"}.
00000150: 0a5e 5b5b 3f32 3568 0a0a                 .^[[?25h..
```

Two things are visible in the raw bytes:
1. The JSON line is missing its opening `{"type":"result",...` and is cut mid-field-name — it starts at `i_error` (almost certainly the tail of `is_error`, missing the leading `s`), so `serde_json::from_str` on this line fails.
2. The tail contains the literal 7-byte ASCII sequence `^[[?25h` (a terminal "show cursor" escape sequence rendered in caret notation, not a raw `0x1b` byte) appended after the JSON's closing `}`.

Ten minutes earlier, the *same* underlying condition (`Failed to authenticate: OAuth session expired and could not be refreshed`) on the *same task* was captured cleanly and stored as a plain, well-formed message — so this is intermittent/rare, not a permanent parser bug for this message shape.

Same task, same model, same underlying failure, correctly parsed one attempt earlier (08:01:01Z):
```
claude failed: Failed to authenticate: OAuth session expired and could not be refreshed
```

## Why it's happening

`src/engine/runner/agents/claude.rs` extracts structured errors (`Auth`, `RateLimit`, `ContextOverflow`, etc.) only by successfully JSON-parsing the `"type":"result"` NDJSON line (`extract_error_from_envelope` at claude.rs:182-213, driven from `extract_stream_json_result_text` at claude.rs:231-297 and the line-scanning fallback `detect_error_envelope` at claude.rs:218-229). Both paths call `serde_json::from_str::<Value>(line)` and silently skip the line via `if let Ok(...)` / `if let Some(err) = ...` when parsing fails — there is no fallback keyword/substring classification for a result line that fails to parse as JSON. When that happens for the one line carrying the error, the raw exit code + leftover text falls through to the generic `AgentError::Unknown{exit_code, message}` bucket in `src/engine/runner/fallback.rs:566-618`, which (for non-zero exit codes) has no model-specific cooldown call at all — unlike the `Auth`/`ModelUnavailable`/`InvalidResponse` branches in the same file, which do call `record_model_failure`.

This project already has precedent for defending against exactly this shape of corruption — `summarize_rate_limit_error()` in `src/engine/runner/fallback.rs:22-34` was added (issue #2839) specifically to handle \"truncated fragments that begin mid-object\" for GLM rate-limit output, matching e.g. `,\"subtype\":\"api_retry\",...` even without the leading `{`. That recovery logic is scoped to the rate-limit summarizer only; the auth-detection path in `claude.rs` has no equivalent.

I can't fully prove the corruption mechanism from static analysis alone, but the literal `^[[?25h` cursor-control sequence appended after the JSON is a strong hint: `src/engine/runner/agent.rs:147` wraps every agent invocation in `script -q /dev/null /bin/bash {cmd} | tee {output_file}` (merged via #3554/531bc51f) specifically to give the child process a real TTY. Under a real TTY, CLI tools commonly emit interactive terminal control codes (spinners, cursor hide/show) that they would suppress on a plain pipe. This is not a request to revert that fix — the PTY wrap is a settled, deliberate change for a different bug (silence-detection false positives) — but it does mean the raw captured output can now legitimately contain terminal control bytes interleaved with the NDJSON stream, which the parser has no defense against.

## Why it matters

- When this fires, the run is misclassified as generic `Unknown`/`Failed` instead of `Auth`, which means it skips the model-specific cooldown (`record_model_failure`) that the properly-parsed `Auth` path applies (fallback.rs:321-323) — so the model can be reselected immediately instead of backing off.
- It also pollutes `task_runs.error` with a raw, hard-to-read JSON fragment instead of the clean message the operator/agent actually needs to act on.
- Low frequency in the sample (1 of ~15 non-success runs in a 12h window) but directly reduces the fidelity of the routing/cooldown signal described in AGENTS.md's \"Cooldown and failure recovery is generic\" section — a corrupted line silently degrades to the least-informative error class instead of the most accurate one.

## Suggested fix

Add a truncated-fragment recovery step to `claude.rs`'s error extraction, mirroring what `summarize_rate_limit_error` already does for GLM: when a candidate `\"type\":\"result\"` line fails to parse as JSON, try matching it as a fragment (e.g. via a regex or manual scan for a `\"result\":\"...\"` key/value pair, or `\"is_error\":true`/`\"is_error\":false` sub-fields) rather than only via full-line `serde_json::from_str`. At minimum, run the existing `super::patterns::detect_auth_error` / `detect_rate_limit` / `detect_context_overflow` keyword scanners directly against the raw unparsed line as a last resort before falling back to `AgentError::Unknown`, so a message that plainly contains \"Failed to authenticate\" still gets classified as `Auth` (and gets `record_model_failure`) even when the surrounding JSON is corrupted.

## Evidence

```sql
sqlite3 ~/.orch/orch.db "SELECT length(error), error FROM task_runs WHERE task_id=162287 AND agent='claude' AND error LIKE '%unknown error%';"
-- 344 | claude agent error: unknown error (exit 1): i_error","fast_mode_state":"off",...
```
Full hex dump available on request; key bytes at offset 0x150: `0a 5e 5b 5b 3f 32 35 68 0a 0a` = literal ASCII `\n^[[?25h\n\n`.

gabrielkoerich:
## Summary

Fixed classify_error() in the claude runner so it no longer discards a cleanly-parsed error message behind a byte-truncated fallback string, closing the gap where truncated/malformed claude NDJSON result lines lost structured error classification.

### What was done

- Root-caused the bug by querying the live orch.db for task 162287's actual stdout (10613 bytes): the captured NDJSON was NOT corrupted in transit — it was a complete, valid `type:result`/`is_error:true` envelope
- Found that classify_error() in src/engine/runner/agents/claude.rs already extracts the clean `result` field text into `error_result`, but then re-scans the full combined stdout+stderr via classify_from_text(), whose generic Unknown catch-all truncates to the last 300 bytes (safe_tail) — cutting the well-formed JSON line off mid-field-name (`"is_error"` -> `"i_error"`) once enough preceding NDJSON pushed it past that window
- Fixed classify_error() to prefer the already-extracted clean error_result message (via AgentError::AgentFailed) over the byte-truncated Unknown fallback when no specific pattern detector (auth/rate_limit/context_overflow) matches
- Added a regression test (classify_error_prefers_clean_result_text_over_byte_truncated_tail) reproducing the exact scenario from the DB evidence, verifying the stored error is the clean message and never contains the truncated 'i_error' fragment
- Ran cargo fmt, cargo clippy --all-targets -- -D warnings, and the full test suite (cargo nextest run: 3964 passed, 0 failed) — all clean
- Verified AgentFailed already triggers record_model_failure/record_agent_failure_with_message in fallback.rs (main runner) and falls into record_review_agent_failure's cooldown-recording catch-all in review.rs, so this fix restores the missing cooldown signal for this failure class in both the main task runner and the review pipeline

### Files changed

- `src/engine/runner/agents/claude.rs`

Closes #3577

---
*Task #3577 · Created by claude[bot] via [Orch](https://github.com/gabrielkoerich/orch) using `sonnet`*

### DIFF
--- a/src/engine/runner/agents/claude.rs
+++ b/src/engine/runner/agents/claude.rs
@@ -444,7 +444,23 @@ impl AgentRunner for ClaudeRunner {
             (None, Some(s)) => format!("{s}\n{stdout}\n{stderr}"),
             (None, None) => format!("{stdout}\n{stderr}"),
         };
-        super::patterns::classify_from_text(exit_code, &combined)
+        let classified = super::patterns::classify_from_text(exit_code, &combined);
+
+        // `classify_from_text`'s generic catch-all truncates to the last 300
+        // bytes of the combined text, which can cut a perfectly well-formed
+        // `error_result` line off mid-field-name (e.g. `is_error` becomes
+        // `i_error`) once enough NDJSON precedes it (issue #3577). When we
+        // already extracted a clean, complete error message from the result
+        // envelope but none of the specific pattern detectors matched it,
+        // prefer that clean message over the byte-truncated fallback.
+        if let (AgentError::Unknown { .. }, Some(message)) = (&classified, &error_result) {
+            if !message.is_empty() {
+                return AgentError::AgentFailed {
+                    message: message.clone(),
+                };
+            }
+        }
+        classified
     }
 
     fn router_command(
@@ -982,6 +998,45 @@ mod tests {
         assert!(
             msg.contains("session limit"),
             "error message should contain the rate limit reason, got: {msg}"
+        );
+    }
+
+    #[test]
+    fn classify_error_prefers_clean_result_text_over_byte_truncated_tail() {
+        // Regression test for issue #3577: a well-formed `type:result`
+        // envelope with `is_error:true` was correctly extracted into
+        // `error_result`, but `classify_from_text`'s generic Unknown
+        // fallback then re-truncates the combined stdout+stderr to its last
+        // 300 bytes — which, once enough preceding NDJSON pushes the result
+        // line's start past that window, cuts it off mid-field-name (e.g.
+        // `"is_error"` becomes `"i_error"`). The stored error ends up a
+        // garbled JSON fragment instead of the clean message, and (in the
+        // fallback.rs non-zero-exit path) skips the model cooldown that the
+        // AgentFailed branch applies.
+        let padding = "{\"type\":\"system\",\"subtype\":\"init\"}\n".repeat(50);
+        let result_line = r#"{"is_error":true,"duration_api_ms":0,"num_turns":1,"stop_reason":"stop_sequence","session_id":"32a7c67c","total_cost_usd":0,"usage":{},"modelUsage":{},"permission_denials":[],"terminal_reason":"api_error","fast_mode_state":"off","fast_mode_disabled_reason":"sdk_opt_in_required","subtype":"success","api_error_status":null,"result":"Failed to authenticate: OAuth session expired and could not be refreshed","type":"result","duration_ms":141,"uuid":"fe0ccff2-54d3-47cd-b5ba-a82f99a72b45"}"#;
+        let stdout = format!("{padding}{result_line}\n");
+        // Sanity check: the message text itself is not caught by any of the
+        // specific keyword detectors (auth/rate_limit/context_overflow), so
+        // without the fix this must fall to the generic Unknown branch.
+        assert!(
+            stdout.len() > 300,
+            "test stdout must exceed the tail-truncation window"
+        );
+
+        let err = runner().classify_error(1, &stdout, "");
+        assert!(
+            matches!(err, AgentError::AgentFailed { .. }),
+            "expected AgentFailed with the clean message, got: {err:?}"
+        );
+        let msg = err.to_string();
+        assert_eq!(
+            msg, "agent failed: Failed to authenticate: OAuth session expired and could not be refreshed",
+            "stored error must be the clean result text, not a byte-truncated JSON fragment"
+        );
+        assert!(
+            !msg.contains("i_error"),
+            "stored error must not contain the byte-truncated JSON fragment, got: {msg}"
         );
     }
 


gabrielkoerich:
🔍 Automated review started

---

gabrielkoerich:
## Automated Review — Approve

Fix is correct and well-tested. When classify_from_text's 300-byte tail truncation garbles the NDJSON result line, the clean message from the envelope parse is used instead, and AgentFailed ensures proper model cooldown via record_model_failure. CI passes, regression test validates the exact scenario from the issue. One minor note: detect_auth_error/detect_rate_limit could be run against error_result for more precise classification, but this is a follow-up, not a blocker.

---

gabrielkoerich:

---

